PR #11, opened 2015-10-22
Fixed bug with string parsing in C/C++

rogerhub:
There are 2 bugs fixed here:

1. The tokenizer did not recognize characters in the \xHH format. For
   example, '\x10'
2. The tokenizer interpreted '\0x10' as a proper character, which should
   actually be interpreted as '\0', 'x', '1', '0' (4 characters), which
   would not be valid if written as '\0x10'.

Here is some sample input for testing: (save as file.c)

```c
'\x30'
```

With the bugged version, I would get error messages similar to:

```
  Scanning Error in 'file.c':
  Lexical error at line 1, column 3.  Encountered: "x" (120), after : "\'\\"
```


rogerhub:
Also, it looks like a character string like `'\r\n'` is valid ([source](http://stackoverflow.com/questions/7755202/multi-character-constant-warnings)), but its value is implementation-dependent. Would you accept a patch that enabled support for those?

lama0206:
Yes. Because JPlag does not compile the program, the values are not important.

PR #13, opened 2016-01-29
Build fixes re. Maven relativePath and ProGuard configuration

khumba:
This fixes a bunch of warnings that Maven spits out when it analyzes the project.


khumba:
(These warnings are with Maven 3.0 and 3.3 by the way.)

lama0206:
Thanks!

PR #17, opened 2016-04-28
Ensure percentA() and percentB() don't truncate the floating-point division

erikmd:
Hello,
This is an slightly-improved patch for #16.
Sorry for opening a new PR (I just closed #16 and force-pushed my rebased branch, then realized that PRs cannot be reopened in that case, cf. isaacs/github#361).
Kind regards


lama0206:
Thanks

PR #26, opened 2016-10-24
Add python frontend

waynetee:
#26

This is my first GitHub contribution, please let me know if there are any issues.

### DIFF
--- a/jplag.frontend.scheme/src/main/java/jplag/scheme/Language.java
+++ b/jplag.frontend.scheme/src/main/java/jplag/scheme/Language.java
@@ -21,7 +21,7 @@ public class Language implements jplag.Language {
 									   // Konstruktor
 
 	public String[] suffixes() {
-		String[] res = { ".scm", ".SCM", "ss" };
+		String[] res = { ".scm", ".SCM", ".ss", ".SS" };
 		return res;
 	}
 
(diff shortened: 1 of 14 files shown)


lama0206:
Thank you for your effort. Looks good to me after quick glance. I don't have a computer with me right now and take a closer look later.

It seems that you have added the new plugin twice to the dependency management though (https://github.com/waynetee/jplag/blob/7883c3b82a16602c11323a641977978d3532dc5c/jplag.parent/pom.xml#L224). Can you please remove one of the entries?

waynetee:
Thanks for the prompt reply, it's fixed.

waynetee:
Hi, have you had the chance to look at it?

waynetee:
Added some minor enhancements

jplag.frontend.scheme/src/main/java/jplag/scheme/Language.java
```diff
@@ -21,7 +21,7 @@ public int errorsCount() {
 									   // Konstruktor
 
 	public String[] suffixes() {
-		String[] res = { ".scm", ".SCM" };
+		String[] res = { ".scm", ".SCM", "ss" };
```
erikmd: Hi @waynetee, shouldn't this be ".ss" (with a dot)?

waynetee:
Whoops, I forgot to mention its fixed

lama0206:
Sorry that it took so long. Thanks for your effort!

PR #33, opened 2017-05-30
Add Python 3.5/3.6 support

alexthornton1:
Add Python 3.5/3.6 support to the python-3 front end.  As far as I
can tell, there are two missing syntax features that have been added
in 3.5 and 3.6:

(1) Formatted string literals, such as this:

        f'Name: {name}'

I generally cleaned up string literals and binary literals in the
grammar, to be in line with the current rules in Python 3.

(2) The 'async' and 'await' keywords, which can appear in a few
places:

* 'async' can appear optionally in front of 'def', 'for', and 'with'
* 'await' can appear in front of most expressions, with a priority
  much higher than most of the other operators.

In both cases, I've added to the existing rules in Python3.g4 and
the rest of JPlag won't see these as distinct (e.g., a formatted
string literal will be seen as a string literal, 'async def' will
be seen as the beginning of a function, and so on).  I think, for
our purposes, that's probably fine, though I'm open to counterargument
on that point.

### DIFF
--- a/jplag.frontend.python-3/src/main/antlr4/jplag/python3/grammar/Python3.g4
+++ b/jplag.frontend.python-3/src/main/antlr4/jplag/python3/grammar/Python3.g4
@@ -170,9 +170,9 @@ decorated
  : decorators ( classdef | funcdef )
  ;
 
-/// funcdef: 'def' NAME parameters ['->' test] ':' suite
+/// funcdef: 'async'? 'def' NAME parameters ['->' test] ':' suite
 funcdef
- : DEF NAME parameters ( '->' test )? ':' suite
+ : ASYNC? DEF NAME parameters ( '->' test )? ':' suite
  ;
 
 /// parameters: '(' [typedargslist] ')'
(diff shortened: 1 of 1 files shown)


waynetee:
Thanks for your work, I have tested your changes and they seem to be working well
However, it seems that there has been a number of changes since 3.3/3.4
https://docs.python.org/3.4/reference/grammar.html
https://docs.python.org/3/reference/grammar.html
https://www.diffchecker.com/Hp41PHlv

This includes typing https://docs.python.org/3/library/typing.html
and matrix multiplication http://legacy.python.org/dev/peps/pep-0465/

We may need to add support for these in order to fully support Python 3.5/3.6

alexthornton1:
It looks like the @ operator from PEP 465 is already in the grammar, so I think we're on pretty safe ground there.

With regard to the new type annotations, this is mainly library support, with a new "typing" module that defines some new tools for describing types.  But, as far as I can tell, there's no new syntax there, so I think we're good on that one, too.

alexthornton1:
Actually, I stand corrected, after looking at your grammar diff in more detail.  There are some new bits in there (e.g., type annotations on variables).

lama0206:
Thanks for your efforts! Closing in favor of #34...

PR #34, opened 2017-05-30
Update Python frontend to 3.6

waynetee:
This update is inspired by @alexthornton1 #33 

This includes the latest grammar specification (3.6) found here: https://docs.python.org/3/reference/grammar.html

For future reference, in order to update the Python frontend, follow these steps:
1. Copy the Python grammar specification from the above link to the Parser rules section of Python3.g4
2. Add a semi-colon to the end of every rule
3. Replace '[' with '(' and ']' with ')?'
4. Replace 'ENDMARKER' with 'EOF'
5. Update JplagPython3Listener.java

The frontend has been tested against the Python Standard Library (https://hg.python.org/cpython/file/3.6/Lib)

### DIFF
--- a/jplag.frontend.python-3/src/main/antlr4/jplag/python3/grammar/Python3.g4
+++ b/jplag.frontend.python-3/src/main/antlr4/jplag/python3/grammar/Python3.g4
@@ -36,19 +36,14 @@ grammar Python3;
 tokens { INDENT, DEDENT }
 
 @lexer::members {
-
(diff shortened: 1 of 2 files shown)


lama0206:
Thanks for your work! Even though your update is inspired by @alexthornton1 #33 - as far as i can tell, it completely includes/replaces/overrides #33. Am I right? If so, I'd merge #34 and close #33.

waynetee:
That is correct, please go ahead.

lama0206:
:) Thanks!

PR #48, opened 2018-03-01
Fix charsets; normalize whitespace

DemiMarie:
This converts all text files to UTF-8 (with matching `<meta charset='UTF-8'> tags), fixes the DOCTYPE in all HTML files, deletes trailing whitespace in most text files, normalizes line endings to LF, deletes trailing whitespace, and adds a compile script.


lama0206:
I think that this should be already done now that i merged the other pull request...

PR #49, opened 2018-03-15
Archival submissions

ksiadzrobak:
Since I keep an archive of submissions from previous years and compare the current submissions also against them, I've been getting a huge number of matches between each and every submission in the results - most of them uninteresting.

To fix this issue, I propose a feature allowing distinguishing new submissions from the old ones (loaded with -a option, for _archival_) - new submissions will be compared with each other, and also compared against the old submissions, but there won't be comparisons of the old ones with each other. This approach reduces both the clutter in the search results and the running time.

At extreme this might be used for n-to-1 comparisons, i.e. comparing a particular submission against the whole archive.

Please review & merge.


thrau:
sorry for necroing, but this is an excellent feature. it merges just fine, and works as expected. using it in my fork, thanks!

lisander-lopez:
I like this, Im currently using this branch aswell. Unfortunately the folder structure must be 
"Archival/student/src/...." BUT, my current setup up is "Archival/FallSemester19/student/src/..."

Even thought I was -s command on, I have to point -a to "FallSemester" and not "Archival" So I have to manually do every semester

tsaglam:
Closed, as this will be implemented in #289.

PR #83, opened 2020-09-11
Docs and typofix

Alberth289346:
While trying to add a new language to JPlag, I added documentation for `jplag.Language` to decrease the ambiguity of its functions.
Most of it is by educated guess from looking at how other people use it, and how Jplag uses the function. Two remarks:

- I have no idea what the purpose of `usesIndex()` is.
- Function `errors()` seems rewritable to `errorsCount() != 0`. It looks like `errors` was intended to count each file, but jplag uses it wrongly in that case (`jplag.Submission`, line 139 is the only use of it).

### DIFF
--- a/jplag.frontend-utils/src/main/java/jplag/Language.java
+++ b/jplag.frontend-utils/src/main/java/jplag/Language.java
@@ -7,27 +7,39 @@ import java.io.*;
  * One interface for all languages...
  */
 public interface Language {
+	/** Get known suffixes for files containing code in the language. */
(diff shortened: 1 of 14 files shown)


Alberth289346:
Conceptually, it might be better to move errors into the returned `Structure` but that is likely much more involved.

tsaglam:
Adressed by PR #94, thus obsolete.

Alberth289346:
Always nice when the original author gets removed from the commit logs

PR #91, opened 2021-03-27
Added a -prior option.

keithmacspace:
This adds ability to check prior submissions to see if students are copying from earlier classes. The prior submissions are not compared against each other, only the current submissions (That way if there was cheating in a prior class it doesn't get flagged).


tsaglam:
I like the idea, however, this PR is outdated through the restructuring for the 3rd party integration support. We need to update this PR to implement the feature for the restructured version of JPlag.

tsaglam:
Closed, as this will be implemented in #289.

PR #154, opened 2021-06-25
Change -s to -m in README.md

Strohgelaender:
Commit https://github.com/jplag/jplag/commit/45d11cd681d10992b520c8b8ea15949ba9c5ecfa changed the CLI-Argument -s into -m without changing the corresponding section in the README.md. This PR changes the section to make it consistent with the current code and to avoid confusion among users (like me) who try to use this parameter when executing JPlag.


sebinside:
You're right, totally forgot about that.

Issue #119, opened 2021-05-31
Enhance overall code quality


sebinside:
There is a major restructuring effort left, even after the v3.0.0 release. This includes:
- [x] Rework JPlag options
- [x] Rework Greedy String Tiling class
- [x] Rework exit exception (moved to #245)
- [x] Clean JPlag class

Alberth289346:
Yesterday I fixed a bug in path handling, and ran into lots of opportunities to refactor

- There are "experimental" features in there, they need a decision whether to keep them or delete them imho.
- There are "hidden options", I'd say either make them unhidden or delete them.
- There is an "emphirical stduy" in there.
- There is a "distr" variable in `Options` denoting a person that got a copy.
- Filter code drags in the Text frontend
- Heaps and heaps of arrays where a `List<>` class would suffice (code was likely written in Java 1 I guess)
- Code uses `Vector<>` which is synchronized, Java docs say to prefer `ArrayList<>` for non-synchronized use.
- `printAllLanguages()` needs a `Program` instance to instantiate a language which is near impossible to get there.

tsaglam:
Thanks for the comprehensive list! I assume you work on a fork of yours based on the legacy version of JPlag. Is that correct?
Some of these issues might have been already refactored (e.g. experimental features and hidden options are gone), however, I am going to check if some of these issues are still open issues in the current v3.0.0 pre-release state.

Alberth289346:
Apparently I have an old version indeed ( fbded8822d02c852fdc017e225a3d036fe417285 which is of May 2020). You did a lot of new work there!

I am extending jplag with a frontend for a declarative automaton language to see if we can use the program.
Being declarative, code is very high level, very few details you can drop. Any change is very soon a "structural" change (swapping declarations or automata).
I tried having having many separated sections, assuming jplag would understand that many small matching sections is also a big blob of code, but it seems it searches for one consecutive big blob. We're still finding out if that is sufficient.

A bit of experiences:

At Java level, I failed to understand how to get the code in Eclipse for editing, so I ended with a copy of a merged source. (There was no documentation at all how to interface, so I needed to jump through source to read code, and for that I need the code in an editor that understands it's one project in many small pieces.)

We also have our own parser, and it's not based on antlr. Trying to use our tokens failed as jplag used more code from antlr, so I now parse the input, and then make a copy of the token list to antlr-style tokens. I had some fun with too small and too large indices as our line and column numbers are different from apparent expected values in `jplag.Token`.

[...]

Issue #107, opened 2021-05-31
Add initial test cases


sebinside:
Prior to the v3.0.0 release, we shall add at least simple test cases to ensure somewhat compatibility between the legacy version and this new release. This shall also include testing of all (important) CLI flags.

sebinside:
... and we maybe should make sure that tests are actually executed by github actions.

sebinside:
PR #175 addresses this issue and adds more initial cases (simple duplicates, no duplicates, and more advanced forms of plagiarism).

Although the testing is not complete (e.g., we should test different options in more depth), I would suggest closing this for now in favor of a sooner v3.0.0 release.

PR #155, opened 2021-06-28
Replace custom-built hash table and overhaul GST algorithm

### LINKED ISSUE #153: Replace custom-built hashtable
In `jplag.frontend-utils` the class `Structure` uses a [custom-built hashtable](https://github.com/jplag/jplag/blob/0d820422f705a43796d5cbc82aecee7c0ef43e11/jplag.frontend-utils/src/main/java/jplag/Table.java) called `Table` to map a integers to arrays of integers.
Among other quirks, the class contains a huge hard-coded array of prime numbers. We should remove the class `Table` and use standard Java API classes instead. However, we need to make sure that does not change any internal behavior.

tsaglam:
This PR replaces the old array-based hash table implementation with Java-API-based functionality. Retains the prime-based key mapping to not break the greedy string tiling algorithm. A specific prime number is used to map all keys with the same `key % prime` to the same values. The greedy string tiling algorithm relies on this mapping, which is why we retain it but replace the hard-coded prime numbers.

This PR also overhauls the greedy string tiling algorithm, removing duplicated code and avoiding the use of custom array list implementations. This thus also changes the Structure class, which is such a custom-built array list implementation.

This PR renames some essential classes:
`Structure` is now `TokenList`
`Table` is now `TokenHashMap`

This PR aims to resolve #153 addresses #119 partly.

**IMPORTANT:** This PR should not be merged before we have some more tests (as discussed in #107).


jplag.frontend-utils/src/main/java/jplag/Table.java
```diff
@@ -1,140 +1,78 @@
...
+     * @return the stored numbers or an empty array if it does not contain the modifiedKey.
+     */
+    public final int[] get(int key) {
+        int actualKey = key % primeNumber;
+        if (!mappedEntries.containsKey(actualKey)) {
+            return null;
+        }
+        List<Integer> result = mappedEntries.get(actualKey);
+        Stream<Integer> stream = Stream.concat(Stream.of(result.size()), result.stream()); // Legacy format: First entry is array length.
+        return stream.mapToInt(Integer::intValue).toArray(); // TODO TS: Requires as the GST algorithm still relies on the array format.
```
tsaglam, line 41: Note that right now we include some quirks in order to support the legacy format.
Specifically we do not only return the mapped values for the key, we append the number of mapped values as the first value to be returned. The reason for this is that the greedy string tiling algorithm relies on this format where the first entry of an array specifies the number of following entries in the array. This is deeply engrained into the algorithm and needs to be adapted before we can change this method here to only return the mapped values themselves.

tsaglam: Adressed by commits below.

PR #169, opened 2021-08-12
Add new parallel comparison strategy

tsaglam:
Enables the parallel submission comparison (except base code comparisons). Currently produces different percentages than the normal strategy, but seems the performance seems promising (S=~2.3, E=~0.29 on an i7-10510U).
Since the GST algorithm writes directly into the submissions and their tokens, a submission cannot be used for two comparisons at the same time. Thus, locks on submissions are used to compare all submission tuples in parallel in a non-blocking manner.

The default comparison mode is currently the sequential one, however, the parallel one can be used with the CLI parameter `-c parallel` or via the JPlag options.


tsaglam:
Copying the tokens has a high overhead. We should avoid that and instead make sure to not touch a submission by two threads concurrently.

tsaglam:
With this recent version, we no longer copy tokens. This current implementation uses locks on submissions to compare all submission tuples in parallel in a non-blocking manner.

jplag/src/main/java/jplag/CLI.java
```diff
@@ -71,11 +69,11 @@ public CLI() {
         parser.addArgument("-t").help("Tune the sensitivity of the comparison. A smaller <n> increases the sensitivity");
         parser.addArgument("-m").setDefault(0f).help("Match similarity Threshold [0-100]: All matches above this threshold will be saved");
         parser.addArgument("-r").setDefault("result").help("Name of directory in which the comparison results will be stored");
+        parser.addArgument("-c").choices(ComparisonMode.allNames()).setDefault(NORMAL.getName()).help("Comparison mode used to compare the programs");
```
sebinside, line 72: Please also update the readme.md

jplag/src/main/java/jplag/CLI.java
```diff
@@ -111,6 +108,8 @@ public JPlagOptions buildOptionsFromArguments(Namespace namespace) {
         options.setSubdirectoryName(namespace.getString("S"));
         options.setFileSuffixes(fileSuffixes);
         options.setExclusionFileName(namespace.getString("x"));
+        ComparisonMode.fromName(namespace.getString("c")).ifPresentOrElse(it -> options.setComparisonMode(it),
```
sebinside, line 111: Does this print the line every time one does not use the flag? Because this is probably a little bit verbose

sebinside (changes requested):
Did not find anything critical from quickly looking over the code.

However, you should probably use volume testing to test this against the sequential comparsion.

jplag/src/main/java/jplag/CLI.java
tsaglam: `Unknown comparison mode, using default` is only printed when explicitly using `-c x` with x ∉{normal, parallel}.

PR #174, opened 2021-08-12
Volume testing

sebinside:
Reads volume testing data and a csv file and verifies that
* the same number of matches are available
* the same submissions have been compared
* the comparison results only differ by 0.1%


tsaglam (approved):
Finally, we can properly test our changes! 👍

PR #224, opened 2021-11-17
Rename misleading option "maximumNumberOfMatches" and clean up JPlagResult interface

### LINKED ISSUE #217: maximumNumberOfMatches has a misleading name
The option `maximumNumberOfMatches` has a misleading name, as it actually influences how many comparisons (of submission pairs) are persisted in the report. Matches are a different thing, they are matching sections between two submissions, a comparison thus has multiple matches. 
Moreover, this option should only affect the report and not the result object. Currently `JPlagResult.getComparisons()` is limited by this option, while all can be accessed via `JPlagResult.getAllComparisons()`.

When fixing this issue one should update:
- [x] the [option with its JDoc](https://github.com/jplag/JPlag/blob/706751d130dc5d3a9634aa730458a64356064a44/jplag/src/main/java/de/jplag/options/JPlagOptions.java#L46-L50) and its getter/setter
- [x] the corresponding [CLI argument and its description](https://github.com/jplag/JPlag/blob/706751d130dc5d3a9634aa730458a64356064a44/jplag/src/main/java/de/jplag/CommandLineArgument.java#L33)
- [x] the readme and the wiki, especially regarding the CLI help text

tsaglam:
Resolves #217:
- Rename the option `maximumNumberOfMatches` to `maximumNumberOfComparisons`, adapt its access methods and documentation
- Rename the corresponding CLI argument `STORED_MATCHES` to `SHOWN_COMPARISONS`, adapt its CLI description and the readme
- Adapt JPlag so this option and CLI argument only affects the report generation and not the functionality of the `JPlagResult`:
    - `JPlagResult.getComparisons()` now returns **all comparisons** and is no longer dependent on the options
    - `JPlagResult.getAllComparisons()` was removed as it now matches the functionality of `JPlagResult.getComparisons()`
    - `JPlagResult.getComparisons(int maxCount)` still returns the first `maxCount` comparisons sorted by similarity (descending)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ named arguments:
   -x X                   All files named in this file will be ignored in the comparison (line-separated list)
   -t T                   Tune the sensitivity of the comparison. A smaller <n> increases the sensitivity
   -m M                   Match similarity Threshold [0-100]: All matches above this threshold will be saved (Standard: 0.0)
-  -n N                   Maximum number of matches that will be saved. If set to -1 all matches will be saved (Standard: 30)
+  -n N                   The maximum number of comparisons that will be shown in the generated report, if set to -1 all comparisons will be shown (Standard: 30)
   -r R                   Name of the directory in which the comparison results will be stored (Standard: result)
   -c {normal,parallel}   Comparison mode used to compare the programs (Standard: normal)
 ```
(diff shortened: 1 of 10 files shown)


tsaglam:
@SimDing, @cholakov11, @Martin9295 FYI, so there is no confusion with the changed interfaces.

PR #230, opened 2021-11-21
Fixes and some refactoring

Alberth289346:
Commits are independent changes (they can be read individually), although they may touch changes made in previous commits.

8deaf13bbd371193c174eb63073f78179c098961 is an extension of 5549b7b where submissions with insufficient tokens were removed. This however also happens when the parse fails, the `parse` function removes the tokens.

My system doesn't have java 14. Java 16 kills the build (maven crashes and I have no other maven version), so I used Java 11, and lowered the required Java version in the parent pom. The `NewJavaFeaturesTest.java` test failed (0 submissions due to parse errors in the input). This may be a follow-up error for not having Java 14. I deleted the source code of that test, and then maven works.

_maintainer edit: Resolves #227_

### DIFF
--- a/jplag/src/main/java/de/jplag/JPlagComparison.java
+++ b/jplag/src/main/java/de/jplag/JPlagComparison.java
@@ -159,6 +159,24 @@ public class JPlagComparison implements Comparator<JPlagComparison> { // FIXME T
     }
 
     /**
+     * @param getFirst Whether to return the first submission,
+     *      else return the second submission.
+     * @return The requested submission.
+     */
+    public Submission getSubmission(boolean getFirst) {
+        return getFirst ? firstSubmission : secondSubmission;
+    }
+
+    /**
+     * @param getFirst Whether to return the first basecode matches,
+     *      else return the second basecode matches.
+     * @return The requested basecode matches.
+     */
+    public JPlagComparison getBaseCodeMatches(boolean getFirst) {
+        return getFirst ? firstBaseCodeMatches : secondBaseCodeMatches;
+    }
+
+    /**
      * @return Maximum similarity in percent of both submissions.
      */
     public final float maximalSimilarity() {
--- a/jplag/src/main/java/de/jplag/JPlagComparison.java
+++ b/jplag/src/main/java/de/jplag/JPlagComparison.java
@@ -4,7 +4,9 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.IntStream;
 
 /**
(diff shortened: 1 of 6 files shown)


Alberth289346:
4e32176 is broken, will add a patch

tsaglam:
> The `NewJavaFeaturesTest.java` test failed (0 submissions due to parse errors in the input). This may be a follow-up error for not having Java 14. I deleted the source code of that test, and then maven works.

Yes, that is due to the Java version. As the Java 9+ frontend uses JavaC (`javax.tools.javacompiler`) to allow usage for Java code > Java 9, it now depends on the version you compile it with. Basically, if you compile JPlag with Java 14 the corresponding version of JavaC is used and thus Java features of a Java 14 are supported.

Alberth289346:
After a night of sleep I found 2 follow-up points

- `iter.remove()` while parsing/checking submissions feels like the wrong approach. It's slow to remove in a long arraylist, and more importantly it messes in the job that `filterValidSubmissions()` should do. I think this needs a decision where invalid submissions are removed. For paranoia checking, it might be useful to check further down the line that only valid submissions exist.
- `JPlagComparison` overrides `Object.equals` but doesn't define `Object.hashCode`. This is dangerous since `hashCode` is expected to return equals values for equal instances. A proper hashing function should be added, `equals` should not be overridden, or (as a hack) `hashCode` should crash the program so it's known not to give invalid results.

jplag/src/main/java/de/jplag/JPlagComparison.java
```diff
@@ -74,37 +76,19 @@ public boolean equals(Object other) {
...
-                }
-            }
-            res[count++] = tokenList.getToken(starti).file;
+        /*
+         * Collect the file names of the first token of each match.
+         */
+        Set<String> collectedFiles = new LinkedHashSet<>();
+        for (Match match: matches) {
+            collectedFiles.add(tokenList.getToken(match.getStart(j == 0)).file);
         }
```
tsaglam, line 86: This is quite an improvement code-style-wise. Just to have some sort of sanity test I will write a quick unit test to ensure identical functionality.

tsaglam (changes requested):
Besides a minor thing, this looks great. Thanks for the improvements!

tsaglam:
>     * `iter.remove()` while parsing/checking submissions feels like the wrong approach. It's slow to remove in a long arraylist, and more importantly it messes in the job that `filterValidSubmissions()` should do. I think this needs a decision where invalid submissions are removed. For paranoia checking, it might be useful to check further down the line that only valid submissions exist.

Ah, I just saw this comment after the review.  See the comment I made in the review. In future, we can try relying only on the filter submission method. 

>     * `JPlagComparison` overrides `Object.equals` but doesn't define `Object.hashCode`. This is dangerous since `hashCode` is expected to return equals values for equal instances. A proper hashing function should be added, `equals` should not be overridden, or (as a hack) `hashCode` should crash the program so it's known not to give invalid results.

Yes, I am aware, this is something that needs change but this is also a change that should be properly verified as it has potential to have some side effects.

jplag/src/main/java/de/jplag/JPlagComparison.java
tsaglam: Works as expected!

Alberth289346:
Fixed by interactive rebase:

Removed `8deaf13 Remove sumbissions with parse errors due to having invalid length (token of submission < min token match)` entirely, since that patch added the `iter.remove` and merged the `if`s (and nothing else).

Merged `4e32176 Simplify extracting a file list from a comparison.` and `Fix 4e32176` since the former removal would destroy any meaning of the hash number.

jplag/src/main/java/de/jplag/JPlagComparison.java
Alberth289346: I ran a test by checking lengths of both results before deleting the old `count` computation, should perhaps have added that.

tsaglam (approved):

PR #231, opened 2021-11-21
SubmissionCollection moves

Alberth289346:
Introducing a `SubmissionCollection` and moving common part of submission parsing to `Submission`.

This is stacked on top of #230 but much bigger, and it may need some tweaking or discussing before it's good. Don't merge before #230 I'd say.

As explained in #230, I am having a fight with Java versions and maven.

I didn't do testing beyond running maven that includes some tests.

### DIFF
--- a/jplag/src/main/java/de/jplag/JPlag.java
+++ b/jplag/src/main/java/de/jplag/JPlag.java
@@ -14,6 +14,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import de.jplag.options.JPlagOptions;
@@ -29,7 +30,6 @@ import de.jplag.strategy.ParallelComparisonStrategy;
 public class JPlag implements Program {
 
     // INPUT:
-    private Submission baseCodeSubmission = null;
     private HashSet<String> excludedFileNames = null; // Set of file names to be excluded in comparison.
     private Language language;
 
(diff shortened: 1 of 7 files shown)


Alberth289346:
Apparently it throws #230 and this patch onto one heap of changes, making the changes unreadable :(  
Don't know how to fix other than wait until #230 is merged,

Marked it as draft to avoid accidental merge, however if you're happy with the changes, feel free to convert it back to ready.

tsaglam:
Maybe you can update the PR, now that #230 is merged.
Also, note that I turned the our discussion regarding submission filtering into issue #232, which might be also relevant for this PR.

Alberth289346:
New version, less involved than the previous try.

I contemplated whether to move the `JPlag` parsing code into `SubmissionSet`, but that likely isn't going to be very useful with all the links back to the `Program` interface of the `JPlag` class.

tsaglam:
> I contemplated whether to move the `JPlag` parsing code into `SubmissionSet`, but that likely isn't going to be very useful with all the links back to the `Program` interface of the `JPlag` class.

I like the changes in this PR. I checked out this branch and I think I found a way to reasonably move most of the parsing code into the submission set (basically separating the error reporting capabilities defined by `Program` into an own class). I will try to fully implement that on top of your changes and then we can maybe discuss them here.

Alberth289346:
WIP: https://github.com/Alberth289346/jplag/compare/subm-collection-moves...Alberth289346:frutsel-program?expand=1

I had a go too, created a new `ProgramProxy`, acting as Program for the language parsers, then started hacking the `implements Program` out of `JPlag`. Also pulled `Language language` out of options as I didn't see much need for it there, but report now dies :)

tsaglam:
> I had a go too, created a new `ProgramProxy`, acting as Program for the language parsers, then started hacking the `implements Program` out of `JPlag`. Also pulled `Language language` out of options as I didn't see much need for it there, but report now dies :p

I also removed the error collecting from the program and changed the language frontend dependencies to no longer reference the main class. However, I think it is better to leave the language in the options, due to other features that are currently worked at. Maybe you can take a look at #234, if it is okay with you I would go ahead with that approach (so first merge #231 and then #234).

PR #234, opened 2021-11-24
De-tangle main class to separate concerns.

### LINKED ISSUE #232: Streamline submission filtering and submission error handling
Currently, there are multiple ways how submissions are excluded. Some are removed during iterating while parsing (see `JPlag.parseSubmissions()`), while others are marked as erroneous and then later filtered in a later step. We should streamline this process.
As pointed out by @Alberth289346 in #230:

> `iter.remove()` while parsing/checking submissions feels like the wrong approach. It's slow to remove in a long arraylist, and more importantly, it messes in the job that `filterValidSubmissions()` should do. I think this needs a decision where invalid submissions are removed. For paranoia checking, it might be useful to check further down the line that only valid submissions exist.

In my opinion, submissions should be marked as erroneous and maybe even given the information why they are erroneous. Then, they should be filtered later on. If we introduce a submission set (i.e. with #231), this could be done in there resulting in two methods: `getValidSubmissions()` and `getInvalidSubmissions()`. The latter is also useful for the new report generation, as the current `JPlagResult` does not expose which submissions were not used for comparison.

tsaglam:
Depends on #231 and should be merged after it!
Separates the different concerns of the main class `JPlag`:

- The error collecting capabilities defined by `Program` (now called `ErrorConsumer`) are moved into the new class `ErrorCollector`
- The submission parsing is mostly moved into the `SubmissionSet` (thus depending on #231)
- The submission reading is moved into a dedicated class called `SubmissionSetBuilder`
- The language frontends no longer need to hold a dependency on the `JPlag` class
- Other classes also no longer have a reference on the `JPlag` class
- Resolves #232 by ensuring proper filtering without removal during iteration and keeping invalid submissions for the report

### DIFF
--- a/jplag.frontend-utils/src/main/java/de/jplag/AbstractParser.java
+++ b/jplag.frontend-utils/src/main/java/de/jplag/AbstractParser.java
@@ -5,7 +5,7 @@ package de.jplag;
  * @date 22.01.2005
  */
 public abstract class AbstractParser { // TODO TS: We should rename this class, as all concrete parsers shadow its name.
-    protected Program program;
+    protected ErrorReporting program;
     protected int errors = 0;
     private int errorsSum = 0;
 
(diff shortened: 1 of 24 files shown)


Alberth289346 (approved):
Looks great!

PR #235, opened 2021-11-24
Store basecode comparisons in each submission instead in the comparison strategy

### LINKED ISSUE #216: Basecode matches should be stored by each submission
Currently, the comparisons between any submission and the basecode are stored in the comparison strategy in a hashtable. Later they are stored in the comparison objects. It would be cleaner to let the submissions store their basecode comparisons internally. See the correlating [lines of code](https://github.com/jplag/JPlag/blob/c91c1605e2709bb5619cbce546a2b41e63743e6e/jplag/src/main/java/de/jplag/strategy/AbstractComparisonStrategy.java#L14-L16).

dogayalcin:
The submissions store their basecode matches internally.
closes #216

### DIFF
--- a/jplag/src/main/java/de/jplag/Submission.java
+++ b/jplag/src/main/java/de/jplag/Submission.java
@@ -51,9 +51,9 @@ public class Submission implements Comparable<Submission> {
     private TokenList tokenList;
 
     /**
-     * baseCodeMatch
+     * Base code comparison
      */
-    private JPlagComparison baseCodeMatch;
+    private JPlagComparison baseCodeComparison;
 
     private final Language language;
     private final ErrorCollector errorCollector;
@@ -100,18 +100,18 @@ public class Submission implements Comparable<Submission> {
     }
 
     /**
-     * Sets the base code match
-     * @param baseCodeMatch is submissions matches with the base code
+     * Sets the base code comparison
+     * @param baseCodeComparison is submissions matches with the base code
      */
-    public void setBaseCodeMatch(JPlagComparison baseCodeMatch) {
-        this.baseCodeMatch = baseCodeMatch;
+    public void setBaseCodeComparison(JPlagComparison baseCodeComparison) {
+        this.baseCodeComparison = baseCodeComparison;
     }
 
     /**
-     * @return base code match
+     * @return base code comparison
      */
-    public JPlagComparison getBaseCodeMatch() {
-        return baseCodeMatch;
+    public JPlagComparison getBaseCodeComparison() {
+        return baseCodeComparison;
     }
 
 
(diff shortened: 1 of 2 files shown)


tsaglam:
Thanks for your contribution. Due to #234 your PR currently has some conflicts with the master branch.
I recommend solving the conflicts first, then I can review your changes.

jplag/src/main/java/de/jplag/Submission.java
```diff
@@ -50,6 +50,11 @@
      */
     private TokenList tokenList;
 
+    /**
+     * baseCodeMatch
+     */
+    private JPlagComparison baseCodeMatch;
+
```
tsaglam, line 57: The name and description are a bit misleading as comparisons are no matches. A comparison can have multiple matches. In some places in our source code, this still might be wrong. I would propose using `baseCodeComparison ` here. Also, the JDoc comment is a bit redundant, as it gives no valuable information.

jplag/src/main/java/de/jplag/Submission.java
```diff
@@ -94,6 +99,22 @@ public int getNumberOfTokens() {
         return tokenList.size();
     }
 
+    /**
+     * Sets the base code match
+     * @param baseCodeMatch is submissions matches with the base code
+     */
+    public void setBaseCodeMatch(JPlagComparison baseCodeMatch) {
```
tsaglam, line 106: Adapt name and comment with field (see above)

jplag/src/main/java/de/jplag/Submission.java
```diff
@@ -94,6 +99,22 @@ public int getNumberOfTokens() {
...
+        this.baseCodeMatch = baseCodeMatch;
+    }
+
+    /**
+     * @return base code match
+     */
+    public JPlagComparison getBaseCodeMatch() {
```
tsaglam, line 113: Adapt name and comment with field (see above)

jplag/src/main/java/de/jplag/strategy/AbstractComparisonStrategy.java
```diff
@@ -11,10 +10,6 @@
 
 public abstract class AbstractComparisonStrategy implements ComparisonStrategy {
 
-    // TODO PB: I think it's better to make each submission store its own matches with the base code.
-    // Hashtable that maps the name of a submissions to its matches with the provided base code.
-    private Hashtable<String, JPlagComparison> baseCodeMatches = new Hashtable<>(30);
```
tsaglam, line 16: Ah, I see, it is falsely named here, so not your fault. Let us use this opportunity to fix it though.

tsaglam (changes requested):
Looks good, just a minor name issue that already persisted in the original code.

tsaglam (approved):

Issue #198, opened 2021-10-15
JPlagComparison contains code duplication


tsaglam:
The class [`JPlagComparison`](https://github.com/jplag/jplag/blob/master/jplag/src/main/java/de/jplag/JPlagComparison.java) contains code duplication which should be cleaned up as this makes the class confusing and unnecessarily complicated.
A `JPlagComparison` object contains two submissions and a List of matches between those submissions (a match just being a sequence of tokens in both submissions). However, the similarity calculation is duplicated for both submissions instead of being done independently of the submission fields.

PR #236, opened 2021-11-24
Clean comparison somewhat

Alberth289346:
Eliminated the hashmap in the abstract comparison strategy base class as suggested, and moved similarity divisor computing as well. 

_Maintainer edit: Resolves #198_


tsaglam:
PR #235 also moved the basecode matches into the submissions, thus your PR has conflicts now.
However, your PR contains a few great changes resolving #198, so I would say let's solve the conflicts and then merge.

Alberth289346:
Path may be to make a `Matches` class and move match-related code in it. Perhaps make a 'good first issue' for it to replenish the supply as I seem to accidentally resolve them :)

tsaglam (approved):

PR #237, opened 2021-11-25
Cleaning up rootdir checking and subdir sanity

Alberth289346:
ADDED:

- `JPlag.getLanguage()` was now the only place where it was used, and asking for a variable of its own instance variable through a public method is a bit weird then, so I deleted the function (Perhaps a left-over from not deriving from `Program` any more?)
- Root directory sanity was checked at several places in different ways which begs for unitfication.
- Basecode option value got sneakily changed while checking sanity of supplied values. Changing it inside the `Options` is more robust.
- I saw the  subdirectory too there, and just added that change there too, as somewhere an equality test against a directory name from the file system is performed so any directory-separators in the name won't do any good.

I forgot to run the tests, and promptly the patch was rejected, as the above names can be `null` since they are optional.

### DIFF
--- a/jplag/src/main/java/de/jplag/JPlag.java
+++ b/jplag/src/main/java/de/jplag/JPlag.java
@@ -87,7 +87,6 @@ public class JPlag {
                             ExitException.BAD_PARAMETER);
                 }
             }
-            options.setBaseCodeSubmissionName(baseCode);
             System.out.println("Basecode directory \"" + baseCodePath + "\" will be used");
         }
     }
--- a/jplag/src/main/java/de/jplag/JPlag.java
+++ b/jplag/src/main/java/de/jplag/JPlag.java
@@ -51,7 +51,7 @@ public class JPlag {
         if (!rootDir.isDirectory()) {
             throw new ExitException(options.getRootDirectoryName() + " is not a directory!");
         }
-        
+
         // 2. Parse and validate submissions:
         SubmissionSetBuilder builder = new SubmissionSetBuilder(language, options, errorCollector);
         SubmissionSet submissionSet = builder.buildSubmissionSet(rootDir);
@@ -72,14 +72,6 @@ public class JPlag {
         System.out.println("Total time for comparing submissions: " + TimeUtil.formatDuration(result.getDuration()));
         return result;
     }
-
-    /**
-     * @return the configured language in which the submissions are written.
-     */
-    public Language getLanguage() {
-        return language;
-    }
-
 
     /**
      * This method checks whether the base code directory value is valid.
@@ -142,8 +134,8 @@ public class JPlag {
             throw new ExitException("Language instantiation failed", ExitException.BAD_LANGUAGE_ERROR);
         }
 
-        this.options.setLanguageDefaults(this.getLanguage());
+        this.options.setLanguageDefaults(this.language);
 
-        System.out.println("Initialized language " + this.getLanguage().getName());
+        System.out.println("Initialized language " + this.language.getName());
     }
 }
--- a/jplag/src/main/java/de/jplag/JPlag.java
+++ b/jplag/src/main/java/de/jplag/JPlag.java
@@ -71,7 +71,7 @@ public class JPlag {
         getRootDirectory(); // Performs checks on the root directory.
 
         if (options.hasBaseCode()) {
-            String baseCode = options.getBaseCodeSubmissionName().replace(File.separator, ""); // trim problematic file separators
+            String baseCode = options.getBaseCodeSubmissionName();
             if (baseCode.contains(".")) {
                 throw new ExitException("The basecode directory name \"" + baseCode + "\" cannot contain dots!", ExitException.BAD_PARAMETER);
             }
--- a/jplag/src/main/java/de/jplag/JPlag.java
+++ b/jplag/src/main/java/de/jplag/JPlag.java
@@ -43,18 +43,9 @@ public class JPlag {
      * @throws ExitException if the JPlag exits preemptively.
      */
     public JPlagResult run() throws ExitException {
-        // 1. Preparation:
-        File rootDir = new File(options.getRootDirectoryName());
-        if (!rootDir.exists()) {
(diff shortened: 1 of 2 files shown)


tsaglam:
Can you add a short description for the PR.
I would be interested in what cases lead to the file separator problems and how the problems manifested themselves.

jplag/src/main/java/de/jplag/JPlag.java
```diff
@@ -67,29 +58,19 @@ public JPlagResult run() throws ExitException {
...
      */
     private void checkBaseCodeOption() throws ExitException {
-        if (options.hasBaseCode()) {
-            if (!new File(options.getRootDirectoryName()).exists()) {
-                throw new ExitException("Root directory \"" + options.getRootDirectoryName() + "\" doesn't exist!", ExitException.BAD_PARAMETER);
-            }
+        getRootDirectory(); // Performs checks on the root directory.
 
+        if (options.hasBaseCode()) {
             String baseCode = options.getBaseCodeSubmissionName().replace(File.separator, ""); // trim problematic file separators
```
tsaglam, line 74: The file separator of the basecode submission name is already trimmed here.

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -218,11 +219,13 @@ public void setRootDirectoryName(String rootDirectoryName) {
     }
 
     public void setBaseCodeSubmissionName(String baseCodeSubmissionName) {
-        this.baseCodeSubmissionName = baseCodeSubmissionName;
+        // Trim problematic file separators.
+        this.baseCodeSubmissionName = (baseCodeSubmissionName == null) ? null : baseCodeSubmissionName.replace(File.separator, "");
```
tsaglam, line 223: This is now duplicated, we should only trim once to ensure maintainability.

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -218,11 +219,13 @@ public void setRootDirectoryName(String rootDirectoryName) {
...
     }
 
     public void setSubdirectoryName(String subdirectoryName) {
-        this.subdirectoryName = subdirectoryName;
+        // Trim problematic file separators.
+        this.subdirectoryName = (subdirectoryName == null) ? null : subdirectoryName.replace(File.separator, "");
     }
```
tsaglam, line 229: Now that I look at that code I wonder if settings null here should even be allowed? If not, we should check for null and throw an `IllegalArgumentException` if null. If not, we can safely trim here without the not to nice `(subdirectoryName == null) ? null` check.

jplag/src/main/java/de/jplag/options/JPlagOptions.java
tsaglam: But I think trimming here is a good idea.

tsaglam (changes requested):

jplag/src/main/java/de/jplag/options/JPlagOptions.java
tsaglam: In which cases did you previously encounter an NPE here? What caused it?

jplag/src/main/java/de/jplag/JPlag.java
Alberth289346: Yes, and at old line 109 below it gets sneaked back into the options, so it's changed in this function.

Since this code is called from the constructor, all other uses of `options.getBaseCodeSubmissionName()` get the trimmed version, and can compare `File.getName()` with the option value and get a correct answer.  
I just moved this trim call to the option setter itself.

I see now it's not deleted, I agree that's not good.

jplag/src/main/java/de/jplag/options/JPlagOptions.java
Alberth289346: One of the test cases crashed.

Afaik `subdirectoryName` is an optional name of a folder inside a submission containing the source code (`root/*` for submissions versus `root/*/subdir`). Being optional, I'd expect it to be `null` sometimes (which seems to be the case for its definition ` SUBDIRECTORY("-S", String.class),`, there is no default value.

Alberth289346: Hmm, actually I fixed the crash on the base code and then used the above idea to conclude that sub directory needed it as well.

tsaglam: Ah, I see, the CLI will set all options, since argparse4j will return null if an argument is not set.
We could change the CLI to only call the setter if the value is non-null, however for now we can leave it like this.

jplag/src/main/java/de/jplag/JPlag.java
tsaglam: > I see now it's not deleted, I agree that's not good.
Yes, let's delete it here and keep it in the options.

Alberth289346:
So remove the `replace` call at new line 74 ?

Alberth289346:
Done.

fyi, both directory names can be null judging from the code, see below.

The only remaining `replace` call is in argument `-` prefix stripping.

The only remaining `set` calls on the basecode name and subdir name are from argument to option copying, ie no more sneaky setting them while processing submissions.

### basecode directory name usage
*SubmissionSet.mapFileNamesToSubmissions*, line 122:
`if (options.hasBaseCode() && options.getBaseCodeSubmissionName().equals(fileName)) {`

There is also a retrieval from html report printing (and inserted without html escaping, yay! :) ).

### subdirectory name usage

*SubmissionSet.mapFileNamesToSubmissions*, line 106:
`if (submissionFile.isDirectory() && options.getSubdirectoryName() != null) {`


*JPlag.checkBaseCodeOption*, lines 83/84:
```
String subdirectory = options.getSubdirectoryName();
if (subdirectory != null && subdirectory.length() != 0) {
```

tsaglam (approved):

tsaglam:
Looks good to me!

> There is also a retrieval from html report printing (and inserted without html escaping, yay! :)

Thankfully, it is only a matter of time now until we get rid of the old hardcoded HTML report generation!

PR #238, opened 2021-11-25
Extend jplag for specifying multiple root directories

Alberth289346:
Stacked on the previous patch, but only cc7d854e0ebd5dd2e42dcbcd41a90b44132c27ba is relevant.
It's not finished yet, but the idea is that you can specify several directories instead of just one.

Known things that are missing:
- The argparse library doesn't know yet it should allow multiple cli arguments. It's a solved problem though, just didn't add it yet.
- There is no basecode selection or checking code. Different submission sets may have different base code or only some of them may have it.
  - The latter is likely simple, check that either all sets have a basecode or all sets don't have a basecode.

  A few solutions for the former:
  - Use the basecode associated with each set for its submissions.
  - Assume or force all basecode is equal between all sets.


Alberth289346:
2nd attempt on having multiple root directories, it runs so you can experiment with it.

@tsaglam  Could you have a look and share some thoughts?

The approach was to first pave the way with the first two commits (deleting the parameter and splitting comparisons computing away from both strategies). Next, I introduced multiple root-directories at the commandline, and then pushed this change from front to back through the program. Extensions for comparison computing were needed, and for submission naming, as I had lots of  "[AB] -> [AB]" comparisons and no clue where they came from.

First experiments raised a problem and a semantical issue.
- A mix of root directories with and without basecode doesn't fly, since the program demands basecode everywhere if you specify it.
- Each root directory may have different basecode, with thus leads to different comparison results, even across root directories.
  The reason for this is that basecode matches are not considered in submission comparisons, so depending on the basecode a larger or smaller part of a submission may be considered for checking against other submissions.

While the former seems solvable (eg assume an empty basecode file in such a case), the latter seems at least tricky.

I pondered about it, and I believe now that glue-ing the basecode to a root directory is the wrong solution. Instead the program should have a separate option for specifying the basecode, and it should be used for all submissions, using the reasoning that for a course there is a single basecode each year. The concept of plagiarism in a year should be related to that and only that basecode.

One way to get this is to have for example a `--global-basecode PATH` option for specifying the basecode, and dropping or ignoring or suppressing or forbidding the `-bc` option.

A more advanced way is to merge all basecode from all root directories and consider that to be basecode (ie if a student copies basecode from last year which isn't included in the basecode this year, is that plagiarism?  Technically it is, in practice I am less sure, especially if the student followed the course previous year as well.)

The latter way seems much more involved, than the former way, so a separate option looks a better solution now.

Alberth289346:
> Or just changing the CLI that the basecode is always specified independently of any root directory.

That's another option but it breaks backwards compatibility. As such, a project strategy decision that needs to be made.

As an experiment I assumed we'd go at least for a way to specify a single global basecode. I added a `-g GLOBAL/BASE/CODE/PATH`  option, and modified the `-bc` code to skipping the basecode submission in a root directory if a `-g` option is specified (`-g` takes precedence). In my experiment this lead to 4 possible cases:

- No `-bc` and no `-g` option disables basecode handling completely (old behavior).
- Only `-bc` works as before, even with multiple root directories (ie behavior as coded in this patch. Old behavior in case of 1 root directory, quite not well-defined otherwise to say the least).
- Only `-g` makes every root directory use the same basecode (specified as value of the `-g` option). New, but equivalent to having the same basecode files added to every set).
- Both '-g' and '-bc' uses the submission specified by `-g` as the basecode, and skips the sub-directory in the roots stated by `-bc`.

The 2nd case should probably be limited to 1 root directory only if it is kept.  
For the 4th case, while "`-g` takes precedence over `-bc`" goes a long way as a user explanation, it might be non-desired. One solution is to block combined use of both `bc` and `g` options at the same run but it doesn't look cost-free (parsing shouldn't run into a basecode submission being interpreted as a use submission, so something must change there).

In case you're not interested in backwards compatibility, one semi-clean path out is to discard the `bc` option and use a new option name for the global basecode path specification. Old scripts will then die so the user is at least warned about the change. One step further could be to continue detecting `-bc`, and print an explanation what to do instead before dying.

Alberth289346:
> You made the decision to keep the separation of the root directories by using multiple submission sets. I am not sure if that is the best design decision as it leads to ....

Indeed, for "cross-check everything" merging is the simplest way out. However for archival support it is not (at least for what I implemented as an experiment).

I added `-unchecked` and `-checked` options for denoting the state of a set. The options allow several argument values and may be used more than once. They denote the interest in checking for plagiarism in the stated sets. Semantically, "unchecked" is what the program currently assumes (ie need to cross-check everything), while "checked" is for submission sets with known (or non-interesting) plagiarism state (what you seem to name "archived"). 
Semantically, all checked submissions are thrown at heap C and all unchecked submissions are thrown at heap U. Everything in U is cross-checked pairwise. Also everything in U is checked against everything in C.

In the implementation there is no need to make these large heaps U and C. Each set gets a "needs checking" value, and the tuple builder simply skips generating all check-tuples involving only sets from heap U. Two simple if statements and done.

More practically:

In my experiment I also kept `rootdirs` for backwards compatibility (they are flagged as "unchecked")
As an example, you can write something like
``` sh
jplag r1 r2 -checked r3 -unchecked r4 r5 -checked r6
```
This example results in sets with root `r1`, `r2`, `r4,` and `r5` as being unchecked and sets with root `r3` and `r6` as being already checked.

Alberth289346:
Sorry for the large explanations, but better getting things clear before more programming imho.

I agree with 1 thing at a time, so perhaps keeping this as draft and make new pull requests for those steps?

tsaglam:
**Regarding global base code:**

> > Or just changing the CLI that the basecode is always specified independently of any root directory.
> 
> That's another option but it breaks backwards compatibility. As such, a project strategy decision that needs to be made.

We could implement in a way that the basecode is specified globally with -bc (`path/to/base`), however, if it is just a single dir name (`base)` which does not resolve globally and one of the root dirs contains a submission with that name then we could show a warning and just use this submission. As a consequence, a user specifying the basecode locally will still execute correctly but will receive a warning.

> As an experiment I assumed we'd go at least for a way to specify a single global basecode. I added a `-g GLOBAL/BASE/CODE/PATH`  option, and modified the `-bc` code to skipping the basecode submission in a root directory if a `-g` option is specified (`-g` takes precedence). In my experiment this lead to 4 possible cases: 
> [...]
> In case you're not interested in backwards compatibility, one semi-clean path out is to discard the `bc` option and use a new option name for the global basecode path specification. Old scripts will then die so the user is at least warned about the change. One step further could be to continue detecting `-bc`, and print an explanation what to do instead before dying.

I am not the biggest fan of two arguments for basecode. I would rather move to global basecode without changing the argument name (so keep bc). And then maybe add the behavior I described above, if possible.

**Regarding multiple root dirs and archival:**
First of all some thoughts: I think we can separate two concerns: The user interfacing CLI arguments (as in: _how_ does the user specify checked/unchecked and multiple root dirs) and the internal representation and handling during comparison.
For the first concern, simplicity is key, making the CLI foolproof and keeping backward compatibility (if possible).

> I added `-unchecked` and `-checked` options for denoting the state of a set. The options allow several argument values and may be used more than once. They denote the interest in checking for plagiarism in the stated sets. Semantically, "unchecked" is what the program currently assumes (ie need to cross-check everything), while "checked" is for submission sets with known (or non-interesting) plagiarism state (what you seem to name "archived"). 

I like the idea. So unchecked would be the default whenever neither is explicitly specified to keep backward compatibility. Not sure if checked or unchecked are the right terms to convey the user the semantics, but that is only a minor concern and easy to change later before merging.

> In the implementation there is no need to make these large heaps U and C. Each set gets a "needs checking" value, and the tuple builder simply skips generating all check-tuples involving only sets from heap U. Two simple if statements and done.

I agree, this allows us to be relatively non-intrusive in regards to the core algorithms, which means not adding too much complexity to an already overly complicated GST-algorithm implementation.

> ``` sh
> jplag r1 r2 -checked r3 -unchecked r4 r5 -checked r6
> ```
> This example results in sets with root `r1`, `r2`, `r4,` and `r5` as being unchecked and sets with root `r3` and `r6` as being already checked.

So in practice, submissions from r3 ∪ r6 are checked amongst each other and against submissions from the other sets, but submissions from r1 ∪ r2 ∪ r4 ∪ r5 are not checked amongst each other, right? 

> EDIT: I was thinking you definitely need at least 2 root directories or you cannot specify both sets. More than 2 sets is then a small step, and it gives more flexibility in how you organize your submissions without a direct need to add globbing. (That is, listing each and every root directory explicitly can be seen as a more basic method of globbing.)

Yes, a single checked root dir should be illegal as well as only checked root dirs. But this should be easy to ensure. So multiple root dirs first, then /checked/unchecked. Globbing is something we can add further down the road. All root dirs should be parsed into one submission set, but submissions have a field specifying whether its a checked or unchecked submission.

> Sorry for the large explanations, but better getting things clear before more programming IMHO.

No problem, I fully agree with that!

> I agree with 1 thing at a time, so perhaps keeping this as draft and make new pull requests for those steps?

Sounds good!

Alberth289346:
Somewhat nit-picking, but you swapped 'checked' and 'unchecked' meaning.

I wrote

> Semantically, "unchecked" is what the program currently assumes (ie need to cross-check everything), while "checked" is for submission sets with known (or non-interesting) plagiarism state (what you seem to name "archived").

>  ```
>  jplag r1 r2 -checked r3 -unchecked r4 r5 -checked r6
> ```
> This example results in sets with root r1, r2, r4, and r5 as being unchecked and sets with root r3 and r6 as being already checked.

You wrote:

> So in practice, submissions from r3 ∪ r6 are checked amongst each other and against submissions from the other sets, but submissions from r1 ∪ r2 ∪ r4 ∪ r5 are not checked amongst each other, right?

It's just the other way around, checked is "I don't care about them", and unchecked is "is there plagiarism in these??".

I agree the option names are non-intuitive. The same problem exists with "known" versus "unknown" I think. I don't like "new" and "archived" or "old" much as they are too much tailored to 1 specific use-case. There are likely other valid reasons to check for duplication in a collection, where "old" and "new" may not make much sense.

We discussed this in the past, and we found other names, but I'll have to look them up when I reach that point in the patches.

Issue #114, opened 2021-05-31
Revisit language modules


sebinside:
Existing frontends shall be revisited, e.g. by:
- [x] Removing outdated frontends
- [x] Clean-up and modernize the remaining legacy frontends (with minimal effort)
- [x] Moving frontends into one common directory
- [x] Verify existing class loader logic, add stability

tsaglam:
Closed, as the final tasks were addressed by #463 and #574 for #454.

PR #242, opened 2021-11-29
Delete old Java frontends

tsaglam:
Removes the old language frontends (contributing towards #114):

- java1
- java2
- java5
- java5dm
- java7

The newest Java frontend is now just called `java` instead of `java9`.


README.md
```diff
@@ -35,7 +35,7 @@ Usage: JPlag [ options ] [<root-dir>]
 
 named arguments:
   -h, --help             show this help message and exit
-  -l                     {java1,java2,java5,java5dm,java7,java9,python3,cpp,csharp,char,text,scheme} Select the language to parse the submissions (Standard: java9)
+  -l                     {java,python3,cpp,csharp,char,text,scheme} Select the language to parse the submissions (Standard: java9)
```
Alberth289346, line 38:
```suggestion
  -l                     {java,python3,cpp,csharp,char,text,scheme} Select the language to parse the submissions (Standard: java)
```

Alberth289346:
Not sure removing java version numbers will work in the long term. Both `C++` and `Java` (and possibly other languages as well) create new versions at such an incredible high rate that existing courses may not always follow while new courses may jump to whatever the newest version is at the time.

To cover both audiences, the program may need support for more than one java version.

To handle removal, the project could have a guideline when a front end is considered obsolete (example, older than 5 years before the newest release).

tsaglam:
I think there is a bit of a misunderstanding here: The legacy Java frontends are only for a single version. For each Java update, a new frontend had to be created to support new language features. This was tedious and just not ideal. The Java 9 frontend is actually a Java 9+ frontend that was introduced for Java 9. It supports future Java versions if JPlag is compiled with that version.

We can remove the old Java frontends, as the Java9+ frontend should be able to parse older code as well (additionally, there this is the legacy version of JPlag). For future java versions, we just need to update the whole project to be compiled with that, and then we should be able to automatically support the new language features. As an example, #159 is such an upgrade. This is why we renamed the Java 9+ frontend, it is actually version-independent and always has been (explicitly to deal with the rapid release of Java versions). The version number was just to distinguish it from the redundant frontends that we will remove with this PR.

For the other frontends, we do not actually know in what state they are and what language versions they support or don't support. Since we use JPlag mainly for Java code, this frontend is well maintained and up to date. The other frontends might already be unmaintained and out-of-date, but I a do not really know.

In my opinion, a frontend is considered as obsolete, if another frontend covers the same language version(s) and also one or more additional language versions. New frontends are always welcome, if they bring some capabilities that the existing frontends do not possess. Overhauls/modernizations of existing frontends are also welcome!

**TL;DR The Java 9+ frontend is version-independent making the legacy java frontends redundant. The state of the other frontends is (largely) unknown.**

README.md
tsaglam: Good catch! I will fix that!

PR #244, opened 2021-11-30
Use generic List interface as declared type instead of ArrayList

Alberth289346:
Cleaning up usage of lists.

First commit replaces use of `ArrayList` in a type instead of the `List` interface.
I noted that `SubmissionSetBuilder.parseFilesRecursively(File file)` returns a more generic `Collection<File>` type instead of a list. Not sure why but I left it untouched. This commit is probably fully acceptable.

Second commit introduces a few small almost-wrapper list functions, and uses them. I think it improves the code but I am used to this style of code :) 
I can live without this change.

Obviously you can extend the idea to other areas in the code to reduce clutter. (I have one for assertion checking too, it's barely used currently, as the `ExitException` isn't entirely clear what it covers (end-user errors or developer-oriented errors).

### DIFF
--- a/jplag.frontend-utils/src/main/java/de/jplag/TokenHashMap.java
+++ b/jplag.frontend-utils/src/main/java/de/jplag/TokenHashMap.java
@@ -13,7 +13,7 @@ import java.util.Map;
  * that is larger or equal to the specified size (see {@link TokenHashMap#Table(int)}).
  */
 public class TokenHashMap {
-    private final Map<Integer, ArrayList<Integer>> mappedEntries;
+    private final Map<Integer, List<Integer>> mappedEntries;
     private final int primeNumber;
 
     /**
@@ -50,7 +50,7 @@ public class TokenHashMap {
         if (mappedEntries.containsKey(actualKey)) {
             mappedEntries.get(actualKey).add(value);
         } else {
-            ArrayList<Integer> entries = new ArrayList<>();
+            List<Integer> entries = new ArrayList<>();
             entries.add(value);
             mappedEntries.put(actualKey, entries);
         }
--- a/jplag/src/main/java/de/jplag/Submission.java
+++ b/jplag/src/main/java/de/jplag/Submission.java
@@ -13,6 +13,7 @@ import java.nio.file.Path;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 
 import de.jplag.options.JPlagOptions;
 
@@ -184,7 +185,7 @@ public class Submission implements Comparable<Submission> {
     public String[][] readFiles(String[] files) throws de.jplag.ExitException {
         String[][] result = new String[files.length][];
         String help;
-        ArrayList<String> text = new ArrayList<>();
+        List<String> text = new ArrayList<>();
 
         for (int i = 0; i < files.length; i++) {
             text.clear();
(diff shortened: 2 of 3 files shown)


tsaglam:
> ExitException isn't entirely clear what it covers (end-user errors or developer-oriented errors).

To my knowledge, the `ExitException` is used for preemptive purposes, meaning end-user errors. However, it is still largely a legacy component of the code and thus not very well implemented.

Regarding this PR in general, there are some changes in this commit I disagree with:
- Encapsulating `new ArrayList<>();` in a utility method `list()` is neither necessary nor useful, as it makes the code unreadable for new developers, hides the actual type of the list (which might matter), and has no benefit at all except maybe saving some characters while typing.
- Encapsulating `new ArrayList<>(tokens);` in a utility method `copy(tokens)` makes the code more obscure (especially for new developers), hides the type of the newly created list, and promotes the overusing of static imports.
- Same for `listc`, but it also is not a speaking name and thus not really intuitive (we don't want abbreviations as method, class or variable names and are trying to replace names that are abbreviated due to legacy code).
- If you just want a short way of grouping elements in a read-only list, using `List.of` is better than creating your own utility method that is unintuitive to new users and relies heavily on static imports.
- `.collect(Collectors.toCollection(ArrayList::new))` cannot be replaced with `collect(toList())`, as `toList()` allows no assumption on the underlying data structure. However, for performance reasons we might want an `ArrayList`.
- Utility methods have their use-cases, but here they don't bring any benefit but come with downsides.

In summary, I do not think these changes benefit the project. I hope you understand that is why I cannot merge these changes.

Alberth289346:
> To my knowledge, the ExitException is used for preemptive purposes, meaning end-user errors.

The following lines are then dubious in `JPlag.java`
- `throw new ExitException("Illegal comparison mode: " + options.getComparisonMode());`
- `throw new ExitException("Language instantiation failed", ExitException.BAD_LANGUAGE_ERROR);`

I had seen both uses but didn't check yet how often it happens. It's not that often used for developers use apparently.

----
> `.collect(Collectors.toCollection(ArrayList::new))` cannot be replaced with `collect(toList())`, as `toList()` allows no assumption on the underlying data structure. However, for performance reasons we might want an `ArrayList`.

In that case, you may want to fix `filterInvalidSubmissions`, as it does use `toList()`.

----
I wasn't sure the second commit would be acceptable as it's a step not directly driven by implementation needs.
However, it's hard to discuss it without having a concrete example, so I took the plunge of adding it for at least having a discussion and possibly acceptance.

So, drop the 2nd commit?

tsaglam:
> The following lines are then dubious in `JPlag.java`

Yeah, they should probably be unchecked exceptions. The first should be an `IllegalArgumentException` and the second could be `IllegalArgumentException` or a `UnsupportedOperationException`, however, both are not an ideal fit. I don't think we need to fix this right now, it would be sufficient to make a dedicated issue for this.

> In that case, you may want to fix `filterInvalidSubmissions`, as it does use `toList()`.

I think for that it is fine, as usually there should not be a huge amount of invalid submissions, and even if so, they are not used for much anyways. The main processing is done on the valid submissions.

> I wasn't sure the second commit would be acceptable as it's a step not directly driven by implementation needs. However, it's hard to discuss it without having a concrete example, so I took the plunge of adding it for at least having a discussion and possibly acceptance.

Having a discussion is always possible, so no problem at all!

> So, drop the 2nd commit?

Yes, drop the second commit and remove the changes to the filtering from the first one.

tsaglam (approved):
This is now reduced to a tiny PR, but I think it is okay for this one as it started much bigger.

PR #247, opened 2021-12-01
Fix NPE on basecode run with empty submission folder.

Alberth289346:
See below the explanation for the testcase output crashing after I removed the fix in `Submission.java`.

There is one empty submission folder in the root directory (containing only a non-source file for git)
The case is a copy of the `basecode` tree, except for the additional `empty` sub directory.

It is correctly seen as not a submission ("nothing to parse"), but the submission is not flagged as error, and thus survives the filter. In the comparison code, the basecode comparison doesn't check for `null` token lists, and crashes on an NPE.

A non-basecode run works because the strategies do explicit checking on non-null token lists before attempting comparisons.

``` text
Initialized language Javac 1.9+ based AST plugin
Basecode directory "src/test/resources/de/jplag/samples/emptysubmission/base" will be used
------ Parsing submission: A
OK
------ Parsing submission: B
OK
------ Parsing submission: empty
ERROR: nothing to parse for submission "empty
ERROR -> Submission removed
2 submissions parsed successfully!
1 parser error!
0 too short submissions!
Total time for parsing: 0 sec
Time per parsed submission: 5 msec

----- Parsing basecode submission: base
Basecode submission parsed!
Time for parsing Basecode: 0 sec
Initialized language Javac 1.9+ based AST plugin
Initialized language Javac 1.9+ based AST plugin
Basecode directory "src/test/resources/de/jplag/samples/basecode/base" will be used
------ Parsing submission: A
OK
------ Parsing submission: B
OK
2 submissions parsed successfully!
0 parser errors!
0 too short submissions!
Total time for parsing: 0 sec
Time per parsed submission: 8 msec

----- Parsing basecode submission: base
Basecode submission parsed!
Time for parsing Basecode: 0 sec
Comparing A-B: 85.0

Total time for comparing submissions: 0 sec
Initialized language Javac 1.9+ based AST plugin
Tests run: 6, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.113 sec <<< FAILURE!
testEmptySubmission(de.jplag.BaseCodeTest)  Time elapsed: 0.038 sec  <<< ERROR!
java.lang.NullPointerException
	at de.jplag.GreedyStringTiling.swapAndCompare(GreedyStringTiling.java:99)
	at de.jplag.GreedyStringTiling.compareWithBaseCode(GreedyStringTiling.java:94)
	at de.jplag.strategy.AbstractComparisonStrategy.compareSubmissionsToBaseCode(AbstractComparisonStrategy.java:30)
	at de.jplag.strategy.NormalComparisonStrategy.compareSubmissions(NormalComparisonStrategy.java:23)
	at de.jplag.JPlag.run(JPlag.java:61)
	at de.jplag.TestBase.runJPlag(TestBase.java:34)
	at de.jplag.BaseCodeTest.testEmptySubmission(BaseCodeTest.java:19)
        ...
```

### DIFF
--- a/jplag/src/main/java/de/jplag/Submission.java
+++ b/jplag/src/main/java/de/jplag/Submission.java
@@ -154,6 +154,8 @@ public class Submission implements Comparable<Submission> {
     public boolean parse(boolean debugParser) {
         if (files == null || files.size() == 0) {
             errorCollector.print("ERROR: nothing to parse for submission \"" + name, null);
+            tokenList = null;
+            hasErrors = true; // invalidate submission
             return false;
         }
 
--- a/jplag/src/test/java/de/jplag/BaseCodeTest.java
+++ b/jplag/src/test/java/de/jplag/BaseCodeTest.java
@@ -15,6 +15,12 @@ public class BaseCodeTest extends TestBase {
     }
 
     @Test
+    public void testEmptySubmission() throws ExitException {
+        JPlagResult result = runJPlag("emptysubmission", it -> it.setBaseCodeSubmissionName("base"));
+        verifyResults(result);
+    }
+
+    @Test
     public void testAutoTrimFileSeparators() throws ExitException {
         JPlagResult result = runJPlag("basecode", it -> it.setBaseCodeSubmissionName(File.separator + "base" + File.separator));
         verifyResults(result);
--- a/jplag/src/test/resources/de/jplag/samples/emptysubmission/A/TerrainType.java
+++ b/jplag/src/test/resources/de/jplag/samples/emptysubmission/A/TerrainType.java
@@ -1,0 +1,27 @@
+package carcassonne.model.terrain;
+
+import java.util.List;
+
+/**
+ * Enumeration for the terrain type. Is used to specify the terrain of a tile on its different positions.
+ * @author Timur Saglam
+ */
+public enum TerrainType {
+    CASTLE,
+    ROAD,
+    MONASTERY,
+    FIELDS,
+    OTHER;
+
+    /**
+     * Generates a list of the basic terrain types, which is every terrain except {@link Other}.
+     * @return a list of CASTLE, ROAD, MONASTERY, FIELDS.
+     */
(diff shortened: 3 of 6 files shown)


jplag/src/main/java/de/jplag/Submission.java
```diff
@@ -154,6 +154,8 @@ public boolean hasErrors() {
     public boolean parse(boolean debugParser) {
         if (files == null || files.size() == 0) {
             errorCollector.print("ERROR: nothing to parse for submission \"" + name, null);
+            tokenList = null;
```
tsaglam, line 157: What is the benefit of setting the token list to null here?

tsaglam (approved):
Looks good, just one minor question.

jplag/src/main/java/de/jplag/Submission.java
Alberth289346: Not sure, but other code there does it too.

https://github.com/jplag/JPlag/blob/master/jplag/src/main/java/de/jplag/Submission.java#L167

https://github.com/jplag/JPlag/blob/master/jplag/src/main/java/de/jplag/Submission.java#L174

Might be an old convention though, code is changing rapidly.

What I wondered about is why does the comparison code check for `tokenList != null` when we are supposedly retrieving valid submissions from the submission set. It is likely also an old convention though.

It might very well be the case that the `tokenList` check was the original way to denote invalid, and `error` was added at a later date for some reason.

tsaglam: I suspected that! I think in the long run the null checks in the strategies should probably be assertions or we should remove them and the submissions set should assert that all valid submission sets have non-null token lists. Then we can also remove setting null here. For now, let's keep it consistent, as you did here with setting it to null.

I will make an issue for that null-checks, null-setting, and error duplication!

tsaglam: Documented in #249!

PR #250, opened 2021-12-05
Add logger to the project

### LINKED ISSUE #117: Add logging library and rethink error collection
Instead of direct prints to the console, a logging library shall be used.

EDIT:

- [x] Add logging library
- [x] Replace all occurence of Syso/Syserr
- [ ] Remove error consumer interface and rethink error collection

dfuchss:
Resolves #117 (Simply added slf4j)
Feel free to modify the different log levels.

Do not use Log4j .. that causes test fails ..


dfuchss:
@tsaglam @sebinside 
I've used `find . -name '*.java' ! -type d -exec bash -c 'expand -t 4 "$0" > /tmp/e && mv /tmp/e "$0"' {} \;` to replace all tabs in Java Code files by 4 spaces.

dfuchss:
@tsaglam @sebinside I've added an initial version of a formatter to this branch. Maybe just adapt it as you like. Afterwards, I can apply it to my code again. It seems that the tabs/spacing was not consistent :(

dfuchss:
Maybe that diff makes it easier :) (diff -w ignores whitespace changes :) )
[feature_logger.diff.txt](https://github.com/jplag/JPlag/files/7663454/feature_logger.diff.txt)

Alberth289346:
Your text diff is full of reformatting changes. Don't do that, it's not readable now.
For the actual patch, maybe you can wrap logging into the `ErrrorConsumer` so you don't need to touch all the parsing frontends.

I started reading here as I recently read about an exploit in log4j, and was curious whether that had been addressed here.
More information is at https://logging.apache.org/log4j/2.x/
```
The Log4j team has been made aware of a security vulnerability, CVE-2021-45046,
that has been addressed in Log4j 2.12.2 for Java 7 and 2.16.0 for Java 8 and up.
```
So it would seems it needs at least `2.16.0`, rather than the `2.14.1` that you added.

tsaglam:
>  For the actual patch, maybe you can wrap logging into the ErrrorConsumer so you don't need to touch all the parsing frontends.

We already discussed that internally, this PR is subject to change.

> I started reading here as I recently read about an exploit in log4j, and was curious whether that had been addressed here.

This was addressed in fe03f58, however, 2.16.0 was released after that.

dfuchss:
@tsaglam @sebinside now it's easy to read :)

dfuchss:
But I have to fix some tests .. :)

dfuchss:
Fun fact of the day:
If I use log4j the tests fail because there is no error output while compiling.
If I simply change to slf4j logging tests pass again

dfuchss:
Closed because of #409

PR #255, opened 2021-12-12
Add global basecode option

Alberth289346:
Dropping `-bc` directory and adding `-g` path for the global basecode submission directory.
If available, the basecode gets parsed before the user submission parsing, and each of the latter gets the basecode submission.

Also changed reporting to use its own `JplagOptions` instance and changed a `HashSet` variable type to its interface class.

### DIFF
--- a/jplag/src/main/java/de/jplag/SubmissionSetBuilder.java
+++ b/jplag/src/main/java/de/jplag/SubmissionSetBuilder.java
@@ -5,9 +5,11 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 
@@ -57,20 +59,35 @@ public class SubmissionSetBuilder {
         // Extract the basecode submission if necessary.
         Optional<Submission> baseCodeSubmission = Optional.empty();
         if (options.hasBaseCode()) {
-            verifyBaseCodeName();
+            Submission baseCode = tryLoadBaseCodeAsPath(rootDirectory);
+            if (baseCode == null) {
+                baseCode = tryLoadBaseCodeAsRootSubDirectory(foundSubmissions);
 
-            String baseCodeName = options.getBaseCodeSubmissionName();
-            String baseCodePath = rootDirectory.toString() + File.separator + baseCodeName;
-
-            // Grab the basecode submission and erase it from the regular submissions.
-            Submission foundBaseCodeSubmission = foundSubmissions.get(baseCodeName);
-            if (foundBaseCodeSubmission == null) {
-                throw new BasecodeException("Basecode submission \"" + baseCodePath + "\" doesn't exist!");
+                String baseCodeName = options.getBaseCodeSubmissionName().get();
+                if (baseCode == null) {
+                    // No base code found at all, report an error.
+                    throw new BasecodeException(
+                            String.format("Basecode path \"%s\" relative to the working directory could not be found.", baseCodeName));
+                } else {
+                    // Found a base code as a submission, report about obsolete usage.
+                    System.out.printf("Deprecated use of the -bc option found, please specify the basecode as \"%s%c%s\" instead.",
+                            rootDirectory.toString(), File.separator, baseCodeName);
+                }
             }
+            baseCodeSubmission = Optional.of(baseCode);
+            System.out.println(String.format("Basecode directory \"%s\" will be used.", baseCode.getRoot().toString()));
 
-            foundSubmissions.remove(baseCodeName);
-            baseCodeSubmission = Optional.of(foundBaseCodeSubmission);
-            System.out.println("Basecode directory \"" + baseCodePath + "\" will be used");
+            // Basecode may also be registered as a user submission. If so, remove the latter.
+            File baseCodeRoot = baseCode.getCanonicalRoot(); // Use canonical form for a more sane equality notion.
+            Iterator<Entry<String, Submission>> submIter = foundSubmissions.entrySet().iterator();
+            while (submIter.hasNext()) {
+                Entry<String, Submission> entry = submIter.next();
+                if (baseCodeRoot.equals(entry.getValue().getCanonicalRoot())) {
+                    submIter.remove();
+                    System.out.println(String.format("Skipping \"%s\" as user submission.", entry.getValue().getRoot().toString()));
+                    break;
+                }
+            }
         }
 
         // Merge everything in a submission set.
@@ -92,13 +109,75 @@ public class SubmissionSetBuilder {
     }
 
     /**
-     * Verify the basecode name.
+     * Try to load the basecode under the assumption of being a path.
+     * @return Base code submission if the option value can be interpreted as global path, else {@code null}.
+     * @throws ExitException when the option value is a path with errors.
      */
-    private void verifyBaseCodeName() throws ExitException {
-        String name = options.getBaseCodeSubmissionName();
+    private Submission tryLoadBaseCodeAsPath(File rootDirectory) throws ExitException {
+        String baseCodeName = options.getBaseCodeSubmissionName().get();
+        File basecodeSubmission = new File(rootDirectory, baseCodeName);
+        if (!basecodeSubmission.exists()) {
+            return null;
+        }
+
+        String errorMessage = isExcludedEntry(basecodeSubmission);
+        if (errorMessage != null) {
+            throw new BasecodeException(errorMessage); // Stating an excluded path as basecode isn't very useful.
+        }
+
+        try {
+            return processDirEntry(basecodeSubmission);
+
+        } catch (SubmissionException ex) {
+            throw new BasecodeException(ex.getMessage()); // Change thrown exception to basecode exception.
+
+        } catch (ExitException ex) {
+            throw ex;
+        }
+    }
+
+    /**
+     * Try to load the basecode under the assumption of being a sub-directory in the user submissions directory.
+     * @return Base code submission if the option value can be interpreted as a sub-directory, else {@code null}.
+     * @throws ExitException when the option value is a sub-directory with errors.
+     */
+    private Submission tryLoadBaseCodeAsRootSubDirectory(Map<String, Submission> foundSubmissions) throws ExitException {
+        String baseCodeName = options.getBaseCodeSubmissionName().get();
+
+        // Is the option value a single name after trimming spurious separators?
+        String name = trimLeft(trimRight(baseCodeName, File.separatorChar), File.separatorChar);
+        if (name.contains(File.separator)) return null; // Not a single name, cannot be a sub-directory.
+
         if (name.contains(".")) {
             throw new BasecodeException("The basecode directory name \"" + name + "\" cannot contain dots!");
         }
+
+        // Grab the basecode submission from the regular submissions.
+        return foundSubmissions.get(baseCodeName);
+    }
+
+    /**
+     * Trim characters k from the left of the string.
+     */
+    private String trimLeft(String s, char k) {
+        int i = 0;
+        while (i < s.length() && s.charAt(i) == k) {
+            i++;
+        }
+        if (i == 0) return s;
+        return s.substring(i);
+    }
+
+    /**
+     * Trim characters k from the right of the string.
+     */
+    private String trimRight(String s, char k) {
+        int i = s.length() - 1;
+        while (i >= 0 && s.charAt(i) == k) {
+            i--;
+        }
+        if (i == s.length() - 1) return s;
+        return s.substring(0, i + 1);
     }
 
     /**
@@ -126,6 +205,51 @@ public class SubmissionSetBuilder {
     }
 
     /**
+     * Check that the given submission entry is not invalid due to exclusion names or bad suffix.
+     * @param submissionEntry Entry to check.
+     * @return Error message if the entry should be ignored.
+     */
+    private String isExcludedEntry(File submissionEntry) {
+        if (isFileExcluded(submissionEntry)) {
+            return "Exclude submission: " + submissionEntry.getName();
+        }
+
+        if (submissionEntry.isFile() && !hasValidSuffix(submissionEntry)) {
+            return "Ignore submission with invalid suffix: " + submissionEntry.getName();
+        }
+        return null;
+    }
+
+    /**
+     * Process the given directory entry as a submission root, the path MUST not be excluded.
+     * @param submissionEntry Entry to process.
+     * @return The entry converted to a submission.
+     * @throws ExitException when an error has been found with the entry.
+     */
+    private Submission processDirEntry(File submissionEntry) throws ExitException {
+        if (isExcludedEntry(submissionEntry) != null) {
+            throw new AssertionError("Pre-condition of non-exclusion is violated.");
+        }
+
+        String fileName = submissionEntry.getName();
+        if (submissionEntry.isDirectory() && options.getSubdirectoryName() != null) {
+            // Use subdirectory instead
+            submissionEntry = new File(submissionEntry, options.getSubdirectoryName());
+
+            if (!submissionEntry.exists()) {
+                throw new SubmissionException(
+                        String.format("Submission %s does not contain the given subdirectory '%s'", fileName, options.getSubdirectoryName()));
+            }
+
+            if (!submissionEntry.isDirectory()) {
+                throw new SubmissionException(String.format("The given subdirectory '%s' is not a directory!", options.getSubdirectoryName()));
+            }
+        }
+
+        return new Submission(fileName, submissionEntry, parseFilesRecursively(submissionEntry), language, errorCollector);
+    }
+
+    /**
      * Process entries in the root directory to check whether they qualify as submissions.
      * @param rootDirectory Root directory being examined.
      * @param fileNames Entries found in the root directory.
@@ -137,33 +261,15 @@ public class SubmissionSetBuilder {
         for (String fileName : fileNames) {
             File submissionFile = new File(rootDirectory, fileName);
 
-            if (isFileExcluded(submissionFile)) {
-                System.out.println("Exclude submission: " + submissionFile.getName());
+            String errorMessage = isExcludedEntry(submissionFile);
+            if (errorMessage != null) {
+                System.out.println(errorMessage);
                 continue;
             }
 
-            if (submissionFile.isFile() && !hasValidSuffix(submissionFile)) {
-                System.out.println("Ignore submission with invalid suffix: " + submissionFile.getName());
-                continue;
-            }
-
-            if (submissionFile.isDirectory() && options.getSubdirectoryName() != null) {
-                // Use subdirectory instead
-                submissionFile = new File(submissionFile, options.getSubdirectoryName());
-
-                if (!submissionFile.exists()) {
-                    throw new SubmissionException(
-                            String.format("Submission %s does not contain the given subdirectory '%s'", fileName, options.getSubdirectoryName()));
-                }
-
-                if (!submissionFile.isDirectory()) {
-                    throw new SubmissionException(String.format("The given subdirectory '%s' is not a directory!", options.getSubdirectoryName()));
-                }
-            }
-
-            Submission submission = new Submission(fileName, submissionFile, parseFilesRecursively(submissionFile), language, errorCollector);
-            foundSubmissions.put(fileName, submission);
+            foundSubmissions.put(fileName, processDirEntry(submissionFile));
         }
+
         return foundSubmissions;
     }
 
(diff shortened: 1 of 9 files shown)


Alberth289346:
First of the patches covering #238 and beyond.

tsaglam:
Without looking at the PR in detail: As commented [here](https://github.com/jplag/JPlag/pull/238#issuecomment-992308915), would prefer the `-bc` as argument name, even if that changes how `-bc` works.

Also maybe we can discuss this here:

> We could implement in a way that the basecode is specified globally with -bc (`path/to/base`), however, if it is just a single dir name (`base`) which does not resolve globally and one of the root dirs contains a submission with that name then we could show a warning and just use this submission. As a consequence, a user specifying the basecode locally will still execute correctly but will receive a warning.

This would mean moving `-bc` to global, but having this inbuilt fallback. The mentioned warning would be something like "_It seems the basecode was specified by name, thus submission XXX was used as basecode. Note that the basecode should be specified globally._"

Alberth289346:
Ah sorry, missed that. Will fix.

Alberth289346:
So while I said I would fix the `-g` option to `-bc` as you proposed, I am not sure you actually want it that way. Let's discuss a case.
Directory structure:
``` text
mycourse/
    /studentA/...
    /studentB/...
    /base/...
```
Course with 2 submissions by students and a basecode directory.
Your proposal is that `jplag -bc base mycourse` will work with a fall-back notice that the user should specify a better path.

In this patch, base code parsing is separate from user submission parsing. The basecode parsing parses `mycourse/base` due to the recognizing the 'old' use. The user submission is unaware of basecode selections (they are not expected to be in a root directory in my current implementation), so  it simply iterates over all directories in `mycourse` and finds 3 submissions where the `base` submission will give 100% match with the basecode, or equivalently, 0% plagiarism by definition. Therefore, it won't show in the results, but a user likely doesn't expect to get 3 user submissions as answer.

We can add code to skip the `mycourse/base`, but with more root directories that gets more tricky. Adding the results of the previous course as an example:
``` text
previouscourse/
    /studentA/...
    /studentB/...
    /base/...
```
Previous year also has 2 students and `base`. This `previouscourse/base` entry is not chosen as basecode submission in the implementation, so it's a regular user submission now, and we have 3 user submissions too here (instead of 2 in the old interpretation of `-bc`).
Very likely there is a large overlap between `previouscourse/base` and `mycourse/base`. For arguments' sake let's say 95% overlap, and 5% what is not in `mycourse/base` (ie I changed 5% of the basecode since the previous course). That 5% will now count as plagiarism in all user submissions (including the entries in `previouscourse`.

While the amount of plagiarism found in this way is likely too small to ever get noticed, it feels like the wrong solution.
"Use `mycourse/base'" like you proposed doesn't seem sufficient. Your proposal could be extended to ignoring the `base` entry in all root-directories in case of fall-back, so they don't get interpreted as user submission.

Experimental implementation breaks if `base` is missing in a root directory, but I am guessing that can be silently ignored here.

tsaglam:
> We can add code to skip the `mycourse/base`, but with more root directories that gets more tricky. 

Of course, that is definitely more tricky there.

> Your proposal could be extended to ignoring the `base` entry in all root-directories in case of fall-back, so they don't get interpreted as user submission.

That would be one option, however, I have another idea. First, my reasoning on which the idea is based:
My fallback is mainly targeted at legacy usage of the changed CLI with global basecode. However, multiple root dirs will also be a new feature, so most people still specifying basecode in a non-global fashion will most likely not use multiple root dirs. If they do, they might realize themselves along the lines of: "Out of which root dir will my basecode be taken?"

Thus here is the solution I envision:

- If basecode could be found as it is correctly globally found, use it
- if it is not found, one root dir is used, and it contains the basecode (as it is specified locally)
    - show a warning
    - use this local base code
    - on submission parsing check for each submission `if submission != basecode`, filter out the basecode as we do with invalid submissions
- if it is not found, and more than one root dir is used, show error (BaseCodeException)
    - (because we cannot know which root dir is meant, multiple might contain a submission that matches)
    - (also, in these cases the multiple root dir feature is used, meaning the old usage of JPlag was already adapted. Especially thinking f programmatic use of JPlag.)

In short, this means the fallback only work for one root dir and _iff_ matching submission exists. 
Yes, we need to ignore the basecode as a normal submission, however, that is something we might need to do with globally specified basecode as well, as a user could correctly specify a global path to a basecode that lies in one of the root dirs (e.g. `-bc ./mycourse/base`). Then need the `submission != basecode` check anyways.

What do you think about that solution?

tsaglam:
Just a small note: I will be out-of-office from tomorrow until January the 10th, thus I will be less active in the meantime.

Alberth289346:
Rebuild of the patch on top of the reworked submission set building code.

Alberth289346:
Fixed the problems. Also, due to using Java-based resources in tests, relative paths and dots in paths cannot be tested.

tsaglam:
What is the current state of this PR? Did your changes reflect my [comment above](https://github.com/jplag/JPlag/pull/255#issuecomment-994665789)? (Sorry, I am a bit out of the loop here)

Alberth289346:
> What is the current state of this PR? Did your changes reflect my [comment above](https://github.com/jplag/JPlag/pull/255#issuecomment-994665789)? (Sorry, I am a bit out of the loop here)

No problem, that happens when you take a break :)

----
I coded what you proposed, except there are no multiple root directories yet due to adding only one feature at a time. Your 3rd point thus never happens currently.

More precise semantics of the `-bc` value are described in `JplagOptions`. (Mostly taken from previous limitations on the `-bc` value.)
Processing of `-bc` value can be found at `SubmissionSetBuilder`, new line 67 and further. The changes are

- All submissions are loaded first (new line 62). You need them anyway, either as user submission or as basecode.
- If global path finding fails for basecode, it falls back to finding a submission with the same name from the loaded submissions.
- Failing to find a loaded submission gives an error that -bc is not a global path that exists (since that is what the help text says -bc is doing).
- Finding a loaded submission from the submission set gives a deprecation warning.

- Removing base code from the submission set is then the next step at line 85. This means that even if you specify a global path to the basecode entry in the submission directory, it will be caught and removed.

I also added some tests. I did run into limits of Java package trees being used as file system though. (Iirc a relative path in a user-submission directory  fails to resolve globally, so the class loader starts throwing exceptions. Also tests with 'weird' characters such as dots don't work.)

Last but not least, I also wandered into the next step a bit, allow having multiple roots. As you wanted all submissions in one set, a `Map<String, Submission>` isn't going to fly for storing all submissions then, since different roots may have the same entry name. That means a different way of storing submissions will be needed where 1 root is then a special case (as for one root the `-bc` fallback must be allowed). Haven't yet done any thinking on that, in particular together with finding a basecode entry in the sea of submissions. Looks solvable to me though.

tsaglam:
> I also added some tests. I did run into limits of Java package trees being used as file system though. (Iirc a relative path in a user-submission directory fails to resolve globally, so the class loader starts throwing exceptions. Also tests with 'weird' characters such as dots don't work.)

Do I understand correctly that these are issues with the test framework and not with JPlag itself?

> As you wanted all submissions in one set, a `Map<String, Submission>` isn't going to fly for storing all submissions then, since different roots may have the same entry name. That means a different way of storing submissions will be needed where 1 root is then a special case (as for one root the `-bc` fallback must be allowed). Haven't yet done any thinking on that, in particular together with finding a basecode entry in the sea of submissions. Looks solvable to me though.

But root ID + submission name (with root ID being either the name of the root folder or maybe even the path) should be sufficient, right?

Alberth289346:
Rebased.

> Do I understand correctly that these are issues with the test framework and not with JPlag itself?

Yes. To check what I was doing I ran the following examples (tests were already checked by `mvn` while building the jar).
``` sh
$ java -jar jplag/target/jplag-3.1.0-SNAPSHOT-jar-with-dependencies.jar jplag/src/test/resources/de/jplag/samples/basecode

Initialized language Javac 1.9+ based AST plugin
JPlag initialized
3 submissions parsed successfully!
0 parser errors!
0 too short submissions!

Comparing A-B: 96.77419
Comparing A-base: 51.162792
Comparing B-base: 48.88889
```
It sees `base` as normal submission, since I didn't specify a `-bc` option.

``` sh
$ java -jar jplag/target/jplag-3.1.0-SNAPSHOT-jar-with-dependencies.jar -bc base  jplag/src/test/resources/de/jplag/samples/basecode 
Initialized language Javac 1.9+ based AST plugin
JPlag initialized
Deprecated use of the -bc option found, please specify the basecode as "jplag/src/test/resources/de/jplag/samples/basecode/base" instead.
Basecode directory "jplag/src/test/resources/de/jplag/samples/basecode/base" will be used.
Skipping "jplag/src/test/resources/de/jplag/samples/basecode/base" as user submission.
2 submissions parsed successfully!
0 parser errors!
0 too short submissions!

----- Parsing basecode submission: base
Basecode submission parsed!
Comparing A-B: 85.0
```
`base` is now seen as basecode, note the deprecation line.

``` sh
$ java -jar jplag/target/jplag-3.1.0-SNAPSHOT-jar-with-dependencies.jar -bc jplag/src/test/resources/de/jplag/samples/basecode/base  jplag/src/test/resources/de/jplag/samples/basecode
Initialized language Javac 1.9+ based AST plugin
JPlag initialized
Basecode directory "jplag/src/test/resources/de/jplag/samples/basecode/base" will be used.
Skipping "jplag/src/test/resources/de/jplag/samples/basecode/base" as user submission.
2 submissions parsed successfully!
0 parser errors!
0 too short submissions!

----- Parsing basecode submission: base
Basecode submission parsed!
Comparing A-B: 85.0
```
Specifying `base` as a path from the current directory, again `base` is not a user submission.

> But root ID + submission name (with root ID being either the name of the root folder or maybe even the path) should be sufficient, right?

That is a good idea, I can use the canonical path name as key, thanks!

And a question, do you mind if I change the following?
``` diff
diff --git a/jplag/src/main/java/de/jplag/JPlag.java b/jplag/src/main/java/de/jplag/JPlag.java
index 291c00c..77e2f1e 100644
--- a/jplag/src/main/java/de/jplag/JPlag.java
+++ b/jplag/src/main/java/de/jplag/JPlag.java
@@ -94,10 +94,11 @@ public class JPlag {
     }
 
     private ComparisonStrategy initializeComparisonStrategy(final ComparisonMode comparisonMode) {
-        return switch (comparisonMode) {
-            case NORMAL -> new NormalComparisonStrategy(options, coreAlgorithm);
-            case PARALLEL -> new ParallelComparisonStrategy(options, coreAlgorithm);
-        };
+        switch (comparisonMode) {
+            case NORMAL: return new NormalComparisonStrategy(options, coreAlgorithm);
+            case PARALLEL: return new ParallelComparisonStrategy(options, coreAlgorithm);
+            default: throw new AssertionError("Unknown comparison strategy found.");
+        }
     }
 
     private Language initializeLanguage() {
```
This was changed in 41f285c5 by Gabriel Laskar. My Java 11 doesn't like the "->" notation here, and perhaps more importantly, Gabriel deleted the `default` case, so a failure here to keep the enum names in sync with the definition will pop up as an NPE at wherever you try to apply the `NULL` strategy.

tsaglam:
> And a question, do you mind if I change the following?
> 
> [...]
>
> This was changed in [41f285c](https://github.com/jplag/JPlag/commit/41f285c56bffa11c1ea9ab4aec8e36bd0b68029b) by Gabriel Laskar. My Java 11 doesn't like the "->" notation here, and perhaps more importantly, Gabriel deleted the `default` case, so a failure here to keep the enum names in sync with the definition will pop up as an NPE at wherever you try to apply the `NULL` strategy.

Yes, you can change it. I mean especially since you restore the default case.
Tell me when you think this PR is ready to merge then I will review it in detail!

GabrielL:
> Rebased.
> 
> > Do I understand correctly that these are issues with the test framework and not with JPlag itself?
> 
> Yes. To check what I was doing I ran the following examples (tests were already checked by `mvn` while building the jar).
> 
> ```shell
> $ java -jar jplag/target/jplag-3.1.0-SNAPSHOT-jar-with-dependencies.jar jplag/src/test/resources/de/jplag/samples/basecode
> 
> Initialized language Javac 1.9+ based AST plugin
> JPlag initialized
> 3 submissions parsed successfully!
> 0 parser errors!
> 0 too short submissions!
> 
> Comparing A-B: 96.77419
> Comparing A-base: 51.162792
> Comparing B-base: 48.88889
> ```
> 
> It sees `base` as normal submission, since I didn't specify a `-bc` option.
> 
> ```shell
> $ java -jar jplag/target/jplag-3.1.0-SNAPSHOT-jar-with-dependencies.jar -bc base  jplag/src/test/resources/de/jplag/samples/basecode 
> Initialized language Javac 1.9+ based AST plugin
> JPlag initialized
> Deprecated use of the -bc option found, please specify the basecode as "jplag/src/test/resources/de/jplag/samples/basecode/base" instead.
> Basecode directory "jplag/src/test/resources/de/jplag/samples/basecode/base" will be used.
> Skipping "jplag/src/test/resources/de/jplag/samples/basecode/base" as user submission.
> 2 submissions parsed successfully!
> 0 parser errors!
> 0 too short submissions!
> 
> ----- Parsing basecode submission: base
> Basecode submission parsed!
> Comparing A-B: 85.0
> ```
> 
> `base` is now seen as basecode, note the deprecation line.
> 
> ```shell
> $ java -jar jplag/target/jplag-3.1.0-SNAPSHOT-jar-with-dependencies.jar -bc jplag/src/test/resources/de/jplag/samples/basecode/base  jplag/src/test/resources/de/jplag/samples/basecode
> Initialized language Javac 1.9+ based AST plugin
> JPlag initialized
> Basecode directory "jplag/src/test/resources/de/jplag/samples/basecode/base" will be used.
> Skipping "jplag/src/test/resources/de/jplag/samples/basecode/base" as user submission.
> 2 submissions parsed successfully!
> 0 parser errors!
> 0 too short submissions!
> 
> ----- Parsing basecode submission: base
> Basecode submission parsed!
> Comparing A-B: 85.0
> ```
> 
> Specifying `base` as a path from the current directory, again `base` is not a user submission.
> 
> > But root ID + submission name (with root ID being either the name of the root folder or maybe even the path) should be sufficient, right?
> 
> That is a good idea, I can use the canonical path name as key, thanks!
> 
> And a question, do you mind if I change the following?
> 
> ```diff
> diff --git a/jplag/src/main/java/de/jplag/JPlag.java b/jplag/src/main/java/de/jplag/JPlag.java
> index 291c00c..77e2f1e 100644
> --- a/jplag/src/main/java/de/jplag/JPlag.java
> +++ b/jplag/src/main/java/de/jplag/JPlag.java
> @@ -94,10 +94,11 @@ public class JPlag {
>      }
>  
>      private ComparisonStrategy initializeComparisonStrategy(final ComparisonMode comparisonMode) {
> -        return switch (comparisonMode) {
> -            case NORMAL -> new NormalComparisonStrategy(options, coreAlgorithm);
> -            case PARALLEL -> new ParallelComparisonStrategy(options, coreAlgorithm);
> -        };
> +        switch (comparisonMode) {
> +            case NORMAL: return new NormalComparisonStrategy(options, coreAlgorithm);
> +            case PARALLEL: return new ParallelComparisonStrategy(options, coreAlgorithm);
> +            default: throw new AssertionError("Unknown comparison strategy found.");
> +        }
>      }
>  
>      private Language initializeLanguage() {
> ```
> 
> This was changed in [41f285c](https://github.com/jplag/JPlag/commit/41f285c56bffa11c1ea9ab4aec8e36bd0b68029b) by Gabriel Laskar. My Java 11 doesn't like the "->" notation here, and perhaps more importantly, Gabriel deleted the `default` case, so a failure here to keep the enum names in sync with the definition will pop up as an NPE at wherever you try to apply the `NULL` strategy.

From my understanding the code was not pinned to java 11, thus the change to use a switch expression.

Alberth289346:
> Tell me when you think this PR is ready to merge then I will review it in detail!

It is, please review.

> From my understanding the code was not pinned to java 11, thus the change to use a switch expression.

I understand that GabrielL, but in this case it is not a big improvement (you didn't save a lot of code by switching to streaming), both alternatives are sort of equal, and it saves me a lot of trouble in getting compilable code.
(I am using Java 11, since it's an LTS version that is still being updated. Java 14 on the other hand is not an LTS version and has ended life since September 2020. It won't get updated for bugs.)

As for the `default` case, it may not look important to you, but consider the case where in some way your code would fail to match one of the cases and the function returns `null`. Then you store that value somewhere, go through a dozen other functions, and then in a totally unrelated package perhaps even in a different thread you attempt to use the returned value.
Obviously that will fail and throw an NPE.

In such a case, finding the root cause of that crash means reverse-searching where originally the `null` was coming from, back to the original thread, back through the dozen routines, back to finding all points that may have stored that value, and then eventually you find the code with (for example) a missing `case`. This reverse-searching may take anything from 2 seconds to 2 weeks, depending on how well the person trying to solve the NPE knows the code.

Now if you had kept the `default`, it would crash right at the spot where `null` is being created. Reverse searching like above is not needed, it literally takes 0 seconds to find the function.
I have been added such checks to my code for 30 years, and in my experience, these checks fail much more often than you think, I have saved myself a lot of time searching for causes in that way.

GabrielL:
> > Tell me when you think this PR is ready to merge then I will review it in detail!
> 
> It is, please review.
> 
> > From my understanding the code was not pinned to java 11, thus the change to use a switch expression.
> 
> I understand that GabrielL, but in this case it is not a big improvement (you didn't save a lot of code by switching to streaming), both alternatives are sort of equal, and it saves me a lot of trouble in getting compilable code. (I am using Java 11, since it's an LTS version that is still being updated. Java 14 on the other hand is not an LTS version and has ended life since September 2020. It won't get updated for bugs.)
> 

The only thing I am saying here is that the build is not locked on java 11, but into java 14. If the wanted java version is 11, it should be written as that in the build.

> As for the `default` case, it may not look important to you, but consider the case where in some way your code would fail to match one of the cases and the function returns `null`. Then you store that value somewhere, go through a dozen other functions, and then in a totally unrelated package perhaps even in a different thread you attempt to use the returned value. Obviously that will fail and throw an NPE.

With a switch expression, the NPE will trigger directly on the switch, that was the point of the change, ie, having the language properly check the enumeration.

Anyway, if the code needs to be build with java 11, let's enforce it in the build.

tsaglam:
> The only thing I am saying here is that the build is not locked on java 11, but into java 14. If the wanted java version is 11, it should be written as that in the build.

It is not pinned. This merely a concession from me as a maintainer to a frequent committer.
However, it is also not a big deal, as it is just one switch statement.

> With a switch expression, the NPE will trigger directly on the switch, that was the point of the change, ie, having the language properly check the enumeration.

It does not really matter if it triggers directly on the switch, as it does not provide a clear error message. Thus, for clarity's sake, having an explicit default case is preferable.

Finally, let us keep the discussion here on-topic to keep the PRs clean. 
Anything regarding LTS versions and version pinning can be continued in the [discussion section](https://github.com/jplag/JPlag/discussions/categories/general).

jplag/src/main/java/de/jplag/SubmissionSetBuilder.java
```diff
@@ -92,13 +109,75 @@ private void verifyRootdirExistence(File rootDir) throws ExitException {
...
+    /**
+     * Trim characters k from the right of the string.
+     */
+    private String trimRight(String s, char k) {
+        int i = s.length() - 1;
+        while (i >= 0 && s.charAt(i) == k) {
+            i--;
+        }
+        if (i == s.length() - 1) return s;
+        return s.substring(0, i + 1);
```
tsaglam, line 180: Three issues here:
- I prefer speaking variable names (sorry for being pedantic)
- These methods are overly generic but only ever used to remove the file separator
- They can be merged and simplified
Thus I propose replacing them with:
```java
private String trimFileSeparators(String text) {
    String trimmedText = text;
    while (trimmedText.startsWith(File.separator)) {
        trimmedText = trimmedText.substring(1);
    }
    while (trimmedText.endsWith(File.separator)) {
        trimmedText = trimmedText.substring(0, trimmedText.length() - 1);
    }
    return trimmedText;
}
````

**Review by tsaglam — CHANGES_REQUESTED** [review]
Looks good, just a few small issues.

File: jplag/src/main/java/de/jplag/SubmissionSetBuilder.java [hunk 2]
```diff
@@ -57,20 +59,35 @@ public SubmissionSet buildSubmissionSet() throws ExitException {
...
+                    System.out.printf("Deprecated use of the -bc option found, please specify the basecode as \"%s%s%s\" instead.\n",
+                            rootDirectory.toString(), File.separator, baseCodeName);
+                }
+            }
+            baseCodeSubmission = Optional.of(baseCode);
+            System.out.println(String.format("Basecode directory \"%s\" will be used.", baseCode.getRoot().toString()));
+
+            // Basecode may also be registered as a user submission. If so, remove the latter.
+            File baseCodeRoot = baseCode.getCanonicalRoot(); // Use canonical form for a more sane equality notion.
+            Iterator<Entry<String, Submission>> submIter = foundSubmissions.entrySet().iterator();
```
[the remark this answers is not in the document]
Alberth289346: Changed. `Iterator` will disappear again in the multi-root patch though.

jplag/src/main/java/de/jplag/SubmissionSetBuilder.java
Alberth289346: Yeah, I assumed Java would have a basic trim functions but I was wrong. Then I got stuck on that road implementing them myself.

Fixed by inlining your suggestion in the calling function.

tsaglam (approved):
Looks good now!

Issue #213, opened 2021-10-27
Maven build shows error "no assembly descriptors found"


tsaglam:
When executing `mvn clean generate-sources package assembly:single` the build succeeds, all jars are built successfully. However, maven shows an error message:

```
[INFO] JPlag Aggregator 0.0.1-SNAPSHOT .................... FAILURE [  0.335 s]
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.2-beta-5:single (default-cli) on project aggregator: Error reading assemblies: No assembly descriptors found. -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.2-beta-5:single (default-cli) on project aggregator: Error reading assemblies: No assembly descriptors found.
```
The assembly plugin in the [pom parent](https://github.com/jplag/JPlag/blob/master/jplag.parent/pom.xml):
```xml
      <plugin>
        <artifactId>maven-assembly-plugin</artifactId>
        <version>3.3.0</version>
        <configuration>
          <descriptorRefs>
            <descriptorRef>jar-with-dependencies</descriptorRef>
          </descriptorRefs>
          <archive>
            <manifest>
              <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
              <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
            </manifest>
          </archive>
        </configuration>
      </plugin>
```

The assembly plugin in the [core project](https://github.com/jplag/JPlag/blob/master/jplag/pom.xml):
```xml
      <plugin>
        <artifactId>maven-assembly-plugin</artifactId>
        <version>3.3.0</version>
        <configuration>
          <descriptorRefs>
            <descriptorRef>jar-with-dependencies</descriptorRef>
          </descriptorRefs>
          <archive>
            <manifest>
              <mainClass>de.jplag.CLI</mainClass>
              <packageName>jplag</packageName>
            </manifest>
          </archive>
        </configuration>
      </plugin>
```

tsaglam:
Closed due to #256, no longer an issue.

PR #256, opened 2021-12-13
Reworked pom.xml to ensure easy build

dfuchss:
I've updated the overall structure (only the pom.xml files) to provide an easy build.

Details:
* I've moved the dependency management to the aggregator plugin & replaced the parent module (by the aggregator)
* I've aligned all submodules to one version since this makes it more naturally to use as dependency.
* I've fixed all versions of plugins & deps in the new "parent"
* I've replaced the old javacc libraries by https://github.com/phax/ph-javacc-maven-plugin (this version seems to be newer and also maintained; source code gets generated as well)
  * https://mvnrepository.com/artifact/org.codehaus.mojo/javacc-maven-plugin
  * https://mvnrepository.com/artifact/com.helger.maven/ph-javacc-maven-plugin

During my refactorings, I found out that your modules are using different antlr plugins:
```
<plugin>
	<groupId>org.codehaus.mojo</groupId>
	<artifactId>antlr-maven-plugin</artifactId>
	<version>2.1</version>
</plugin>
<plugin>
	<groupId>org.antlr</groupId>
	<artifactId>antlr4-maven-plugin</artifactId>
	<version>4.9.3</version>
</plugin>
```

Please share your thoughts about the structure. Actually the benefit is that you can simply run "mvn clean package" to create all jars and build the whole project :)


tsaglam:
Thanks for the PR, I assume this PR will passively resolve #213 due to the restructuring.
Just a few questions:

- is `mvn clean install` still required? If so we should adapt the readme.
- Regarding the multiple versions of the maven antlr plugin I assume different legacy frontends once used different versions of antlr, however, due to #242 some of these dependencies might be superfluous. Or are they still required?

tsaglam (approved):
Looks good to me, definitely way cleaner now!

dfuchss:
ad 1: you don't need `mvn install` anymore. A simple `mvn package` does the job .. I'll change the readme

ad 2: I'll have a look at it again.

dfuchss:
C# and Python3 are using different antlr plugins.

PR #263, opened 2021-12-15
Refactor submission set building

Alberth289346:
With the additional variations on `-bc` in #255 we need case distinctions, but the various pieces of information that we need seemed to be scattered between `JPlag.java` and `SubmissionSetBuilder.java`.

There didn't seem tot be an easy path for this, as decision making and checking is scattered between functions. So I concluded we could use a single point with all available information, simplifying the decision process.  
That spot ended up to be `SubmissionSetBuilder.buildSubmissionSet`. That function now does all error checking (including what was in `JPlag`), and finding and checking submissions. Near the end, it then finds the base code submission (which is at that time then already found to be a valid submission).
The more low-level code has been moved to other functions to avoid getting lost in the details.

For reviewing, this patch shouldn't do functional changes.
You may want to read the new `SubmissionSetBuilder` class by itself rather than trying to understand the diff, as it was heavily modified.

I haven't tried putting #255 on top of this, but I am pretty sure the job will less difficult than before.

### DIFF
--- a/jplag/src/main/java/de/jplag/JPlag.java
+++ b/jplag/src/main/java/de/jplag/JPlag.java
@@ -1,12 +1,9 @@
 package de.jplag;
 
-import java.io.File;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 
-import de.jplag.exceptions.BasecodeException;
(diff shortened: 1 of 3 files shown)


tsaglam (approved):
Looks good to me!

PR #272, opened 2022-01-17
Allow external frontend development

GabrielL:
This patchset aims to simplify a bit the Language loading facilities, and allow external language providers.

## Introducing SPI for Language frontends

It introduces a new interface, [`de.jplag.LanguageProvider`](https://github.com/GabrielL/JPlag/blob/cc51c80d36fdf950c7b24a706796695e140a821b/jplag.frontend-utils/src/main/java/de/jplag/LanguageProvider.java) (in frontend-utils) that is implemented by every language frontend.

For example, for [java](https://github.com/GabrielL/JPlag/blob/cc51c80d36fdf950c7b24a706796695e140a821b/jplag.frontend.java/src/main/java/de/jplag/java/JavaLanguageProvider.java) :

```java
@MetaInfServices
public class JavaLanguageProvider implements LanguageProvider {
    public static final String NAME = de.jplag.java.Language.NAME;

    @Override
    public Language create(final ErrorConsumer errorCollector) {
        return new de.jplag.java.Language(errorCollector);
    }

    @Override
    public String getDisplayName() {
        return NAME;
    }
}
```

With this, we can use the [Java Service Loader facilities](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) to discover frontend in the class path, instead of having to add them inside the main repository.

All the loading, and usage of the Service Loader is located inside [`de.jplag.Languages`](https://github.com/GabrielL/JPlag/blob/cc51c80d36fdf950c7b24a706796695e140a821b/jplag/src/main/java/de/jplag/Languages.java)

In order to be discoverable, this also adds a `META-INF/services/de.jplag.LanguageProvider` file inside the jar (example for java: https://github.com/GabrielL/JPlag/blob/cc51c80d36fdf950c7b24a706796695e140a821b/jplag.frontend.java/src/main/resources/META-INF/services/de.jplag.LanguageProvider)

## Refactoring and other changes

In order to make these changes, some refactoring were done:

* simplification of options/language usage in report, both are now directly passed to the Report constructor.
* removal of `language` in `JPlagOption`, options are now nearly read only.
* removal of `JPlagOptions` in `JPlagResult`
* removal of `LanguageOption`, its work is covered by `Languages` and `LanguageProvider` SPI.
* small refactoring in `JPlag` class, mostly making all fields final where possible

## Build Changes

In order to have working service providers, jars must have a `META-INF/services` directory with all the services exposed.

For the assembly jar with all dependencies, there is some changes to be able to merge all the `META-INF/services` files inside all the frontend jars.


dfuchss:
@GabrielL seems to be a nice enhancement.

One suggestion: 
If you are using [ServiceLoader](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/ServiceLoader.html) (and you are using it :) ), you might also use [MetaInfServices](https://mvnrepository.com/artifact/org.kohsuke.metainf-services/metainf-services/1.8) (see also: [Docs](http://metainf-services.kohsuke.org/)). Actually it generates the `META-INF/services/de.jplag.LanguageProvider` automatically :)

GabrielL:
> @GabrielL seems to be a nice enhancement.
> 
> One suggestion: If you are using [ServiceLoader](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/ServiceLoader.html) (and you are using it :) ), you might also use [MetaInfServices](https://mvnrepository.com/artifact/org.kohsuke.metainf-services/metainf-services/1.8) (see also: [Docs](http://metainf-services.kohsuke.org/)). Actually, it generates the `META-INF/services/de.jplag.LanguageProvider` automatically :)

I didn't know this one, this is nicer like this, updated, thanks!

jplag/src/main/assembly/jar-with-deps-merge-services.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<assembly xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.0 http://maven.apache.org/xsd/assembly-1.1.0.xsd">
+  <!-- copied from jar-with-dependencies (http://maven.apache.org/plugins/maven-assembly-plugin/descriptor-refs.html#jar-with-dependencies) -->
+  <id>jar-with-deps-merge-services</id>
```
tsaglam, line 6: Jar name should still be `jar-with-dependencies` so adapt accordingly if required.

jplag/src/main/java/de/jplag/JPlag.java
```diff
@@ -100,24 +106,7 @@ private ComparisonStrategy initializeComparisonStrategy(final ComparisonMode com
...
-            System.out.println("Initialized language " + language.getName());
-            return language;
-        } catch (NoSuchMethodException | SecurityException | ClassNotFoundException | InstantiationException | IllegalAccessException
-                | IllegalArgumentException | InvocationTargetException e) {
-            e.printStackTrace();
-            throw new IllegalStateException("Language instantiation failed:" + e.getMessage());
-        }
-
+    private Language loadLanguage(final ErrorCollector errorCollector, final String classPath) {
+        return Languages.provider(classPath).create(errorCollector);
```
tsaglam, line 110: This became so short, there might not be a need for a dedicated method.

jplag/src/main/java/de/jplag/Languages.java
```diff
@@ -0,0 +1,38 @@
...
+                .filter(p -> p.getDisplayName().equals(languageName))
+                .findAny()
+                .orElseThrow(() -> new LanguageProviderNotFoundException(languageName));
+    }
+
+    public static class LanguageProviderNotFoundException extends ProviderNotFoundException {
+        public LanguageProviderNotFoundException(final String languageName) {
+            super(String.format("Language Provider for language %s not found", languageName));
+        }
+    }
```
tsaglam, line 37: Please avoid nested classes!

README.md
```diff
@@ -63,7 +63,7 @@ named arguments:
 The new API makes it easy to integrate JPlag's plagiarism detection into external Java projects:
 
 ```java
-JPlagOptions options = new JPlagOptions("/path/to/rootDir", LanguageOption.JAVA);
+JPlagOptions options = new JPlagOptions("/path/to/rootDir", Languages.DEFAULT_LANGUAGE);
```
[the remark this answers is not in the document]
tsaglam: Actually, I am now wondering how you would use the JPlag API during 3rd party integration.
Imagine an application that depends on JPlag and only ever runs it with a specific language.
Right now, how would you specify here to run JPlag with C++?

tsaglam (changes requested):
There are some issues that need to be adressed.

dfuchss (changes requested):
Instead of using the xml to create the assembly, you may have a look at https://maven.apache.org/plugins/maven-shade-plugin/examples/resource-transformers.html#ServicesResourceTransformer

jplag/src/main/assembly/jar-with-deps-merge-services.xml
GabrielL: Since there is already a descriptor `jar-with-dependencies` I can't reuse the name, hence the changed name.

Initially I wanted to avoid adding a new dependency into the build, just to do that, but I'll look into shade to come up with something else.

jplag/src/main/java/de/jplag/JPlag.java
GabrielL: This is the factory, I know this is short, but the goal here is to document the line, and hide construction details.

I will try to find another way to do that, like integrate it directly into `Languages`

README.md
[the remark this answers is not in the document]
tsaglam: I am not a big fact that the users of the API need to enter random strings. This was one of the reasons for the language options enum. Maybe we can restore the enum (without the classpath field of course), just to give the users some options here. Users can then still specify the string or they can just pick one of the enumerals in the language option enum. What do you think?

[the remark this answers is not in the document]
dfuchss: you may also define constants in the languages itself that can be used as keys. you could also replace the constant DefaultLanguage with JavaLanguageProvider.NAME

.github/workflows/maven.yml
```diff
@@ -31,6 +31,6 @@ jobs:
       uses: actions/upload-artifact@v2
       with:
         name: "JPlag"
-        path: "jplag/target/jplag-*-jar-with-dependencies.jar"
+        path: "jplag/target/jplag-*-jar-with-deps-merge-services.jar"
```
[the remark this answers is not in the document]
GabrielL: fixed using shade instead of the assembly plugin.

As a side effect, it is no longer necessary to call specifically `assembly:single` in order to have the uber jar, this is already included in the `package` phase

jplag/pom.xml
```diff
@@ -1,104 +1,109 @@
...
+					<shadedClassifierName>jar-with-dependencies</shadedClassifierName>
+				</configuration>
+				<executions>
+					<execution>
+						<phase>package</phase>
+						<goals>
+							<goal>shade</goal>
+						</goals>
+					</execution>
+				</executions>
```
GabrielL, line 60: The change is here.

It seems that the file was using CRLF lines instead of LF only

dfuchss, line 60: See above. I would not create that by default.

jplag/src/main/java/de/jplag/Languages.java
```diff
@@ -0,0 +1,39 @@
...
+                .map(LanguageProvider::getDisplayName)
+                .sorted()
+                .collect(Collectors.toList());
+    }
+
+    public static List<LanguageProvider> providers() {
+        List<LanguageProvider> services = new ArrayList<>();
+        var loader = ServiceLoader.load(LanguageProvider.class);
+        loader.forEach(services::add);
+        return services;
```
dfuchss, line 21: maybe throw an exception/ log an error if a name is duplicated.

jplag/src/main/java/de/jplag/reporting/Report.java
```diff
@@ -58,10 +61,18 @@ public void writeResult(JPlagResult result) throws ReportGenerationException {
...
     }
 
+    private JPlagOptions getOptions() {
+        return options;
+    }
+
+    private Language getLanguage() {
+        return language;
+    }
+
```
dfuchss, line 75: I think you should delete private getters. They are not needed.

pom.xml
```diff
@@ -96,6 +96,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
+        <version>3.3.0</version>
```
dfuchss, line 99: I would also remove the version of the standard maven plugins.
The current versions are inherited by the maven super pom. if there is no reason to fix them to a certain version.

pom.xml
```diff
@@ -108,6 +109,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-source-plugin</artifactId>
+        <version>3.2.1</version>
```
dfuchss, line 112: see above

pom.xml
```diff
@@ -120,6 +122,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-gpg-plugin</artifactId>
+        <version>3.0.1</version>
```
dfuchss, line 125: see above

pom.xml
```diff
@@ -183,19 +186,9 @@
...
-            <archive>
-              <manifest>
-                <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
-                <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
-              </manifest>
-            </archive>
-          </configuration>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-shade-plugin</artifactId>
+          <version>3.2.4</version>
```
dfuchss, line 191: see above

dfuchss (changes requested):
Nice work! :)

I've found some points for discussion and some points that maybe could be improved :)

pom.xml
GabrielL: maven is complaining about it during the build, so it seems that it needs/wants it. I don't know if it is really useful or not

jplag/src/main/java/de/jplag/reporting/Report.java
GabrielL: Yes, they are not needed. Their main usage here was during refactoring, I will inline them

.github/workflows/maven.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       run: mvn test -B
       
     - name: Build Assembly
-      run: mvn clean package assembly:single
+      run: mvn clean package
```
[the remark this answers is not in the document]
GabrielL: Why split the jar generation in two phases ? Here there is a known command `mvn package` that will build the needed jars, and no need to search the documentation for which plugin/command is used in this specific project.

jplag.frontend.chars/src/main/java/de/jplag/chars/Language.java
```diff
@@ -10,6 +10,7 @@
  * read in text files as characters
  */
 public class Language implements de.jplag.Language {
+    public static final String NAME = "char";
```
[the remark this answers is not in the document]
GabrielL: There is no documentation in all the Language class

jplag/src/main/java/de/jplag/Languages.java
GabrielL: Oh yes, good catch, now it is possible to have duplicated names

jplag/src/main/java/de/jplag/Languages.java
```diff
@@ -0,0 +1,39 @@
+package de.jplag;
+
+import java.nio.file.ProviderNotFoundException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+
+public class Languages {
```
GabrielL, line 9: documentation is missing

.github/workflows/maven.yml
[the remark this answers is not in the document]
dfuchss: Alternatively you could move that to a profile that you can control the build process. (If I invoke package I assume maven to build the library artifacts nothing more .. but that's mainly  personal expectation .. I will / we should have a look at the docs how they define the package phase )

I personally like the idea that 'package' build the necessary stuff. Here that would only be the libraries that are going to be deployed to maven central. For the complete jar. You could define the Profile "XYZ" that can be used or not by the users of JPlag.


That's something the maintainers should decide in the end :)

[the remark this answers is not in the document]
tsaglam: I second what @dfuchss said. With `package` we build the jars for maven central, with `assembly:single` we build the fat jar. In general, details regarding the process should be left untouched by PRs if possible. Meaning how to build JPlag, how the JAR is named, and even how to use JPlag itself. If something _needs_ to be changed, it should be declared explicitly in the PR description.

jplag.frontend.chars/src/main/java/de/jplag/chars/Language.java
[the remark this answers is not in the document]
dfuchss: Yes, but that are methods that are inherited from the `de.jplag.Language`.
You introduce a new public constant. Therefore, (in my opinion) its purpose should be documented :)

tsaglam:
There are conflicts that need to be resolved before we can proceed. Also, please mark all suggestions as resolved if you addressed them.

tsaglam:
Closed due to lack of activity, happy to re-open anytime.

PR #275, opened 2022-01-19
refactor: move non generic frontend-utils code into their respective frontends

GabrielL:
in frontend-utils, there is some classes that are only used by a specific frontend:

* NewlineStream is used by cpp
* UnicodeReader by csharp
* InputState & ParserToken only by text

This small change move them out of frontend-utils and into their respective packages


tsaglam (approved):
Looks good, thanks for the contribution!

Issue #274, opened 2022-01-19
JPlagOptions are modified during the execution


tsaglam:
After passing an options instance to JPlag, some changes are made to the options instance.
This includes but is not limited to:
- Setting the language
- Setting language defaults
- Passing excluded file names

This is convoluting for new developers and may lead to maintenance issues further down the road. We should treat it as a record, and not modify option instances after they are configured. This would require storing parameters determined during the execution of JPlag separately.

GabrielL:
There is another issue with `JPLagOptions`: *everything* depends on it! most of the classes don't need the full options, only one or two fields. It would be better to pass them directly, to avoid the dependency. This will allow to test the separate pieces separately.

From what I see, the cli driver for the code is:

* CLI
* JPlag
* SubmissionSetBuilder
* other small classes like JPlagOptions, Verbosity, etc…

Nothing should depend on this, maybe except `Report`, as it is "legacier" than the rest, and probably need the options for reporting purpose.

For things like Language or the excluded file names, this is already hold inside `JPlag`, just passing them into the report should be fine (see #272  https://github.com/GabrielL/JPlag/blob/languages-as-spi/jplag/src/main/java/de/jplag/CLI.java#L59 for example with Language)

The real issue would be `minimumTokenMatch` and `fileSuffixes`, since it comes from the cli and the language default configuration. It is not used for the moment by `Report`, but I am guessing that the information will be relayed in the new json formatted output.

For reporting purpose, it could be possible to create an ad-hoc object that will contain every relevant parameters, already process and ready to be consumed.

GabrielL:
Since I am looking into the different options, 2 question arise:
* Is the `hasBaseCode` option is really necessary in `GreedyStringTiling` ? https://github.com/jplag/JPlag/blob/master/jplag/src/main/java/de/jplag/GreedyStringTiling.java#L224. From what I understand, `token.basecode` can only be true if `hasBaseCode()` is also true.
* In `Submission`/`SubmissionSet`, why do we care about `isdebugParser()` ? https://github.com/jplag/JPlag/blob/master/jplag/src/main/java/de/jplag/Submission.java#L176. It seems to me that this copies the non parsable files. Maybe better error reporting could simply describe what are those files, instead of trying to copy them elsewhere ?

tsaglam:
[...]

PR #276, opened 2022-01-19
Make JPlagOptions read only

GabrielL:
This patchset refactor JPlagOptions usage out of the main classes, in order to keep it outside of these classes.

* create a builder for JPlagOptions
* make JPlagOptions read-only
* move all rewritten fields inside JPlag class
* inject only the necessary options into the rest of the code
* add a ConfigurationReport class, accessible from JPlag directly, that will gives all the parameters used for the Report subsystem (unused for the moment)

## issues

* `GreedyStringTiling` still use `hasBaseCode`, I don't really know why. For me it should not be necessary, but the code it a bit hairy, and there is not enough test to have complete trust when changing the code.
* The JPlagOptions builder is horribly long, I really miss lombok here

Closes: #274

### DIFF
--- a/jplag/src/main/java/de/jplag/CLI.java
+++ b/jplag/src/main/java/de/jplag/CLI.java
@@ -57,7 +57,7 @@ public class CLI {
             System.out.println("JPlag initialized");
             JPlagResult result = program.run();
             File reportDir = new File(arguments.getString(RESULT_FOLDER.flagWithoutDash()));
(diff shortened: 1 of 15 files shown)


tsaglam:
The conflicts need to be resolved before we can merge. Thus I will also wait with the review.

tsaglam:
Closed due to lack of activity, happy to re-open anytime.

Issue #233, opened 2021-11-22
JPlagComparison overrides Object.equals but doesn't define Object.hashCode


tsaglam:
As pointed out by @Alberth289346 in #230:

> `JPlagComparison` overrides `Object.equals` but doesn't define `Object.hashCode`. This is dangerous since `hashCode` is expected to return equals values for equal instances. A proper hashing function should be added, `equals` should not be overridden, or (as a hack) `hashCode` should crash the program so it's known not to give invalid results.

Currently, comparisons are only used in hashmaps in the outdated report generation that will be replaced soon. However, this is still a scary anti-pattern that should be avoided to avoid future issues.
In detail, the main danger comes from _[bucket adjustment](https://www.andreinc.net/2021/11/08/a-tale-of-java-hash-tables#separate-chaining-or-how-hashmapkv-works-internally)_, where buckets of a certain size are restructured from chained data to red-black trees for performance reasons. Without a hashcode method consistent with the equals method the hashmap might behave differently once bucket adjustment starts.

PR #277, opened 2022-01-19
fix(Comparison): add hashCode()

GabrielL:
Just add `hashCode()` to JPlagComparison

 closes: #233

### DIFF
--- a/jplag/src/main/java/de/jplag/JPlagComparison.java
+++ b/jplag/src/main/java/de/jplag/JPlagComparison.java
@@ -1,12 +1,6 @@
 package de.jplag;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.IntStream;
 
 /**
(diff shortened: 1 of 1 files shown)


tsaglam (changes requested):
This `hashCode()` implementation breaks the [contract](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/lang/Object.html#hashCode()) defined by `Object.hashCode()`, as equal objects declard by `equals()` do not have the same hash code.

Either we remove the `equals()` method and adapt JPlag to work without it, or we adapt the `hashCode()` implementation.

GabrielL:
> This `hashCode()` implementation breaks the [contract](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/lang/Object.html#hashCode()) defined by `Object.hashCode()`, as equal objects declard by `equals()` do not have the same hash code.

Oh yes, my bad.

> Either we remove the `equals()` method and adapt JPlag to work without it, or we adapt the `hashCode()` implementation.

maybe something like this would be better

```java
    @Override
    public boolean equals(Object other) {
        if (!(other instanceof JPlagComparison)) {
            return false;
        }
        return (compare(this, (JPlagComparison) other) == 0)
                && firstSubmission.equals(((JPlagComparison) other).firstSubmission)
                && secondSubmission.equals(((JPlagComparison) other).secondSubmission);
    }
```

tsaglam:
> maybe something like this would be better

I tried to find out where the equals/compare methods were actually used in order to find out if we can implement your proposed equals method. However, it seems that the comparator-defined `compare()` method is never used and the `equals()` method is only used in the legacy report generation when comparisons are put into hash maps.
Thus we could remove the implements-relation, remove the `compare()` method and implement your `equals()` method.

PR #279, opened 2022-01-26
Attach parent exceptions to JPlag exceptions to preserve the root cause.

Alberth289346:
Please consider for merge


tsaglam (approved):
Looks good, this is a reasonable change!

Issue #116, opened 2021-05-31
Readd clustering and min/max/avg scores


sebinside:
Several functionalities were removed by #89 but should be restored. This includes:
- [ ] Clustering of multiple submissions based on similarity (groups of plagiarism)
- [ ] Max similarity to detect plagiarism even when one solution is obfuscated with additional code
- [ ] Min similarity (lesser priority, maybe not relevant at all)

tsaglam:
@simding

SimDing:
The "Max similarity" [is really just the maximum of both of the similarities](https://github.com/jplag/JPlag/blob/fbded8822d02c852fdc017e225a3d036fe417285/jplag/src/main/java/jplag/Program.java#L1069).

There are two other metrics calculated at that place, but omitted in the report:

- *Min similarity*: like max similarity but with minimum
- *Avg similarity*: intersection over union

When looking at it from a vulnerability perspective, both of them are worse than than max similarity. The only advantage over plain similarity I can see in them over plain similarity is that they are symmetric. It seems like IOU was used for clustering. I don't plan to add *Min similarity again*, I might use IOU for clustering again, but I guess I won't add it to the report.

However I am playing with the thought of adding additional metrics that disregard submission size.

tsaglam:
Good to have that confirmed, then I would say let us focus on clustering first and after that, you can experiment with additional metrics.
> I might use IOU for clustering again, but I guess I won't add it to the report.
Yes, you do not need to add it to the report generation, as this is done as part of #192 for the new report anyways.
Just make sure to provide the clustering as part of the result object in some meaningful way.

SimDing:
Should it be possible to use different clusterings in the same JPlag run? For example:

1. Do hierarchical clustering twice, but use a different threshold and inter-cluster-similarity each time
2. Cluster with *max-similarity*, *min-similarity* and *avg-similarity*

tsaglam:
> Should it be possible to use different clusterings in the same JPlag run? For example:
Here are some thoughts:

- At first glance, different types of clusterings seem useful
- However, we do not know yet, if it really provides more useful information for the users in the end, however, we can test that
- Looking at the clusters in the new report should not be complicated, it should not feel like a parameter-tuning-based task where you need to check different clustering types and see if they make a difference for your JPlag run.
- So if we allow multiple clusterings because it is actually useful, we should limit it to maybe 2-3 types max

[...]

PR #89, opened 2020-12-10
Java API for JPlag

krusche:
In this pull request, @philippbauch has implemented a Java API for JPlag that allow other tools (e.g. [Artemis](http://github.com/ls1intum/Artemis)) to invoke JPlag in a more object oriented way instead of using command line options.

We also added new objects (shown in the README file in a UML class diagram) to improve the result reporting so that other tools can interpret the results more easily.

#### Example 

```java
JPlagOptions options = new JPlagOptions("/path/to/rootDir", LanguageOption.JAVA_1_9);
options.setBaseCodeSubmissionName("template");

JPlag jplag = new JPlag(options);
JPlagResult result = jplag.run();

List<JPlagComparison> comparisons = result.getComparisons();

// Optional
File outputDir = new File("/path/to/output");
Report report = new Report(outputDir);

report.writeResult(result);
```

#### Class Diagram

This PR also updates the JPlag version to 3.0.0, but tries to stay compatible with 2.x versions and previous workflows.

We are happy to discuss any changes or improvements.

Please note: Some folders have been deleted to simplify the project structure, but we can also add them again in case they are still needed

PR #281, opened 2022-01-28
Readd clustering

SimDing:
Concerning issue #116 

This adds the clustering that was removed in #89 again, not including any user interfaces.

# Clustering

The clustering is implemented in the de.jplag.clustering package.
It includes two clustering algorithms (spectral and agglomerative), preprocessing, decoupling logic, clustering options and a factory class which can be used to run the clustering in just two statements.

### Agglomerative Clustering

Uses a bottom-up approach to successively merge similar clusters. It stops once there are no clusters left that are similar enough to merge. An implementation of this algorithm was originally included in the code base, but removed in #89. It is still included because it is a much simpler approach than the spectral clustering.

### Spectral Clustering

Spectral clustering is a clustering approach specifically for for graph data. This matches the problem since the similarities between all submissions can be thought of as a fully connected graph.
Spectral clustering works by computing the Laplace Matrices of the graphs and representing the nodes as k-dimensional vectors using it's Eigenvectors.
At that point the resulting vectors can be clustered using a space-partitioning algorithm, I used _k-Means++_.
Still for both k-Means as well as the reduction to k dimensions, the unknown final number of clusters k needs to be known.
In addition, k-Means++ yields probabilistic results.
To find a good choice for k and a "good" clustering I employ Baysian Optimization.
A metric I found in line with my notion of a "good" clustering is:

"The average of the clusters _modularity_ times their average inner cluster similarity over the number of the clusters connections."
With _modularity_ I mean the measure introduced by Newman, M.; Girvan, M. in "Finding and evaluating community structure in networks (2004)".

Spectral clustering is used by default.

## Preprocessing

As it can be advantageous to apply some preprocessing before clustering (in particular with Spectral clustering) I included three options for preprocessing.

### CDF Preprocessor

Estimates the cumulative distribution function of all similarities and multiplies each similarity with the CDF evaluated at that similarity. This has the effect of driving the lowest similarities close to zero while hardly changing the highest ones.
Since this preprocessor is non-parametric and worked well during my experiments I made it the default.

### Threshold Preprocessor

Suppresses all similarities below a given threshold. Good values for the threshold vary greatly with the set of input submissions.

### Percentile Preprocessor

Is the same as the threshold preprocessor, but the threshold is given as a percentile of the calculated similarities, making it more robust.

## Options and CLI

The (many) options for clustering are all defined in a new `ClusteringObjects` class. This class also contains sane defaults, that should allow users the run the clustering without specifying any additional CLI flags or defining them programmatically.

If refactored the `CommandLineArgument` enum a little because I did need to add addition optimal parameters and that many constructors became confusing. It now works in a _builder-pattern-ish_ fashion. I also added a small class for dealing with groups of arguments (`Clustering` and `Clustering - Preprocessing` in the help text).

This is how the help message is now displayed:

```
Benutzung: jplag [-h] [-l {java,python3,cpp,csharp,char,text,scheme}] [-bc BC] [-v {quiet,long}] [-d] [-S S] [-p P] [-x X] [-t T] [-m M] [-n N] [-r R] [-c {normal,parallel}]
                 [--cluster-skip] [--cluster-alg {AGGLOMERATIVE,SPECTRAL}] [--cluster-metric {AVG,MIN,MAX,INTERSECTION}] [--cluster-spectral-bandwidth bandwidth]
                 [--cluster-spectral-noise noise] [--cluster-spectral-min-runs min] [--cluster-spectral-max-runs max] [--cluster-spectral-kmeans-interations iterations]
                 [--cluster-agglomerative-threshold threshold] [--cluster-agglomerative-inter-cluster-similarity {MIN,MAX,AVERAGE}] [--cluster-pp-none | --cluster-pp-cdf |
                 --cluster-pp-percentile percentile | --cluster-pp-threshold threshold] rootDir [rootDir ...]

JPlag - Maintained by SDQ
Created by IPD Tichy, Guido Malpohl, and others. JPlag logo designed by Sandro Koch. Currently maintained by Sebastian Hahner and Timur Saglam.

Positions-Argumente:
  rootDir                Root-directory that contains submissions

Benannte Argumente:
  -h, --help             zeigt diese Hilfe und beendet sich.
  -l {java,python3,cpp,csharp,char,text,scheme}
                         Select the language to parse the submissions (Standard: java)
  -bc BC                 Path of the directory containing the base code (common framework used in all submissions)
  -v {quiet,long}        Verbosity of the logging (Standard: quiet)
  -d                     Debug parser. Non-parsable files will be stored (Standard: false)
  -S S                   Look in directories <root-dir>/*/<dir> for programs
  -p P                   comma-separated list of all filename suffixes that are included
  -x X                   All files named in this file will be ignored in the comparison (line-separated list)
  -t T                   Tunes the comparison sensitivity by adjusting  the  minimum  token  required  to  be  counted  as  a  matching  section. A smaller <n> increases the
                         sensitivity but might lead to more false-positives
  -m M                   Comparison similarity threshold [0-100]: All comparisons above this threshold will be saved (Standard: 0.0)
  -n N                   The maximum number of comparisons that will be shown in the generated report, if set to -1 all comparisons will be shown (Standard: 30)
  -r R                   Name of the directory in which the comparison results will be stored (Standard: result)
  -c {normal,parallel}   Comparison mode used to compare the programs (Standard: normal)

Clustering:
  --cluster-skip         Skips the clustering (Standard: false)
  --cluster-alg {AGGLOMERATIVE,SPECTRAL}
                         Which clustering algorithm to use. Agglomerative merges similar  submissions  bottom  up. Spectral clustering is combined with Bayesian Optimization
                         to execute the k-Means clustering algorithm multiple times, hopefully finding a "good" clustering automatically. (Standard: SPECTRAL)
  --cluster-metric {AVG,MIN,MAX,INTERSECTION}
                         The metric used for clustering. AVG is intersection over union, MAX can expose some attempts of obfuscation. (Standard: MAX)
  --cluster-spectral-bandwidth bandwidth
                         Bandwidth of the matern kernel in the Gaussian Process used  during  the  search  for  a  good number of clusters for spectral clustering. If a good
                         clustering result is found during the search, numbers of clusters  that  differ  by  something  in range of the bandwidth are also expected to good.
                         (Standard: 20.0)
  --cluster-spectral-noise noise
                         The result of each run in the search for good clusterings are random.  The  noise level models the variance in the "worth" of these results. It also
                         acts as a regularization constant. (Standard: 0.0025000002)
  --cluster-spectral-min-runs min
                         Minimum number of k-Means executions during spectral clustering. With these initial clustering sizes are explored. (Standard: 5)
  --cluster-spectral-max-runs max
                         Maximum number of k-Means executions during spectral  clustering.  Any  execution  after  the  initial  runs tries to balance between exploration of
                         unknown clustering sizes and exploitation of clustering sizes known as good. (Standard: 50)
  --cluster-spectral-kmeans-interations iterations
                         Maximum number of iterations during each execution of the k-Means algorithm. (Standard: 200)
  --cluster-agglomerative-threshold threshold
                         Only clusters with an inter-cluster-similarity greater than this threshold are merged during agglomerative clustering. (Standard: 0.2)
  --cluster-agglomerative-inter-cluster-similarity {MIN,MAX,AVERAGE}
                         How to measure the similarity of two clusters during  agglomerative  clustering.  Minimum,  maximum or average similarity between the submissions in
                         each cluster. (Standard: AVERAGE)

Clustering - Preprocessing:
  --cluster-pp-none      Do not use any preprocessing before clustering. Not recommended for spectral clustering. (Standard: false)
  --cluster-pp-cdf       Before clustering, the value of the cumulative distribution function  of  all  similarities is estimated. The similarities are multiplied with these
                         estimates. This has the effect of supressing similarities that are low compared to other similarities. (Standard: false)
  --cluster-pp-percentile percentile
                         Any similarity smaller than the given percentile will be suppressed during clustering.
  --cluster-pp-threshold threshold
                         Any similarity smaller than the given threshold value will be suppressed during clustering.

```

## Technical Stuff

- This adds two new maven dependecies: commons-math (k-Means++, vectors, matrices, and many small algorithms used here and there) and mockito (testing)
- I tried to minimize coupling between the clustering code and other code
  - The main package `de.jplag` is only coupled to `de.jplag.clustering` through the `ClusteringOptions` and `ClusteringFactory` classes
  - The clustering package `de.jplag.clustering` is only coupled to `de.jplag` through the `JPlagComparison` and `Submission` classes
  - The clustering code only uses `JPlagComparison` and `Submission` in the `ClusteringFactory` and `ClusteringAdapter` classes, the latter replacing submissions with integer indices and comparisons with matrices.


SimDing:
I'm not quite finished yet, but I have a question and would like you to see my current state:

Currently I've put every setting about the clustering plainly inside the JPlagOptions class. This feels pretty messy. Do you have a suggestion?

tsaglam:
> Currently I've put every setting about the clustering plainly inside the JPlagOptions class. This feels pretty messy. Do you have a suggestion?

That is a good question. Currently, I count 14 additional options you added in this PR. The question here would be how many of those the user really modifies. If some are parameters that might be tweaked in the future, but most users will not change the default values, then we should not expose these parameters as options. Even from a usability context, modifying 14 options for clustering alone via the CLI seems excessive and very unlikely (think about the flag you need to define). 
Thus we can maybe reduce the number of options to the user by keeping some clustering parameters as internal parameters that may be tweaked by devs but not the users.

SimDing:
> The question here would be how many of those the user really modifies

I think the defaults are set kind of sane, so I hope most users would not have to change much.

> Even from a usability context, modifying 14 options for clustering alone via the CLI seems excessive

Users would not use all options at the same time.

I see two cases in which a user would want to change the options:

- Clustering takes too long?
  - Disable clustering
  - Use a threshold preprocessor
  - (spectral) use less kMeans iterations
  - (spectral) use less runs
- The result is not good? The problem is that then there is not really much users can do but fiddle with the parameters that change the result of the clustering. I don't know any one of those parameters that would be best kept from users. In that case at most five parameters would be set (preprocessor, preprocessor option, noise, kernel bandwidth, and similarity metric).

A user who had both problems would use 7 options at most.

The only thing I can really remove without bad aftertaste is the option about pruning bad clusters. There does not seem to be a practical reason to look at those.

SimDing:
@tsaglam I think this can be reviewed now.

jplag/src/main/java/de/jplag/clustering/algorithm/SpectralClustering.java
```diff
@@ -0,0 +1,172 @@
...
+            this.eigenVector = eigenVector.toArray();
+        }
+
+        @Override
+        public double[] getPoint() {
+            return eigenVector;
+        }
+    }
+
+    public static class ClusteringOptions {
```
tsaglam, line 139: It seems that different classes in your PR implement their own clustering options class? I am not sure If I understand what the design rationale behind it is. Can you elaborate?

tsaglam (changes requested):
@SimDing this is only an initial and shallow review to allow you to start integrating the feedback asap. A second review will follow.
My main concerns are:

- Lack of JDoc comments, especially for interfaces and facade classes, but more generally for all public types and members
- Heavy use of abbreviations, which should be avoided to guarantee long term maintainability and code quality in general
- Heavy use of nested types, which should only be used when its lifecycle is completely connected to the outer type or when the nested type must not be seen or when they significantly improve the code (note that nested classes can also lead to unreadable code so try to avoid them if there is no explicit reason to use them)

jplag/src/main/java/de/jplag/clustering/algorithm/BFGS.java
```diff
@@ -0,0 +1,835 @@
...
+ * <ol>
+ * <li>Roger Fletcher. Practical methods of optimization.</li>
+ * <li>D. Liu and J. Nocedal. On the limited memory BFGS method for large scale optimization. Mathematical Programming B
+ * 45 (1989) 503-528.</li>
+ * <li>Richard H. Byrd, Peihuang Lu, Jorge Nocedal and Ciyou Zhu. A limited memory algorithm for bound constrained
+ * optimization.</li>
+ * </ol>
+ * @author Haifeng Li
+ */
+public class BFGS {
```
dfuchss, line 70: Not sure, but if this algorithm is simply copied, you could simply add the dependency

```
    <dependency>
      <groupId>com.github.haifengl</groupId>
      <artifactId>smile-core</artifactId>
      <version>2.6.0</version>
    </dependency>
```

jplag/src/main/java/de/jplag/clustering/algorithm/BayesianOptimization.java
```diff
@@ -0,0 +1,189 @@
...
+    private static final int MAX_NON_ZERO_ACQ_FN_EVALS_PER_ITERATION = 50;
+    private static final int MAXIMUM_ACQ_FN_EVALS_PER_ITERATION = 1000;
+
+    private RealVector minima;
+    private RealVector maxima;
+    private int maxEvaluations;
+    private int initialPoints;
+    private double noise;
+    private RealVector lengthScale;
+    public boolean debug = false;
```
dfuchss, line 30: avoid public attributes

jplag/src/main/java/de/jplag/clustering/algorithm/GaussianProcess.java
```diff
@@ -0,0 +1,210 @@
...
+        brokenXIndex.ifPresent(brokenIndex -> {
+            throw new IllegalArgumentException(MessageFormat.format("X has different dimensions at index {0} ({2}) and index {1} ({3})",
+                    brokenIndex - 1, brokenXIndex, X.get(brokenIndex - 1).getDimension(), X.get(brokenIndex).getDimension()));
+        });
+        if (noise <= 0) {
+            throw new IllegalArgumentException(MessageFormat.format("noise must be strictly positive, got {0}", noise));
+        }
+        if (lengthScale.length != X.get(0).getDimension()) {
+            throw new IllegalArgumentException(
+                    MessageFormat.format("lengthScale is of different dimension {0} than X values {1}", lengthScale.length, X.get(0).getDimension()));
```
dfuchss, line 108: Multiple Index Out of bounds exceptions and other exceptions might be possible if X or Y ist empty or null

jplag/src/main/java/de/jplag/clustering/algorithm/GaussianProcess.java
```diff
@@ -0,0 +1,210 @@
...
+        double dist = left.ebeDivide(lengthScale).getDistance(right.ebeDivide(lengthScale));
+        dist *= SQRT_5;
+        return (1 + dist) * Math.exp(-dist);
+    }
+
+    /**
+     * ascii graph for debugging
+     */
+    public String toString(RealVector min, RealVector max, int width, int height, double minY) {
+        char[][] out = new char[height][width];
```
dfuchss, line 166: Maybe cache the string output since this method seems to be complex.

jplag/src/main/java/de/jplag/clustering/algorithm/SpectralClustering.java
```diff
@@ -0,0 +1,172 @@
...
+        EigenDecomposition ed = new EigenDecomposition(L);
+
+        List<Integer> eigenValueIds = new ArrayList<>(N);
+        for (int i = 0; i < N; i++) {
+            eigenValueIds.add(i);
+        }
+        eigenValueIds.sort(Comparator.comparingDouble(ed::getRealEigenvalue));
+
+        // find number of clusters as the multiplicity of eigenvalue 0
+        double eps = 0.05;
```
dfuchss, line 82: extract to constant

jplag/src/main/java/de/jplag/clustering/algorithm/SpectralClustering.java
```diff
@@ -0,0 +1,172 @@
...
+                options.maxRuns, options.GPVariance, lengthScale);
+        // bo.debug = true;
+        BayesianOptimization.OptimizationResult<Collection<Collection<Integer>>> boResult = bo.maximize(r -> {
+            int k = (int) Math.round(r.getEntry(0));
+            k = Math.max(minK, k);
+            k = Math.min(maxK, k);
+            Collection<Collection<Integer>> clustering = cluster(k, N, eigenValueIds, ed);
+            IntegerClusteringResult modularityRes = new IntegerClusteringResult(clustering, similarityMatrix);
+            return new BayesianOptimization.OptimizationResult<>(modularityRes.getWorth((a, b) -> (float) similarityMatrix.getEntry(a, b)),
+                    clustering);
```
dfuchss, line 98: Maybe extract to method.

jplag/src/test/java/de/jplag/cli/ClusteringTest.java
```diff
@@ -0,0 +1,34 @@
...
+import de.jplag.clustering.Preprocessors;
+
+public class ClusteringTest extends CommandLineInterfaceTest {
+
+    @Test
+    public void parsePercentilePreProcessor() {
+        String argument = CommandLineArgument.CLUSTER_PREPROCESSING_PERCENTILE.flag() + "=" + Float.toString(0.5f);
+        buildOptionsFromCLI(argument, CURRENT_DIRECTORY);
+        assertEquals(Preprocessors.PERCENTILE, options.getClusteringOptions().getPreprocessor());
+        assertEquals(0.5, options.getClusteringOptions().getPreprocessorPercentile(), 0.000001);
```
dfuchss, line 17: Move eps to constant

dfuchss (changes requested):
I've added some comments :)

jplag/src/main/java/de/jplag/CommandLineArgument.java
```diff
@@ -9,62 +11,103 @@
...
+                    .argumentGroup(CLUSTERING_GROUP_NAME).defaultsTo(ClusteringOptions.DEFAULTS.getAgglomerativeInterClusterSimilarity())),
+    CLUSTER_PREPROCESSING_NONE(
+            new Builder("--cluster-pp-none", Boolean.class).mutuallyExclusiveGroup(CLUSTERING_PREPROCESSING_GROUP_NAME)
+                    .action(Arguments.storeTrue())),
+    CLUSTER_PREPROCESSING_CDF(
+            new Builder("--cluster-pp-cdf", Boolean.class).mutuallyExclusiveGroup(CLUSTERING_PREPROCESSING_GROUP_NAME).action(Arguments.storeTrue())),
+    CLUSTER_PREPROCESSING_PERCENTILE(
+            new Builder("--cluster-pp-percentile", Float.class).mutuallyExclusiveGroup(CLUSTERING_PREPROCESSING_GROUP_NAME).metaVar("percentile")),
+    CLUSTER_PREPROCESSING_THRESHOLD(
+            new Builder("--cluster-pp-threshold", Float.class).mutuallyExclusiveGroup(CLUSTERING_PREPROCESSING_GROUP_NAME).metaVar("threshold"));
```
watob, line 85: Is the preprocessing the same for Agglomerative and Spectral Clustering? In case it is not, it would be nice to also 
group the preprocessing in the CLI options or to mention the related clustering algorithm.

jplag/src/main/java/de/jplag/CommandLineArgument.java
```diff
@@ -136,4 +184,56 @@ private String retrieveDescriptionFromMessages() {
         }
         return Messages.getString(getClass().getSimpleName() + "." + builder.toString());
     }
+
+    private static class Builder {
```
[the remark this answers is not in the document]
SimDing: I've made this Builder Nested because it makes no sense to use it outside of the scope of this class.
And since it is private I felt that a generic name suffices.

Should I still move it outside and rename it?

jplag/src/main/java/de/jplag/clustering/Preprocessor.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Interface for classes that process similarity matrices before any clustering. Classes implementing this interface
+ * must ensure that they do not produce zero rows/columns inside the similarity matrix. They must also be able to
+ * calculate the original indices of rows/columns after use through their {@link Preprocessor#postProcessResult} method.
+ */
+public interface Preprocessor {
+    double[][] preprocessSimilarities(double[][] similarityMatrix);
+
+    Collection<Collection<Integer>> postProcessResult(Collection<Collection<Integer>> result);
+
```
watob, line 24: Ist the naming postProcessResult intentional or would postPreprocessResult be a better fitting name?

dfuchss: Even more: Is a collection<collection<Integer>> a good type here?

watob (changes requested):
I agree with the requests of @tsaglam. In his feedback he mentioned the main points to work on that came up during a first shallow review.

jplag/src/main/java/de/jplag/clustering/Preprocessors.java
```diff
@@ -0,0 +1,20 @@
...
+import de.jplag.clustering.preprocessors.PercentileThresholdProcessor;
+import de.jplag.clustering.preprocessors.ThresholdPreprocessor;
+
+/**
+ * List of all usable {@link Preprocessor}s.
+ */
+public enum Preprocessors {
+    NONE,
+    /** {@link CdfPreprocessor} */
+    CDF,
```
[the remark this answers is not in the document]
SimDing: These are also the names used in the CLI parameters. I would rather not have have users type `--clustering-pp CUMULATIVE_DISTRIBUTION_FUNCTION`.

jplag/src/main/java/de/jplag/clustering/algorithm/AgglomerativeClustering.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        @Override
+        public int compareTo(ClusterConnection other) {
+            return (int) Math.signum(other.similarity - similarity);
+        }
+
+    }
+
+    private class Cluster {
```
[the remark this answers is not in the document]
SimDing: Because this is juggled around in a hash maps and -sets.
Lists usually implement hashing based on their members, which would impact performance.
Adding a comment.

jplag/src/main/java/de/jplag/clustering/algorithm/AgglomerativeClustering.java
```diff
@@ -0,0 +1,189 @@
...
+            this.submissions = submissions;
+        }
+
+        public List<Integer> getSubmissions() {
+            return submissions;
+        }
+
+    }
+
+    private class MergedList extends AbstractList<Integer> {
```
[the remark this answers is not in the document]
SimDing: I'll just remove this and use ArrayList

jplag/src/main/java/de/jplag/clustering/algorithm/AgglomerativeClustering.java
```diff
@@ -0,0 +1,189 @@
...
+            }
+        }
+
+        if (options.similarity == InterClusterSimilarity.AVERAGE) {
+            similarity /= leftCluster.size() * rightCluster.size();
+        }
+        return similarity;
+    }
+
+    public static class ClusteringOptions {
```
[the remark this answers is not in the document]
SimDing: The rational with the private classes is that that they should be hidden.
I'll work on the others.

jplag/src/main/java/de/jplag/clustering/algorithm/SpectralClustering.java
SimDing: Just before reading this comment I deleted the nested ClusteringOptions classes

jplag/src/main/java/de/jplag/clustering/algorithm/GaussianProcess.java
```diff
@@ -0,0 +1,210 @@
...
+     * @param x position to predict at
+     * @return mean prediction at x
+     */
+    public double predict(RealVector x) {
+        RealVector k = maternKernel(X, x, lengthScale);
+
+        double mu = weight.dotProduct(k);
+
+        return mu * this.standardDeviation + this.mean;
+    }
```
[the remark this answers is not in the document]
SimDing: Predict whatever the gp is trained to predict :)

jplag/src/main/java/de/jplag/clustering/IntegerMapping.java
```diff
@@ -0,0 +1,51 @@
+package de.jplag.clustering;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * Creates a mapping of any set to integers
+ */
+public class IntegerMapping<T> {
```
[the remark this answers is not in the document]
SimDing: This class also includes a counter.
```
IntegerMapping mapping = new IntegerMapping(2);
mapping.map("foo");
mapping.map("bar");
System.out.println(mapping.unmap(0)); // foo
System.out.println(mapping.unmap(1)); // bar
```

I could swap the bi-mapping functionality with the one from commons, but don't think this trivial usage of bi mapping justifies adding a new dependency.

jplag/src/main/java/de/jplag/clustering/algorithm/AgglomerativeClustering.java
```diff
@@ -0,0 +1,189 @@
...
+            clusters.add(combined);
+        }
+
+        return clusters.stream().map(Cluster::getSubmissions).collect(Collectors.toList());
+    }
+
+    private float clusterSimilarity(List<Integer> leftCluster, List<Integer> rightCluster, RealMatrix similarityMatrix) {
+        float similarity = 0;
+        switch (options.similarity) {
+            case MIN:
```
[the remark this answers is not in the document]
SimDing: I don't think so, because there might be inter-cluster-similarities that can not be calculated in this fashion.

jplag/src/main/java/de/jplag/clustering/algorithm/AgglomerativeClustering.java
```diff
@@ -0,0 +1,189 @@
...
+
+    }
+
+    private class MergedList extends AbstractList<Integer> {
+
+        private List<Integer> left, right;
+        private int size;
+
+        public MergedList(List<Integer> left, List<Integer> right) {
+            this.left = left;
```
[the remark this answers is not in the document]
SimDing: I've deleted this class.

jplag/src/main/java/de/jplag/clustering/algorithm/AgglomerativeClustering.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+    }
+
+    private class MergedList extends AbstractList<Integer> {
+
+        private List<Integer> left, right;
+        private int size;
+
+        public MergedList(List<Integer> left, List<Integer> right) {
```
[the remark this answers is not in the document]
SimDing: Yeah, I know I deleted it. I had a misconception while writing this and thought it would lead to better performance. It's all ArrayLists now.

jplag/src/main/java/de/jplag/clustering/algorithm/GaussianProcess.java
SimDing: I've added a check for empty X at the beginning. And I mean in calling this with empty lists / null can not work anyway.

jplag/src/main/java/de/jplag/clustering/algorithm/GaussianProcess.java
SimDing: The only use of this method is for creating debug output and then each output would be different.

jplag/src/main/java/de/jplag/clustering/Preprocessor.java
SimDing: It's like this: Some preprocessing might introduce rows or columns that are zero and a problem for spectral clustering. So pre-processors must ensure that there are no zero rows/cols (remove them).
But this is changes the indices in the results from the clustering algorithm.
So the pre-processor is also expected to do a post-processing on the results so that the indices match with the indices of the original similarity matrix.

I've changed the interface to only map back single indices not collections of collections but that only moved stuff around.

tsaglam:
@SimDing I see, you already fixed quite a few issues. Can you please resolve every conversation above that you addressed by your commits? Those where you added an explanation or that are not addressed yet you can leave open. This allows me to see what the current state is and when to start the second review.

jplag/src/main/java/de/jplag/clustering/algorithm/SpectralClustering.java
```diff
@@ -0,0 +1,172 @@
...
+        });
+
+        return boResult.getValue();
+    }
+
+    private Collection<Collection<Integer>> cluster(int k, int n, List<Integer> eigenValueIds, EigenDecomposition ed) {
+        RealMatrix Q = new Array2DRowRealMatrix(n, k);
```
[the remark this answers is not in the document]
SimDing: I've changed to a little more descriptive names.
Do you expect a textual description of the algorithm?
To really understand what's going on there one needs to read related literature such as the paper in the comment.

jplag/src/main/java/de/jplag/CommandLineArgument.java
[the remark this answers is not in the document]
tsaglam: I would say leave it in for now.

jplag/src/main/java/de/jplag/clustering/IntegerMapping.java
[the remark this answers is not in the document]
tsaglam: Makes sense!

jplag/src/main/java/de/jplag/clustering/IntegerClusteringResult.java
```diff
@@ -0,0 +1,76 @@
...
+        int clusterIdx = 0;
+        List<Collection<Integer>> clusters = new ArrayList<>(clustering);
+        for (Collection<Integer> cluster : clusters) {
+            for (Integer submissionIdx : cluster) {
+                submissionIdx2ClusterIdx.put(submissionIdx, clusterIdx);
+            }
+            clusterIdx++;
+        }
+        if (clusters.size() > 0) {
+            RealMatrix E = new Array2DRowRealMatrix(clusters.size(), clusters.size());
```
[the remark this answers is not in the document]
tsaglam: Yes, I think `percentagesOfSimilaritySums` is way better than `E`.

jplag/src/main/java/de/jplag/clustering/Preprocessors.java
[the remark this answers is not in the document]
tsaglam: For CLI parameters, this is a different story. Abbreviations are commonly used for command-line interfaces, and we do that as well (e.g. `-bc`). However, the enumeral should be renamed nonetheless.

jplag/src/main/java/de/jplag/clustering/algorithm/GaussianProcess.java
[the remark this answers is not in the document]
tsaglam: If there is no better name than predict, that's ok, but mu/k should be changed.

jplag/src/test/java/de/jplag/clustering/ClusteringTest.java
```diff
@@ -0,0 +1,267 @@
...
+        URL url = getClass().getClassLoader().getResource(file);
+        if (url == null) {
+            throw new FileNotFoundException(file + " not found. Was the PseudonymizedReports submodule initialized?");
+        }
+        return url;
+    }
+
+    @Test
+    @Ignore
+    public void aClusteringOld() throws FileNotFoundException, URISyntaxException {
```
tsaglam, line 84: I am a bit confused by these ignored tests. Obviously, those were designed for the pseudonymized reports. Leaving them in for us to use is fine, but they should be well documented. Also, they should be separate from normal unit tests like the non-descriptive `testClustering` test a few lines above. Please refer to the [VolumeTest](https://github.com/jplag/JPlag/blob/aab5927fa8e44c1eee2bb575790096c64316493c/jplag/src/test/java/de/jplag/VolumeTest.java) class to see what I mean by documented and separate.

Moreover, I fail to see any assertions. A test case without assertions is not a test case. Also, without automated checking, this test case has little to no value for future maintenance. I would be important that at least some result checking is done. Else, we can remove these tests.

jplag/src/test/java/de/jplag/clustering/ClusteringTest.java
```diff
@@ -0,0 +1,267 @@
...
+import org.junit.Test;
+import org.mockito.invocation.InvocationOnMock;
+
+import de.jplag.JPlagComparison;
+import de.jplag.Submission;
+import de.jplag.clustering.algorithm.GenericClusteringAlgorithm;
+import de.jplag.clustering.algorithm.SpectralClustering;
+import de.jplag.clustering.preprocessors.CumulativeDistributionFunctionPreprocessor;
+
+public class ClusteringTest {
```
tsaglam, line 35: This is a comment regarding all tests: When looking only at the coverage of your test cases, it seems that there are a few untested classes: The preprocessor classes, the spectral clustering, the clustering factory, and the metric in the cluster class. Could you please add a few test cases for that?

jplag/src/main/java/de/jplag/clustering/algorithm/GaussianProcess.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * @param x position to predict at
+     * @return array containing the predicted [mean, std] at x.
+     */
+    public double[] predictWidthStd(RealVector x) {
```
tsaglam, line 53: If std means standard, please name the method accordingly.

jplag/src/main/java/de/jplag/clustering/algorithm/GaussianProcess.java
```diff
@@ -0,0 +1,212 @@
...
+        this.standardDeviation = standardDeviation;
+        this.cholesky = cholesky;
+        this.lengthScale = lengthScale;
+    }
+
+    /**
+     * @param x position to predict at
+     * @return mean prediction at x
+     */
+    public double predict(RealVector x) {
```
tsaglam, line 41: This method has 0 references, meaning it is never called.

jplag/src/main/java/de/jplag/clustering/algorithm/GaussianProcess.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Identical to {@link GaussianProcess#fit(List, double[], double, boolean, double[])} but using a one vector as length
+     * scale.
+     * @param X
+     * @param Y expected to have zero mean, unit variance if normalize is false
+     * @param noise variance of noise in Y
+     * @param normalize for Y
+     */
+    public static GaussianProcess fit(List<RealVector> X, double[] Y, double noise, boolean normalize) {
```
tsaglam, line 81: This method is only called from the test cases.

jplag/src/main/java/de/jplag/clustering/algorithm/BFGS.java
tsaglam: To be honest, the copied class is a bit problematic maintenance-wise. It would improve the quality of this PR a lot if we can replace it with a dependency. Could you please take a look at [this commit](https://github.com/dfuchss/jplag/commit/72d5b5357340f8be5bed6f9ee811020e32fe2609). I do not know what you changed in the copied code, so maybe you can tell us why you think it is not feasible. If we can replace the code with a dependency, we should!

jplag/src/main/java/de/jplag/clustering/ClusteringResult.java
```diff
@@ -0,0 +1,101 @@
...
+        for (Collection<Integer> cluster : clustering) {
+            for (Integer submissionIdx : cluster) {
+                clusterIndicesOfSubmissionIndices.put(submissionIdx, clusterIdx);
+            }
+            clusterIdx++;
+        }
+        List<Cluster<Integer>> clusters = new ArrayList<>(clustering.size());
+        float communityStrength = 0;
+        if (clustering.size() > 0) {
+            RealMatrix E = new Array2DRowRealMatrix(clustering.size(), clustering.size());
```
tsaglam, line 76: Naming! We will probably not mark all naming issues. But you should fix them in your whole code.
So no variable/parameter/field names starting in upper case. No abbreviations in names, no names that are only a letter or a greek letter.

jplag/src/main/java/de/jplag/clustering/algorithm/BayesianOptimization.java
```diff
@@ -0,0 +1,196 @@
...
+        return GaussianProcess.fit(X, Y.stream().mapToDouble(Double::doubleValue).toArray(), noise, true, lengthScale.toArray());
+    }
+
+    private double acquisitionFunction(GaussianProcess gpr, double[] r, double yMax) {
+        // expected improvement
+        double[] meanAndStandardDeviation = gpr.predictWidthStd(new ArrayRealVector(r));
+        double mean = meanAndStandardDeviation[0];
+        double standardDeviation = meanAndStandardDeviation[1];
+        double a = mean - yMax;
+        double z = a / standardDeviation;
```
tsaglam, line 79: You know the drill! ;)

jplag/src/main/java/de/jplag/clustering/algorithm/BayesianOptimization.java
```diff
@@ -0,0 +1,196 @@
...
+        RealVector size = maxima.subtract(minima);
+        return Stream.generate(generator::nextVector).map(x -> new ArrayRealVector(x)).map(x -> x.ebeMultiply(size)).map(x -> x.add(minima));
+
+    }
+
+    private GaussianProcess fit(List<RealVector> X, List<Double> Y) {
+        return GaussianProcess.fit(X, Y.stream().mapToDouble(Double::doubleValue).toArray(), noise, true, lengthScale.toArray());
+    }
+
+    private double acquisitionFunction(GaussianProcess gpr, double[] r, double yMax) {
```
tsaglam, line 73: What goes for fields and variables also goes for parameters.

tsaglam (changes requested):
A further review, addressing some more issues.
Good work so far including our review feedback, keep it going!

jplag/src/test/java/de/jplag/clustering/ClusteringTest.java
dfuchss: You might not want to use Ignore. 

Simply add an assume statement (JUnit Assume) that checks whether the neccessary files are present.

jplag/src/main/java/de/jplag/clustering/algorithm/BFGS.java
SimDing: smile relies on BLAS and LAPACK for this, which is a pain to set up (at least I did not manage to). Apart from using different implementations of many functions I did the following:

```
505a168
>         int runs = 50;
521a185,186
>             } else if (runs-- < 0) {
>                 return f;
783c453
<         double dt = t[i];
---
>         double dt = i < n ? t[i] : 0;
```

First the fallback return, which is a dirty fix for an infinite loop which occasionally arises and I think the second fixes a bug in the original implementation.
At least without  these changes the spectral clustering fails sometimes.

jplag/src/main/java/de/jplag/clustering/algorithm/SpectralClustering.java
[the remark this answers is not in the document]
SimDing: The paper is mentioned at the beginning of the public cluster function.
```
    @Override
    public Collection<Collection<Integer>> cluster(RealMatrix similarityMatrix) {
        // Calculate points to cluster according to "On spectral clustering: analysis and an algorithm" by Ng, Jordan & Weiss
        // 2001

```

jplag/src/main/java/de/jplag/clustering/algorithm/BayesianOptimization.java
SimDing: Tbh I don't even know how to call half of the intermediate values of the algorithms I implemented.

jplag/src/main/java/de/jplag/clustering/ClusteringFactory.java
```diff
@@ -0,0 +1,72 @@
...
+            case THRESHOLD:
+                preprocessor = new ThresholdPreprocessor(options.getPreprocessorThreshold());
+                break;
+            case PERCENTILE:
+                preprocessor = new PercentileThresholdProcessor(options.getPreprocessorPercentile());
+                break;
+            case NONE:
+            default:
+                preprocessor = null;
+                break;
```
dfuchss, line 48: My honest opinion:

Such switch case statements should be avoided. 

Simply add a attribute of Type `Function<ClusteringOptions, ClusteringPreprocessor>` to the enum values.
This would make it more easy to create new preprocessors later on & you can simply delete the switch-case

jplag/src/main/java/de/jplag/clustering/algorithm/AgglomerativeClustering.java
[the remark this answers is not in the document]
dfuchss: Why? You can simply add a set an attribute similarity in the Enum for each value: Float.MAX_VALUE,  and so on .. then you yo not need the switch-case

jplag/src/main/java/de/jplag/clustering/Preprocessing.java
```diff
@@ -0,0 +1,20 @@
...
+
+import de.jplag.clustering.preprocessors.CumulativeDistributionFunctionPreprocessor;
+import de.jplag.clustering.preprocessors.PercentileThresholdProcessor;
+import de.jplag.clustering.preprocessors.ThresholdPreprocessor;
+
+/**
+ * List of all usable {@link Preprocessor}s.
+ */
+public enum Preprocessing {
+    NONE,
```
dfuchss, line 13: Here you could simply add some attributes to the values

dfuchss (changes requested):

jplag/src/main/java/de/jplag/clustering/algorithm/AgglomerativeClustering.java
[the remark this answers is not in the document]
SimDing: Well I've done it now. I was just concerned that in future one might want to add an inter-cluster-similarity that can not be calculated in a loop like this.
Meaning there might be inter-cluster-similarities that can not meaningfully implement these functions.

jplag/src/main/java/de/jplag/clustering/ClusteringFactory.java
```diff
@@ -31,25 +29,11 @@
...
-                preprocessor = null;
-                break;
-        }
-        if (preprocessor != null) {
+        Optional<ClusteringPreprocessor> preprocessor = options.getPreprocessor().constructPreprocessor(options);
+
+        if (preprocessor.isPresent()) {
             // Package preprocessor into a clustering algorithm
-            clusteringAlgorithm = new PreprocessedClusteringAlgorithm(clusteringAlgorithm, preprocessor);
+            clusteringAlgorithm = new PreprocessedClusteringAlgorithm(clusteringAlgorithm, preprocessor.orElseThrow());
```
dfuchss, line 36: why orElseThrow? Simply use get()

jplag/src/main/java/de/jplag/clustering/algorithm/InterClusterSimilarity.java
```diff
@@ -1,7 +1,44 @@
...
+     * @return similarity between the two clusters
+     */
+    public float clusterSimilarity(List<Integer> leftCluster, List<Integer> rightCluster, RealMatrix similarityMatrix) {
+        float similarity = this.neutralElement;
+
+        for (int leftIndex = 0; leftIndex < leftCluster.size(); leftIndex++) {
+            int leftSubmission = leftCluster.get(leftIndex);
+            for (int rightIndex = 0; rightIndex < rightCluster.size(); rightIndex++) {
+                float submissionSimilarity = (float) similarityMatrix.getEntry(leftSubmission, rightCluster.get(rightIndex));
+                similarity = (float) this.accumulator.applyAsDouble(similarity, submissionSimilarity);
```
dfuchss, line 35: Instead of casting simply use BiFunction<Double, Double, Float> or replace float by double

dfuchss (changes requested):
Minor comments :)

jplag/src/main/java/de/jplag/clustering/ClusteringFactory.java
SimDing: From the Optional#get() API doc:
> The preferred alternative to this method is [orElseThrow()](https://docs.oracle.com/javase/10/docs/api/java/util/Optional.html#orElseThrow()).

SimDing: See also https://bugs.openjdk.java.net/browse/JDK-8140281 on why this is the case.

jplag/src/test/java/de/jplag/clustering/ClusteringTest.java
SimDing: I've added test cases for preprocessor classes, the spectral clustering, the metrics and the calculation of the community strength.
I can't think off a meaningful test for the clustering factory, that is not just entirely replicating what's in there.

tsaglam (approved):

tsaglam:
@SimDing https://github.com/SimDing/JPlag/pull/5

PR #283, opened 2022-02-01
Moved CLI from JPlag main project to separate project 

dfuchss:
The CLI module is configured to be not deployed to maven central as it's no library in the narrow sense.
Feel free to adopt, comment, etc. :)


dfuchss:
Please do not merge this before #250 
I want to move the logger implementation (dependency) to the CLI. 
After that, JPlag (-> Library) would only depend on the Logger API.

dfuchss:
Closed because of #409

Issue #192, opened 2021-10-12
Replace the outdated HTML report generation


tsaglam:
- Replace legacy report generation with a modern system
- Generate reports that look cleaner and more modern
- Should roughly cover the functionality of the  [legacy web report](https://github.com/jplag/ExampleReport) [internal link] (including clustering and max similarity)

sebinside:
The current report generation generates HTML files by combining strings. Java 1.0 style, but considered bad practise for many years . Thus, I would propose the strict separation of machine-readable report data generation and GUI code to view those. 21b9bd3 prepares this by adding a [Vue 3](https://v3.vuejs.org/) boilerplate for a web-based report viewer. By updating the report generation, e.g., to a JSON-based format, we could separate generation and viewing and minimize dependencies.

To do so, I would propose the following steps:
- [ ] Getting started with the report generation of JPlag
- [ ] Identify all information from the plagiarism detection that is currently used in the HTML report generation
- [ ] Identify which of this information is already accessible using the new Java API
- [ ] Define new functionality to create JSON files with all required information on plagiarism (e.g., following the current format of one overview file and one file for each match, should be loaded dynamically)
- [ ] Setup npm + vue 3 environment, e.g., by using vscode + ventur
- [ ] Develop overview page with joined information of all plagiarism
- [ ] Develop detailed viewer for different matches
- [ ] Profit!

An alternative approach would be to create a new REST interface and to ask a (also not yet developed) JPlag server for details on demand. This would replace the need of JSON-file based generation but also require new interfaces. Also, JPlag is currently not used in a way that it runs in the background but to generate reports once and then stop again. This shall be discussed in advance.

tsaglam:
@cholakov11

tsaglam:
The max similarity should be already provided in the current result object. Thus it can be used in the new report generation to additionally display the top n matches according to max similarity.

cholakov11:
I have some questions regarding the old report and the API, since it is lacking a java doc.
1. What is the difference between program and submission, both are mention in the old html report?
2. What is the difference between _getComparisons_ and _getAllComparisons_ functions of a _JPlagResult_?
3. What is the difference between language and language option classes. I was able to find a couple of name representations of the language java9, JAVA_1_9, Javac 1.9+ based AST plugin, java19. Which one should be relevant for the report?
4. What are the language number of tokens? Are they relevant for the end report?
5. Is max number of matches, from the options class, all matches including those under the threshold?
[...]

Issue #248, opened 2021-12-02
Missing information in JPlagResult and JPlagOptions objects


cholakov11:
The following information is missing in the JPlagResult/JPlagOptions object but it could be useful for the report:

## JPlagResult
- [x] **List of failed submissions.** Through the list size their amount can also be acquired.

## JPlagOption
- [x] **List of names of excluded files.** The object contains name of excluded file but it needs to be read in order to obtain the names. However the file is already being read one time when starting the JPlag comparison. I would suggest this read to be stored so we can avoid duplicated coded and multiple same file accesses at different places in the code.

GabrielL:
For the second part, I have started this https://github.com/GabrielL/JPlag/pull/1 but since in my actual PR #272 there is a lot of refactoring around these parts, it is for the moment based on it, (hence the local pull request on my personal fork instead of here)

tsaglam:
I just took your commits from GabrielL#1 and put them on top of #273 (author-preserving) in order to have everything in the same PR for this issue!

tsaglam:
@cholakov11 Failed submissions can now be accessed via the result object, which contains the submission set, while the excluded file names are currently available through the options, which is temporary and might change due to #274. I hope this helps with the new report generation!

PR #287, opened 2022-02-09
Replace the outdated HTML report generation #192

cholakov11:
Until now the JPlag report functionality was creating report html files by generating "hard coded" html code. In this pull request changes to the report generation as well as report presentation were introduced.

Report generation
----
The report generation was moved from generating "hard coded" html to generating JSON files which represent the report. An _overview.json_ file is generated which contains the overall information of a _JPlagResult_ as well as a _.json_ file for each comparison done in the JPlag comparison process.
DTO classes _(de.jplag.reportingV2.reportobject.model)_ are introduced which are used to model the JSON files. _ReportObjectFactory_ is a factory class which takes a _JPlagResult_ object and converts it to a _JPlagReport_ consisting of an OverviewReport and a list of _ComparisonReport_ objects.
The _Jackson_ library is then used to generate json strings from the given _JPlagReport_. The generated strings are then saved as files on the hard drive. 
The _.json_ files can be used by any program which can display _.json_ files or a JPlag report.

Report presentation
----
A new JPlag report viewer application is introduced in the form of a Vue + Typescript application. The application can be started on local host from its folder _report-viewer_ and accepts _.zip_ files containing overveiw and comparison json files, single json files or searches for report files in its local _files_ folder. 
When receiving the _.json_ files the application convets them to Typescript objects (_Overview.ts_, _Comparison.ts_) and gives these object to the Vue components to be displayed.
The report viewer consists of three types of pages - a FileUpload page which the entry point to the application and where files to display are specified, an Overview page which display the overall information of a report and a Comparison page for each comparison. 
The report viewer supports: 
- displaying of overall report information ( folder, submission ids, language, etc.)
- displaying used metrics and their information ( threshold, distribution diagram)
- displaying comparison list sorted by similarity
- displaying of files in a comparison
- displaying matches in a comparison with navigation 
- code highlighting
- anonymization of ids
- displaying of clusters (needs further discussion and refinement)


dfuchss:
@here this is a PR to jplag:192-webreport and not to jplag:master :) 
Therefore, it might be hard to see the actual changes which this PR might introduce :)

sebinside:
@cholakov11 Please resolve the merge conflicts, then I will start my review

cholakov11:
@sebinside  I think I resolved the merge conflict however I am getting build failed on mvn spotless: check. When I run it locally I don't get any errors. Any suggestions?

dfuchss:
@cholakov11 you have to fix the line endings in the jplag/pom.xml

`git add --renormalize .` should do the job :)

jplag/src/main/java/de/jplag/reportingV2/jsonfactory/JsonFactory.java
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Uses Jackson to create Json Strings from JPlagReport object.
+     * @return A list, first element is Json String of Overview object. The rest elements are Json Strings of Comparison
+     * objects.
+     */
+    public static List<String> getJsonStrings(JPlagReport jPlagReport) {
+        List<String> jsonReports = new ArrayList<>();
+        try {
+            ObjectMapper mapper = new ObjectMapper();
```
dfuchss, line 27: you could use a static ObjectMapper

jplag/src/main/java/de/jplag/reportingV2/jsonfactory/JsonFactory.java
```diff
@@ -0,0 +1,60 @@
...
+        return jsonReports;
+    }
+
+    /**
+     * Creates Json Files for the given JPlagReport and saves them in the given folder.
+     * @return A boolean, representing whether the process was successful.
+     */
+    public static boolean saveJsonFiles(JPlagReport jPlagReport, String folderPath) {
+        String sanitizedFolderPath = folderPath;
+        if (!folderPath.endsWith("/")) {
```
dfuchss, line 44: You assume linux paths

dfuchss: Maybe simply use [Path](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#relativize-java.nio.file.Path-)

jplag/src/main/java/de/jplag/reportingV2/reportobject/ReportObjectFactory.java
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * Generates detailed ComparisonReport DTO for each comparison in a JPlagResult.
+     * @return A list with ComparisonReport DTOs.
+     */
+    private static List<ComparisonReport> generateComparisonReports(JPlagResult result) {
+        List<ComparisonReport> comparisons = new ArrayList<>();
+        result.getComparisons().forEach(c -> comparisons.add(new ComparisonReport(c.getFirstSubmission().getName(), c.getSecondSubmission().getName(),
+                c.similarity(), getFilesForSubmission(c.getFirstSubmission()), getFilesForSubmission(c.getSecondSubmission()), c.getMatches().stream()
+                        .map(m -> convertMatchToReportMatch(c, m, result.getOptions().getLanguage().usesIndex())).collect(Collectors.toList()))));
```
dfuchss, line 68: I would suggest to replace the lambdas by non-lambda functions/methods .. that may be more readable :)

jplag/src/main/java/de/jplag/reportingV2/reportobject/ReportObjectFactory.java
```diff
@@ -0,0 +1,209 @@
...
+     * @param usesIndex Indicates whether the language uses indexes.
+     * @return A Match DTO.
+     */
+    private static Match convertMatchToReportMatch(JPlagComparison comparison, de.jplag.Match match, Boolean usesIndex) {
+        TokenList tokensFirst = comparison.getFirstSubmission().getTokenList();
+        TokenList tokensSecond = comparison.getSecondSubmission().getTokenList();
+        Token startFirstToken = tokensFirst.getToken(match.getStartOfFirst());
+        Token endFirstToken = tokensFirst.getToken(match.getStartOfFirst() + match.getLength() - 1);
+        Token startSecondToken = tokensSecond.getToken(match.getStartOfSecond());
+        Token endSecondToken = tokensSecond.getToken(match.getStartOfSecond() + match.getLength() - 1);
```
dfuchss, line 168: Use one naming convention.

E.g., `startTokenSecond`

jplag/src/main/java/de/jplag/reportingV2/reportobject/ReportObjectFactory.java
```diff
@@ -0,0 +1,209 @@
...
+
+    // TODO implement after PR Readd clustering #281
+    private static List<Cluster> getClusters(JPlagResult result) {
+        // List<ClusteringResult<Submission>> clusters = result.getClusteringResult();
+        // return clusters.map( c -> new Cluster(getAvgSimilarity, getStrength, c.getMembers().map(Submission::getName)))
+        return List.of();
+    }
+
+    private static List<String> readFileLines(File file) {
+        ArrayList<String> lines = new ArrayList<>();
```
dfuchss, line 187: Replace static type with List

jplag/src/main/java/de/jplag/reportingV2/reportobject/model/Cluster.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.List;
+
+public class Cluster {
+    private final float average_similarity;
+    private final float strength;
+    private final List<String> members;
+
+    public Cluster(float average_similarity, float strength, List<String> members) {
+        this.average_similarity = average_similarity;
```
dfuchss, line 11: underscore is not aligned with Java Naming Conventions

jplag/src/main/java/de/jplag/reportingV2/reportobject/model/ComparisonReport.java
```diff
@@ -0,0 +1,72 @@
+package de.jplag.reportingV2.reportobject.model;
+
+import java.util.List;
+
+public class ComparisonReport {
+
+    private String first_submission_id;
```
dfuchss, line 7: see above

jplag/src/main/java/de/jplag/reportingV2/reportobject/model/FilesOfSubmission.java
```diff
@@ -0,0 +1,30 @@
...
+    public List<String> getLines() {
+        return lines;
+    }
+
+    public void setFile_name(String file_name) {
+        this.file_name = file_name;
+    }
+
+    public void setLines(List<String> lines) {
+        this.lines = lines;
```
dfuchss, line 28: setter should copy the list

jplag/src/main/java/de/jplag/reportingV2/reportobject/model/JPlagReport.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.util.List;
+
+public class JPlagReport {
+    private final OverviewReport overviewReport;
+    private final List<ComparisonReport> comparisons;
+
+    public JPlagReport(OverviewReport overviewReport, List<ComparisonReport> comparisons) {
+        this.overviewReport = overviewReport;
+        this.comparisons = comparisons;
```
dfuchss, line 11: you should copy lists. (if comparison results are mutable, you should also copy the elements of the list)

jplag/src/main/java/de/jplag/reportingV2/reportobject/model/Match.java
```diff
@@ -0,0 +1,80 @@
...
+    public Match(String first_file_name, String second_file_name, int start_in_first, int end_in_first, int start_in_second, int end_in_second,
+            int tokens) {
+
+        this.first_file_name = first_file_name;
+        this.second_file_name = second_file_name;
+        this.start_in_first = start_in_first;
+        this.end_in_first = end_in_first;
+        this.start_in_second = start_in_second;
+        this.end_in_second = end_in_second;
+        this.tokens = tokens;
```
dfuchss, line 22: See above

jplag/src/main/java/de/jplag/reportingV2/reportobject/model/Metric.java
```diff
@@ -0,0 +1,33 @@
...
+    private final float threshold;
+    private final List<Integer> distribution;
+    private final List<TopComparison> topComparisons;
+
+    public Metric(String name, float threshold, List<Integer> distribution, List<TopComparison> topComparisons) {
+        this.name = name;
+        this.threshold = threshold;
+        this.distribution = distribution;
+        this.topComparisons = topComparisons;
+    }
```
dfuchss, line 16: See comments on lists. (I won't mark this issue for the other classes)

jplag/src/main/java/de/jplag/reportingV2/reportobject/model/OverviewReport.java
```diff
@@ -0,0 +1,146 @@
...
+public class OverviewReport {
+
+    private String submission_folder_path;
+    private String base_code_folder_path;
+    private String language;
+    private List<String> file_extensions;
+    private List<String> submission_ids;
+    private List<String> failed_submission_names;
+    private List<String> excluded_files;
+    private int match_sensitivity;
```
dfuchss, line 14: Naming. I won't mark this issue for the other classes from now on.

jplag/src/main/java/de/jplag/reportingV2/reportobject/model/TopComparison.java
```diff
@@ -0,0 +1,38 @@
+package de.jplag.reportingV2.reportobject.model;
```
dfuchss, line 1: package reportingV2 should be reporting (in my opinion)

report-viewer/package.json
```diff
@@ -0,0 +1,52 @@
+{
```
dfuchss, line 1: You may want to add node/npm to the dependabot.yml to ensure updates.

report-viewer/src/App.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+  <router-view/>
+</template>
+
+<style>
+@font-face {
+  font-family: "JetBrains Mono";
+  src: url("./fonts/jetbrains-mono/JetBrainsMono-Light.ttf");
```
dfuchss, line 9: What is the license of the fonts you've added to the fonts folder?

dfuchss (changes requested):
I've added some further comments

report-viewer/src/App.vue
tsaglam: I think it has the [[SIL Open Font License 1.1](https://github.com/JetBrains/JetBrainsMono/blob/master/OFL.txt)](https://www.jetbrains.com/lp/mono/#license).

jplag/src/main/java/de/jplag/reportingV2/jsonfactory/JsonFactory.java
cholakov11: Agree on that, paths are currently platform dependant. I will fix this.

jplag/src/main/java/de/jplag/reportingV2/reportobject/ReportObjectFactory.java
cholakov11: This is actually a copied method from another part of JPlag due to missing attributes. See #248 . It won't be present in the final version.

jplag/src/main/java/de/jplag/reportingV2/reportobject/model/Cluster.java
cholakov11: I left the attributes with an underscore naming in order to easily match them to the json generated file. I will look up the Jackson annotaions. I believe there was an annotation to have different names for the same attribute in java and json.

jplag/src/main/java/de/jplag/reportingV2/reportobject/model/TopComparison.java
cholakov11: The old reporting package is still present and I didn't want to mess up the both packages that is why I created it in another one. @tsaglam @sebinside  can you also provide opinion on this question?

report-viewer/src/App.vue
cholakov11: It has SIL Open Font Licence. See https://www.jetbrains.com/lp/mono/#license.

jplag/src/main/java/de/jplag/reportingV2/reportobject/ReportObjectFactory.java
```diff
@@ -0,0 +1,209 @@
...
+        return overviewReport;
+    }
+
+    /**
+     * Generates detailed ComparisonReport DTO for each comparison in a JPlagResult.
+     * @return A list with ComparisonReport DTOs.
+     */
+    private static List<ComparisonReport> generateComparisonReports(JPlagResult result) {
+        List<ComparisonReport> comparisons = new ArrayList<>();
+        result.getComparisons().forEach(c -> comparisons.add(new ComparisonReport(c.getFirstSubmission().getName(), c.getSecondSubmission().getName(),
```
tsaglam, line 66: either use the standard identifier `it` or use a non-abbreviated one.

tsaglam (changes requested):
@cholakov11 here is some more feedback for you.
If you fix an issue that was mentioned in a review, please resolve the corresponding conversation so we know what is fixed and what isn't. Conversations, where you left a comment for us to read, can be left unresolved.

report-viewer/package.json
sebinside: Also: Please ensure that you use current versions in your `package.json`. The Installation already yields several vulnerabilities:
```
found 30 vulnerabilities (2 low, 22 moderate, 6 high)
  run `npm audit fix` to fix them, or `npm audit` for details
```

report-viewer/src/views/ComparisonView.vue
```diff
@@ -0,0 +1,209 @@
...
+     * Name of the comparison file. Comparison files should be named {ID1}-{ID2}
+     * @type {string}
+     */
+    const fileName = props.id1.concat("-").concat(props.id2)
+
+    let comparison;
+    //getting the comparison file based on the used mode (zip, local, single)
+    if(store.state.local) {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      comparison = ComparisonFactory.getComparison(require(`../files/${fileName}.json`))
```
sebinside, line 59: This line breaks the build / serve process, e.g., if there are no report files.

jplag/src/main/java/de/jplag/reportingV2/JsonReport.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public List<String> getReportStrings(JPlagResult result) {
+        JPlagReport report = ReportObjectFactory.getReportObject(result);
+        return JsonFactory.getJsonStrings(report);
+    }
+
+    @Override
+    public boolean saveReport(JPlagResult result) {
+        JPlagReport report = ReportObjectFactory.getReportObject(result);
+        return JsonFactory.saveJsonFiles(report, "./report-viewer/src/files/");
```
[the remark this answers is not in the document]
tsaglam: `  -r R                   Name of the directory in which the comparison results will be stored (default: result)`

jplag/src/main/java/de/jplag/reportingV2/JsonReport.java
```diff
@@ -0,0 +1,29 @@
...
+import de.jplag.reportingV2.jsonfactory.JsonFactory;
+import de.jplag.reportingV2.reportobject.ReportObjectFactory;
+import de.jplag.reportingV2.reportobject.model.JPlagReport;
+
+// ReportImplementation -> JsonReport
+
+/**
+ * A report generator which reports the JPlagResult in Json format.
+ */
+public class JsonReport implements Report {
```
sebinside, line 15: Currently, this report is not used by Jplag? In the course of replacing the old reporting code you could also insert the new reference in the main class

jplag/src/main/java/de/jplag/reportingV2/JsonReport.java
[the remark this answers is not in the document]
sebinside: Oh and bonus tip: `./` breaks iff you use windows :)

jplag/src/main/java/de/jplag/reportingV2/reportobject/model/TopComparison.java
tsaglam: But this means this PR should ensure your new report generation is called in the main class and thus default behavior!

report-viewer/src/views/ComparisonView.vue
```diff
@@ -0,0 +1,209 @@
...
+    id2: {
+      type: String,
+    },
+  },
+  setup(props) {
+    /**
+     * Name of the comparison file. Comparison files should be named {ID1}-{ID2}
+     * @type {string}
+     */
+    const fileName = props.id1.concat("-").concat(props.id2)
```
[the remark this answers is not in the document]
sebinside: Also: Add some error handling. The user shall not see a white page without any text if the URL is incorrect ;)

report-viewer/src/views/Overview.vue
```diff
@@ -0,0 +1,235 @@
...
+      overview = OverviewFactory.getOverview(JSON.parse(store.state.fileString))
+    }
+
+    /**
+     * Handles the selection of an Id to anonymize.
+     * If all submission ids are provided as parameter it hides or displays them based on their previous state.
+     * If a single id is provided it hides all of the other ids except for the chosen one.
+     * @param id
+     */
+    const handleId = (id) => {
```
sebinside, line 83: The representation of anonymized/non-anonymized IDs is still missing.

cholakov11:
@tsaglam @sebinside @dfuchss I might need some assist here. I updated the project to the latest state and now I am missing classes in the de.jplag.text.Parser.java. TextLexer and TextParser are not found. When I try to reload project I get "Cannot resolve plugin" for:
org.apache.maven.plugins:maven-javadoc-plugin
org.apache.maven.plugins:maven-source-plugin
org.apache.maven.plugins:maven-gpg-plugin
Any suggestions?

dfuchss:
Do you use intellij?
If so, maybe that's the issue: https://stackoverflow.com/questions/20496239/maven-plugins-can-not-be-found-in-intellij

cholakov11:
@dfuchss Thanks for the quick answer. I took some time to respond because I was trying the solutions mentioned in the link you sent. None actually helped. In the end I had to specify versions of the missing plugins in the pom. This fixed the can't resolve plugin errors. I will be ignoring the pom from now on when commiting. However the error with missing TextLexer and TextParser still remains.

dfuchss:
@cholakov11  Since `mvn package` works, that seems to be an IntelliJ bug.

One bug is also described here https://youtrack.jetbrains.com/issue/IDEA-190202


Regarding Text Parser & Text Lexer:
image
I can't reproduce the problem. I've used your current branch and everything regarding the two classes seems to be good.

cholakov11:
@dfuchss I played a little bit more with the settings, some manual deletions and finally I was able to fix it. Not sure what actually did the job but I'll take the win. I also don't need to specify the versions now. Big thank you for the help!

jplag/src/main/java/de/jplag/reportingV2/JsonReport.java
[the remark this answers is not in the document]
cholakov11: So I removed the hardcoded path and left it as parameter. I believe we can now use the path after the -r flag and give it directly to the report generator. A question and suggestion - Is report generation without specifying path possible through the CLI? And if it so, is it a good idea to provide a default folder to store the result if such is not provided?

report-viewer/src/views/Overview.vue
cholakov11: @sebinside Can you provide some more detail?

jplag/src/main/java/de/jplag/reportingV2/JsonReport.java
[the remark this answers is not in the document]
sebinside: That is a good question, I thought we have a default folder for this case...

report-viewer/src/views/Overview.vue
sebinside: @cholakov11 I meant that one can click on a submission in the ID list to anonymize all others, but there is no visible feedback which are currently shown. E.g., in the attached image, one cannot see that only A is selected
image

report-viewer/src/components/DistributionDiagram.vue
```diff
@@ -0,0 +1,146 @@
...
+    //Highest count of submissions in a percentage range. We set the diagrams maximum shown value to maxVal + 5,
+    //otherwise maximum is set to the highest count of submissions and is one bar always reaches the end.
+    let maxVal = ref(Math.max(...props.distribution));
+
+    let chartData = ref( {
+        labels: ['0-10%', '11-20%', '21-30%', '31-40%', '41-50%', '51-60%', '61-70%', '71-80%', '81-90%', '91-100%'],
+        datasets: [{
+          label: 'Count',
+          data: props.distribution,
+          backgroundColor: 'rgba(149, 168, 241, 0.5)',
```
sebinside, line 34: Can this be derived from the main color definitions?

report-viewer/src/components/DistributionDiagram.vue
```diff
@@ -0,0 +1,146 @@
+<template>
```
sebinside, line 1: General remark: Please write 1-2 sentences of documentation for each component and view to briefly explain what it does.

report-viewer/src/components/LineOfCode.vue
```diff
@@ -0,0 +1,83 @@
...
+pre {
+  /* In hljs pre tag is used. We need this style to override attributes in the styles of higlightjs to show code more
+   compact */
+  margin: 0;
+  padding: 0;
+  float: left;
+}
+
+
+.first-line {
```
sebinside, line 70: `.first-line` and `.last-line` should not be the job of the lines but belong to the surrounding container

report-viewer/src/libs/highlightjs/styles/edited/default.min.css
```diff
@@ -0,0 +1,82 @@
+/*!
+  Theme: Default
+  Description: Original highlight.js style
+  Author: (c) Ivan Sagalaev <maniac@softwaremaniacs.org>
+  Maintainer: @highlightjs/core-team
+  Website: https://highlightjs.org/
+  License: see project LICENSE
```
sebinside, line 7: If we embed others people code we shall not violate their license. Also: why is a copy of the CSS file required here?

report-viewer/src/store/store.ts
```diff
@@ -0,0 +1,62 @@
...
+             */
+            zip: Boolean,
+            /**
+             * Indicates whether single file mode is used.
+             */
+            single: Boolean,
+            /**
+             * Files string if single mode is used.
+             */
+            fileString: String
```
sebinside, line 32: I dont't know whether vuex supports optional parameters, but if this is the case, they could be used (also) here

report-viewer/src/router/index.ts
```diff
@@ -0,0 +1,33 @@
...
+    path: "/overview",
+    name: "OverviewView",
+    component: Overview,
+  },
+  {
+    path: "/comparison",
+    name: "ComparisonView",
+    component: ComparisonView,
+    props: route => ({
+      id1: route.query.id1,
```
sebinside, line 22: It's known as bad practice to name variables nearly identical, i.e. `id1` and `id2`. A spelling mistake is made quickly and then hard to find. Better use something like `firstId`.

report-viewer/src/router/index.ts
```diff
@@ -0,0 +1,33 @@
...
+    component: FileUpload
+  },
+  {
+    path: "/overview",
+    name: "OverviewView",
+    component: Overview,
+  },
+  {
+    path: "/comparison",
+    name: "ComparisonView",
```
sebinside, line 19: Only the `ComparisonView` View has the word `View` in his name. Should be consistent.

report-viewer/src/model/Match.ts
```diff
@@ -0,0 +1,10 @@
+export type Match = {
```
sebinside, line 1: Doc missing

report-viewer/src/model/Metric.ts
```diff
@@ -0,0 +1,8 @@
+import {ComparisonListElement} from "./ComparisonListElement";
+
+export type Metric = {
```
sebinside, line 3: Doc missing

report-viewer/src/views/ComparisonView.vue
```diff
@@ -0,0 +1,225 @@
...
+                    @line-selected="showMatchInSecond"/>
+    <FilesContainer :matches="comparison.matchesInSecondSubmissions" :files="filesOfSecond" :container-id="2"
+                    files-owner="Submission 2"
+                    @toggle-collapse="toggleCollapseSecond"
+                    @line-selected="showMatchInFirst"/>
+  </div>
+</template>
+
+<script>
+import { defineComponent, ref } from "vue";
```
sebinside, line 29: This also looks like a wrong ESLint setting (or ESLint is not running at all). Formatting should be consistent.

sebinside (changes requested):
This is what i found from having another look at the vue code. Overall, pretty solid and well modularized.

report-viewer/src/views/ComparisonView.vue
```diff
@@ -0,0 +1,225 @@
...
+     */
+    const fileName1 = props.id1.concat("-").concat(props.id2)
+    const fileName2 = props.id2.concat("-").concat(props.id1)
+
+    let comparison;
+    //getting the comparison file based on the used mode (zip, local, single)
+    if(store.state.local) {
+      try {
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        comparison = ComparisonFactory.getComparison(require(`../files/${fileName1}.json`))
```
sebinside, line 62: Again, good question whether `require` is the correct way to load json files at runtime...

report-viewer/src/assets/double_arrow_black_18dp.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" enable-background="new 0 0 20 20" height="18px" viewBox="0 0 20 20" width="18px" fill="#000000"><g><rect fill="none" height="20" width="20"/></g><g><g><path d="M15.79,9.71l-2.97-4.16C12.57,5.2,12.18,5,11.76,5c-1.06,0-1.68,1.2-1.06,2.07L12.79,10l-2.09,2.93 c-0.62,0.87,0,2.07,1.06,2.07c0.42,0,0.82-0.2,1.06-0.55l2.97-4.16C15.92,10.12,15.92,9.88,15.79,9.71z"/><path d="M10.79,9.71L7.82,5.55C7.57,5.2,7.18,5,6.76,5C5.69,5,5.07,6.2,5.69,7.07L7.79,10l-2.09,2.93C5.07,13.8,5.69,15,6.76,15 c0.42,0,0.82-0.2,1.06-0.55l2.97-4.16C10.92,10.12,10.92,9.88,10.79,9.71z"/></g></g></svg>
```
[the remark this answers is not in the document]
cholakov11: https://github.com/google/material-design-icons, their license is Apache-2.0 License

report-viewer/src/views/ComparisonView.vue
cholakov11: Well I couldn't find any alternative way except for _fs_, for which I have read that is server-side and that is why I avoided it.

report-viewer/src/components/DistributionDiagram.vue
cholakov11: I could not access the CSS variables from here.

report-viewer/src/libs/highlightjs/styles/edited/default.min.css
cholakov11: I have marked the attributes which I have edited in the local copy in order to keep my styling of code while maintaining the highlighted words. If I use highlight.js styles directly from the import I cannot override css and the the lines of code have large font, margins, paddings, don't use my font and have their own background which prevents the match colors from displaying properly.

report-viewer/src/components/LineOfCode.vue
cholakov11: Well the LineOfCode component is actually the container which takes the line and the information about it and displays it. Maybe renaming LineOfCode to LineOfCodeContainer could improve understandability.

report-viewer/src/assets/double_arrow_black_18dp.svg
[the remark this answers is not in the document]
sebinside: It seems that we're lucky then: https://www.apache.org/licenses/GPL-compatibility.html

report-viewer/src/views/ComparisonView.vue
sebinside: I wonder whether this would be actually a problem here. Not sending files to server is only a real problem when dealing with files that are drag 'n' dropped by the user client-side (i.e., zip files). Files that are read from the system are already on the servers side. But maybe I'm wrong here, please correct me.

report-viewer/src/libs/highlightjs/styles/edited/default.min.css
sebinside: Maybe I'm wrong but you can always extend and overwrite CSS, e.g. by defining own styles marked with `!important`.

sebinside: See: https://www.w3schools.com/css/css_important.asp

sebinside:
@cholakov11 Thank you for implementing the change requests. In my opinion, there are only three things left:
- Resolve the `require` discussion (if there is no easy solution, I would accept the current version)
- Resolve the CSS discussion (should be easy to fix, or I am completely wrong)
- Resolve the merge conflicts

Then, I'll be happy to merge this pull request! 🚀

cholakov11:
@sebinside Conflicts are fixed, ready to merge.

sebinside (approved):

Issue #206, opened 2021-10-25
Generalize submission path specification


Alberth289346:
The current input path specification of `root/*/dir` doesn't quite work for us, as we have a `course/year` directory tree, with several (but possibly not all) years to be included in the check.

Also, looking a bit ahead, we see the need for specifying multiple sets tomorrow, to add support for already checked sets, ie along the lines of #91 or #49 .
As such, this is the first step towards a more unified and general way to specify submission sets  than the current `<root-dir>`, `-S sub`, and `-bc` option triplet together.

----
The proposal for generalizing submission path specification is to allow having one or more file-system paths instead of the current single `<root-dir>` to specify a set submissions, with general `*` globbing (we'd use `course/202*` for checking the last 2 years). Several file system paths allow us to have `course/202* course/2019` as set specification for checking the last 3 years.

As the form of the path isn't fixed, the currently accepted `<root-dir>/*/sub` path also fits in the proposal without needing the  `-S` option.

An extension to the above could be curly brace expansion, the above would become `course/20{2*,19}`. This is basically another form of writing more than one file-system path, not sure it's worth the effort of implementing it.

`-bc` base-code path specification can also be changed to the more general form then (for globbing to reduce typing mainly).
Not strictly needed at this point, as it's still one directory.

----
For tomorrow needs, a possible path seems to be (at this time):

1. Allow more than one set submissions
2. Allow some sets to be marked as "already checked"

Two new items then pop up.

1. _Each submission set may have a slightly different base code._
   I ignore the theoretical issue of how to deal with different base code for each set in the comparison.
   The other more practical issue is how to specify that comparison should not see the base code directory as a submission. (it makes sense to store it as `course/year/basecode`, but that's in the submission set path.)

   Having multiple `-bc` basecode options might work, except you get into trouble in selecting the version that should be used in the comparison.

   Another option is to specify one or more relative paths from the submission root. Makes it easier to attach a basecode to the correct submission set, but selecting the basecode to use in the comparison is still unsolved.

   Likely there are other options.

2. _Not all submission sets are equal any more, some are already checked, and don't need more checking._
   This means one has to say which of the above submission paths are done or not done yet. It needs at least two different path specifications, so one `<root-dir>` likely won't fly.
 
Thoughts?

@davbeek FYI

tsaglam:
Seems that we now support the suggestions made here, thus closed.

PR #288, opened 2022-02-10
Stacked-patch: Extend JPlag with checking against prior submissions

Alberth289346:
**Currently stacked onto "multi-root-cleanup" to avoid breaking patches.**
EDIT: Patch is considered finished, marked as draft to prevent accidentally merging only.

JPlag currently perform plagiarism checks for all combinations of submissions. However, as described in #49 and #91 there is a wish to extend checks with an additional set of "old" submissions. These old submissions are also used in the plagiarism checks, but checks between two old submission are skipped.
This improves speed of the checking process, and reduces clutter in the result, as plagiarism between old submissions is not reported.  
This patch implements adding such an old submission set.

Notes:

- In the code I opted for descriptive names `plagiarismCheckRootDirectories` and `priorSubmssionsRootDirectories` to distinguish both sets.
- At the command-line, I opted for `-new` and `-old`, as that is reasonably short and clearly different from each other.
  For completeness the variants we considered were
  ``` text
  --new-submissions <foldername>

  --checked DIR and --unchecked DIR

  --checked-dir DIR and --check-dir DIR   imho too subtle difference

  thinking a sentence like "checking ..... submissions against .... submissions" and trying to find
  good words there. Perhaps "unchecked" and ("referenced" or "verified" or "done" or "old")?
  Don't like "unchecked" much, "unknown" perhaps?

  Without "unchecked", "checked" might be good at the second dots too.

  --old-dir and --target-dir
  --old sounds a bit negative to me, --existing

  --target we could have --new-dir

  Result:
  --old-dir and --new-dir
  ```

  Also there is `-archive` and `-prior` too from #49 and #91 .
  I somewhat added `-check` and `-prior` as notions in the code, but since most users don't read source code I am not sure how relevant that is.
  We should decide about these names, and adapt the code to that.

- Clearly, code names and option names are not equal, not sure if that's desired. Programmers and program users are mostly disjoint groups of people.

- #91 only added the `-prior` option, and adapted processing, so this PR can be closed I think.
- #49 does additional HTML output things, such changes may need considering for the new HTML report backend.
- #206 is implemented except for globbing support of root-directory paths.

Closes #238 
Adresses #49 #91 #206


tsaglam:
@Alberth289346 I did not close this purposely, I think GitHub [auto closed this because I deleted my branch](https://github.community/t/if-a-base-branch-is-deleted-does-it-close-all-active-prs-on-that-branch/1879) in the other PR?

PR #290, opened 2022-02-14
Improve token data structure (minor frontend overhaul)

tsaglam:
- Clean up frontend classes, deduplicate code, fix visibilities, remove unused code, avoid abbreviated variable names
- Clean up token constant interfaces by de-duplicating the file end and separator token
- Add more documentation to the frontend-utils classes
- Enriches the C/C++ and Scheme frontend tokens with column index and token length information
- Remove spaces from the textual representation of the Java tokens
- Use `EOF` as the end of file token representation
- Reduce the number of realizations of the token constants interfaces

_Comment: I will merge this after the larger feature PRs that are currently in review_


jplag.frontend-utils/src/main/java/de/jplag/TokenConstants.java
```diff
@@ -1,7 +1,14 @@
...
 
+/**
+ * Super interface for all language parsers. Provides the only two token that are required from every parser.
+ */
 public interface TokenConstants {
+    /**
+     * Marks the end of the file, has a special purpose in the comparison algorithm.
+     */
     final static int FILE_END = 0;
+
```
Alberth289346, line 11: It might be useful to explicitly have a constant denoting the first value available to a frontend. In that way you can add more tokens here without breaking all frontends.

tsaglam: What do you mean by "first value available to a frontend"?

tsaglam: Ah, I think I understand. Every int is fine except 0 and 1. Thus I do not really think we need a constant here.
The core problem is that this interface hierarchy in combination with the integer constants is not an ideal solution.
It would be cleaner using enums that implement a marker interface.

tsaglam: Yes, the `Parser implements TokenConstants` pattern is weird, and there is probably no need to implement the constants anyway as a parser. The frontends are still overwhelmingly legacy code and thus not ideal. However, this PR is not meant to fix all frontend-related issues.

My suggestion was to define a interface with a method signature such as `getTokenType()`. And then each frontend can implement an enum for the token types, where each token type is an enumeral with a type id and the string representation (thus replacing the weird `type2string` methods). The method signature of the interface then can describe a contract where the id 0 and 1 are set, but all other ids are free to choose by the implementing enums.

Anyway, this is getting a bit off-topic, as this is not really the intent of this PR.

PR #316, opened 2022-03-10
Fixes Routing

### LINKED ISSUE #310: Fix Automatic JPlag Report Viewer Deployment
Since #287, we have a new web-based report viewer that is automatically deployed at https://jplag.github.io/JPlag/. However, it does not load correctly. This is probably due to a Vue router bug or something ™️ .

dfuchss:
Resolves #310


dfuchss:
See https://dfuchss.github.io/JPlag/

PR #332, opened 2022-03-29
Simplified versioning

dfuchss:
Created a maven property: Future releases would need to change only one position to bump the version of JPlag the property in the main pom.xml


dfuchss:
This should be tested in a snapshot deployment :)

tsaglam (approved):

tsaglam (approved):

PR #333, opened 2022-03-29
Changed all dependencies to the name of the main branch

dfuchss:
Since most repositories (from my point of view :) ) are moving to `main` as new default branch. I've prepared a possible transition to a new default branch.


tsaglam:
@sebinside and I already talked about this internally, we will probably do this as part of a larger change in branch management. However, timing is important here. Also, this is something we should announce (e.g. via discussions) in advance.

PR #347, opened 2022-04-05
Change antlr.preprocessor.Preprocessor to ClusteringPreprocessor.

Alberth289346:
While building an executable I ran into an `antlr.preprocessor.Preprocessor` import that is only used in the comment below. Antlr is still used in a few frontends but not in the main program code anymore it seems.

I think it should be the new `ClusteringPreprocessor` class, as all processors there derive from it.


tsaglam (approved):
You are right! Good Find!

Issue #193, opened 2021-10-12
Develop an end-to-end testing strategy


tsaglam:
We currently have no way of telling whether a change made to JPlag affects the quality of plagiarism detection. This makes refactoring code hard because we never know when a PR has unintended consequences. The reason for this is that JPlag has not enough test cases. We thus need a carefully designed test framework, that uses different data sets and runs JPlag with different configuration options. The JPlag result object is then used to check if JPlag produces the expected results. Of course, when we first add the test cases, the expected results are the current results JPlag produces for these inputs.

The following initial requirements have been identified:
- Should cover different kinds/types/classes of plagiarism (could be based on existing research)
- Should provide multiple input sets (multiple source code submissions, some are plagiarisms some not) that are somewhat realistic
- Should test the functional qualities of JPlag (so we can see if future changes to JPlag make it better/worse), meaning the similarities and so on
- Should integrate with existing test cases in a well-rounded test suite
- (Hopefully increases source code coverage by using different configurations of the options)

Related research works:
- [Larsson 2021](http://urn.kb.se/resolve?urn=urn:nbn:se:lnu:diva-105235) references [Novak 2019](https://dl.acm.org/doi/10.1145/3313290)
- [Karnalim 2016](https://ieeexplore.ieee.org/abstract/document/7910274)
- [Devore-McDonald et al. 2020](https://dl.acm.org/doi/abs/10.1145/3428206) (less closely related)
- [Poon et al. 2012](https://dl.acm.org/doi/abs/10.1145/2325296.2325328) (less closely related)

Steps to do:
- [x] Carefully design a strategy (which inputs, which configurations of JPlag, what info of the result object is checked, how are the expected values stored?)
- [x] Discuss the strategy and the underlying design choices here
- [x] Find or create test data (source code submissions, preferably java)
- [x] Implement the test framework  based on the strategy and merge it into the master branch
- [ ] Document the contribution in the wiki

tsaglam:
@martin9295

tsaglam:
The technical report on JPlag contains different plagiarism classes, this might also be a starting point.

SuyDesignz:
First of all, I would like to explain the goal of the End-To-End testing strategy.
Here, changes to the detection of plagiarism in JPlage are to be tested. 
The main reason is to ensure consistent results when adapting the application.
Likewise I try to find out, with purposeful tests, in which recognition phase the change finds itself (optional and still in discussion).

[...]

PR #398, opened 2022-04-28
Ignore volume tests.

sebinside:
Ignore volume tests as they represent an earlier testing approach (and break sometimes :)). Should be removed in a future version.

An enhanced testing method shall be implemented in #193.


tsaglam (approved):
Massive contribution! Thank you for your work! 👌

PR #409, opened 2022-05-05
Add slf4j as logger and extracted CLI to own module

dfuchss:
Successor of #250 and #283 just a bit simpler :)

EDIT: Contributes towards #117.


jplag/src/main/java/de/jplag/SubmissionSetBuilder.java
```diff
@@ -216,7 +212,7 @@ private Submission loadBaseCodeViaName(String baseCodeName, File rootDirectory,
             throw new BasecodeException(String.format("Basecode path \"%s\" relative to the working directory could not be found.", baseCodeName));
         } else {
             // Found a base code as a submission, report about legacy usage.
-            System.out.println("Legacy use of the -bc option found, please specify the basecode by path instead of by name!");
+            logger.info("Legacy use of the -bc option found, please specify the basecode by path instead of by name!");
```
tsaglam, line 215: We should make this a warning, as we might not support the legacy backward compatibility forever.

tsaglam (approved):
Just one small thing but I can also change this.

Issue #345, opened 2022-04-04
Support more (modern) programming languages


tsaglam:
We want to extend the capabilities of JPlag to support more modern languages.
Thus I propose:

- Merging existing but outdated frontends from forks for Scala, R, and Jupyter Notebook (see #202)
    - Some of them might not be easily mergeable, thus this should be evaluated beforehand
- Creating new frontends based on [available ANTLR grammars](https://github.com/antlr/grammars-v4) for Kotlin, Go, and Rust
- Creating a small set of tests for each new frontend (related to #408 and #296)
- If there is time, modernizing the Python frontend (see #280 and for reference also #34)

For each frontend, a parser is required that parses the source code. Then, in an abstraction step, a subset of the tokens of the language AST need to be transformed into generic JPlag tokens. These tokens can be then passed as a token list to JPlag.

**Steps to take:**

- Understand how language frontends work and how they are designed (look at the C# and Java frontends)
- Incrementally, add new frontends (from forks or from scratch) and test them
- New frontends should contain a readme that documents what tokens are extracted and why
- Add documentation on how to create new frontends
- (Optionally) modernize the Python frontend

**Current state:**

- [x] Rlang
- [x] Kotlin
- [x] Golang
- [x] Scala
- [x] Rust

ijaureguialzo:
Hello, maybe it's not considered a "modern" language, but I would love to see SQL and PL/SQL supported by JPlag.

Thanks for the awesome work.

chr15t0ph:
I would love to see JavaScript supported by JPlag.

(I already use jplag for Java courses and would love to have it for JavaScript courses, too.)
Kind regards, Christoph

tsaglam:
I cannot promise that we will find time to add the languages requested here. But in the future, there will be comprehensive documentation on how to add language frontends with relatively low effort.

ijaureguialzo:
That would be nice. If there would be some kind of framework, I could spare some time and try to do it myself.

tsaglam:
@ijaureguialzo @chr15t0ph we added documentation regarding adding new languages to our wiki:
https://github.com/jplag/JPlag/wiki/4.-Adding-New-Languages

tsaglam:
Closed, as we now collect language requests in #685.

PR #451, opened 2022-06-09
Add R language frontend

robinmaisch:
Let's add R to JPlag! The R branch should be perfectly runnable.
If there is any trouble or suggestions, let me know.

Adressing Issues #75 #345


tsaglam:
Currently, this PR contains multiple newly added frontends. I would suggest one frontend per PR. This would include a short readme in the frontend and some test cases. Any other changes should be explained and well-argued for in the PR description. E.g. changes to the token printer. For the PR description, try to point out what you changed/added and why. Smaller bug fixes, like the fix in the ErrorConsumer, should be another small PR.

robinmaisch:
Got it, I'll work on these points right away.

robinmaisch:
@tsaglam In a new frontend, are the tests supposed to be disabled, like the tests in MinimalCSharpFrontendTest?

tsaglam:
No, they should be enabled. When you design meaningful tests (as we discussed recently), they should be enabled by default.

PR #452, opened 2022-06-10
Fix the -n argument for limiting the number of exported comparisons

tsaglam:
The CLI argument `-n` for limiting the exported comparisons was not working with the new report viewer.
The argument works as follows: 

```
  -n N           The maximum number  of  comparisons  that  will be
                 shown in the generated  report,  if  set to -1 all
                 comparisons will be shown (Standard: 30)
```

This PR fixes that by enforcing the limit during the JSON generation. On the overview page, the report viewer will still show all comparisons. However, only the top N can be clicked. Currently, there is no warning message when clicking a comparison that was not exported. We should add an error message here.

@jplag/studdev @jplag/maintainer Note that from now on the default of max. 30 comparisons is enabled again.

### DIFF
--- a/jplag/src/main/java/de/jplag/reporting/JsonReport.java
+++ b/jplag/src/main/java/de/jplag/reporting/JsonReport.java
@@ -24,9 +24,9 @@ public class JsonReport implements Report {
     @Override
     public boolean saveReport(JPlagResult result, String path) {
         JPlagReport report = ReportObjectFactory.getReportObject(result);
-        File dir = new File(path);
-        if (!dir.exists()) {
-            if (!dir.mkdir()) {
+        File directory = new File(path);
+        if (!directory.exists()) {
+            if (!directory.mkdir()) {
                 logger.error("Failed to create dir.");
             }
         }
--- a/jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
+++ b/jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
@@ -34,9 +34,11 @@ public class ReportObjectFactory {
 
     /**
      * Generates an Overview DTO of a JPlagResult.
+     * @param numberOfExportedComparisons
      */
     private static OverviewReport generateOverviewReport(JPlagResult result) {
-        List<JPlagComparison> comparisons = result.getComparisons();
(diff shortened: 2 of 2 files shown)


jplag/src/main/java/de/jplag/reporting/JsonReport.java
```diff
@@ -24,9 +24,9 @@ public List<String> getReportStrings(JPlagResult result) {
     @Override
     public boolean saveReport(JPlagResult result, String path) {
         JPlagReport report = ReportObjectFactory.getReportObject(result);
-        File dir = new File(path);
-        if (!dir.exists()) {
-            if (!dir.mkdir()) {
+        File directory = new File(path);
+        if (!directory.exists()) {
+            if (!directory.mkdir()) {
```
dfuchss, line 29:
```suggestion
            if (!directory.mkdirs()) {
```

jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -65,11 +67,13 @@ private static OverviewReport generateOverviewReport(JPlagResult result) {
 
     /**
      * Generates detailed ComparisonReport DTO for each comparison in a JPlagResult.
+     * @param numberOfExportedComparisons
```
dfuchss, line 70: Missing docs :)

jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -34,9 +34,11 @@ public static JPlagReport getReportObject(JPlagResult result) {
 
     /**
      * Generates an Overview DTO of a JPlagResult.
+     * @param numberOfExportedComparisons
```
dfuchss, line 37: Missing docs :)

dfuchss (changes requested):
Some minor comments :)

jplag/src/main/java/de/jplag/reporting/JsonReport.java
tsaglam: Good call, did not catch that!

dfuchss (approved):

dfuchss (approved):

nestabentum (approved):

Issue #357, opened 2022-04-11
Enhance the new Report Viewer


sebinside:
In #287, we added a new web-based [report viewer](https://jplag.github.io/JPlag/) that was developed from scratch. However, there are still many enhancements and fixes open until the viewer is release-ready for version 4.0:

## General
* [x] Include clustering in JSON files and in the frontend (see #281)
* [x] Include max metric in JSON files and in the frontend (see #281)
* [x] Revisit JSON file generation when comparing large submissions and remove code redundancy in JSON files
* [ ] Enhance logging and error reporting in both JSON generator and report viewer code
* [x] Add support for multiple roots (see #354)

### Fixes
* [x] Fix crashes of JSON file generation, e.g., if folders do not exist (see #354)
* [x] JSON file generation crashes with large submissions, probably requires incremental serialization

### Fixes
* [x] Remove JetBrains Mono Font Ligatures (see #355)
* [x] Investigate row numbers in report viewer being off by one (see #355)
* [x] Report viewer file comparison does sometimes not show matches (see #355)
* [x] Fix the directory path on the overview view being wrongly displayed
* [x] Anonymization has a very bad performance
* [x] Sometimes, the highlighting in the comparison does not work (see #573, additional investigation required)

### Enhancements
* [x] (Additionally) zip all files after the creation
* [ ] Switching between `avg` and `max` metric freezes the UI if there are many submissions in the overview
* [ ] <s>Built report viewer files shall be deployed with results (see #192)</s>
* [ ] Include and enforce eslint
* [ ] <s> Accept multiple files as input </s>
* [ ] Show a warning if some/all comparison files are missing
* [ ] Add a "back to overview" button to the comparison view
* [x] Reverse sorting order of the histogram (and fix intervals being closed)
* [ ] Change the color scheme of matches (something without implicit semantics)
* [ ] Enhance file comparison view: Find a way to better display the sidebar (currently quite hidden)
* [ ] Change comparison view file order, e.g., based on packages rather than alphabetical 
* [ ] Add button "collapse all" to the comparison view
* [ ] Highlight base code, e.g., gray
* [ ] Enable collapsing by clicking on the file header in the comparison view
* [ ] Dark mode (e.g., by user choice or even by default)
* [ ] Enhance overall frontend design/impression
* [ ] The CD [deployed version](https://jplag.github.io/JPlag/) should have sample files ready to go (added in the deployment pipeline)

tsaglam:
[...]

PR #453, opened 2022-06-10
Write Clusters To JSON Report

nestabentum:
Implementation of the upper most bullet point from #357.

- Adds averageSimilarity as an attribute to the Cluster class.
- In report creation clustering info is now included
- Removed ClusteringRealDataTest.java because of its limited use


jplag/src/main/java/de/jplag/clustering/ClusteringResult.java
```diff
@@ -92,11 +89,33 @@ public static ClusteringResult<Integer> fromIntegerCollections(List<Collection<I
...
+        var sumOfSimilarities = 0f;
+        var submissionIndicesWithoutIndicesAlreadyProcessed = new ArrayList<>(List.copyOf(cluster));
+        for (Integer indexOfSubmission1 : cluster) {
+            for (Integer indexOfSubmission2 : submissionIndicesWithoutIndicesAlreadyProcessed) {
+                if (!Objects.equals(indexOfSubmission1, indexOfSubmission2)) {
+                    sumOfSimilarities += similarityMatrix.getEntry(indexOfSubmission1, indexOfSubmission2);
+                }
+            }
+            submissionIndicesWithoutIndicesAlreadyProcessed.remove(indexOfSubmission1); // remove indices we already processed from second list to avoid adding them anew unnecessary
+        }
```
robinmaisch, line 108: If the dimensions of `similarityMatrix` are roughly equivalent to `cluster.size()`, I would consider calculating the sum using a `RealMatrixVisitor` instead, like so:
```suggestion
        double sumOfSimilarities = similarityMatrix.walkInOptimizedOrder(new DefaultRealMatrixPreservingVisitor() {
            double sum;

            @Override
            public void visit(int row, int column, double value) {
                if (row > column) sum += value;
            }

            @Override
            public double end() {
                return sum;
            }
        });
```
If, on the other hand, `similarityMatrix` is big and sparse, I suggest using the standard `for`-loop to iterate over the indexes in `cluster`, because you are essentially emulating that.
```suggestion
        var sumOfSimilarities = 0f;
        for (int index1 = 1; index1 < cluster.size(); index1++) {
            for (int index2 = 0; index2 < index1; index2++) {
                sumOfSimilarities += similarityMatrix.getEntry(index1, index2);
            }
        }
```
Both approaches save you from having to test for any special cases or managing the `submissionIndicesWithoutIndicesAlreadyProcessed`.

jplag/src/main/java/de/jplag/clustering/Cluster.java
```diff
@@ -15,17 +14,23 @@
     private final float communityStrength;
     private final Collection<T> members;
     private ClusteringResult<T> clusteringResult = null;
+    private final float averageSimilarity;
 
-    public Cluster(Collection<T> members, float communityStrength) {
+    public Cluster(Collection<T> members, float communityStrength, float averageSimilarity) {
```
tsaglam, line 19: Just for understanding: The average similarity of a cluster is the mean value of all submissions tuple similarities among the submissions in the cluster? Maybe add a JDoc comment for the constructor and clarify it in there. The author of the class did not specify one it seems.

jplag/src/main/java/de/jplag/reporting/reportobject/model/Cluster.java
```diff
@@ -4,32 +4,6 @@
...
-    }
-
-    public float getStrength() {
-        return strength;
-    }
-
-    public List<String> getMembers() {
-        return members;
-    }
+public record Cluster(@JsonProperty("average_similarity") float averageSimilarity, @JsonProperty("strength") float strength,
```
tsaglam, line 7: Great use of records!

tsaglam (changes requested):
Looks good. Just one small thing.

jplag/src/main/java/de/jplag/clustering/ClusteringResult.java
nestabentum: Hey Robin, thanks for your feedback! While this is not a fitting case for the `RealMatrixVisitor` I'm happy you told me about it, as I did not know it before.
Concerning changes to the loop: The streams approach is clever, I really like it. But I do share your concerns with readability. I do find your standard for-loop approach more precise and efficient - I will incorporate it.

jplag/src/main/java/de/jplag/clustering/Cluster.java
nestabentum: Yes, exactly. I added the JDoc.

tsaglam (approved):

PR #455, opened 2022-06-14
Add R language frontend

robinmaisch:
This PR contains a newly added frontend for the R language, complete with some tests and a brief README. Other submodules have only been altered to the extent necessary to make the R frontend usable.

Addressing issues #345, #451


jplag.frontend.R/src/main/java/de/jplag/R/RParserAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+    private boolean parseFile(File directory, String fileName) {
+        File file = new File(directory, fileName);
+        try (FileInputStream inputStream = new FileInputStream(file)) {
+            currentFile = fileName;
+
+            // create a lexer, a parser and a buffer between them.
+            RLexer lexer = new RLexer(CharStreams.fromStream(inputStream));
+            CommonTokenStream tokens = new CommonTokenStream(lexer);
+
+            RFilter filter = new RFilter(tokens);
```
tsaglam, line 64: Just out of interest, what does the R filter do?

jplag.frontend.R/src/main/java/de/jplag/R/RTokenConstants.java
```diff
@@ -0,0 +1,41 @@
...
+    int WHILE_BEGIN = 18;
+    int WHILE_END = 19;
+    int REPEAT_BEGIN = 20;
+    int REPEAT_END = 21;
+    int NEXT = 22;
+    int BREAK = 23;
+    int COMPOUND_BEGIN = 24;
+    int COMPOUND_END = 25;
+
+    int NUM_DIFF_TOKENS = 26;
```
tsaglam, line 40: Currently, this frontend does not have many tokens. From your own testing, do you think R is just less expressive than e.g. java or do you think this token set should be expanded in the future?

jplag.frontend.R/src/test/java/de/jplag/R/RFrontendTest.java
```diff
@@ -0,0 +1,113 @@
...
+     */
+    private void testSourceCoverage(String fileName, TokenList tokens) {
+        File testFile = new File(testFileLocation, fileName);
+
+        try {
+            List<String> lines = Files.readAllLines(testFile.toPath());
+            String commentExpression = getCommentExpression();
+
+            // All lines that contain code
+            var codeLines = IntStream.range(1, lines.size() + 1).filter(idx -> !lines.get(idx - 1).matches("\\s*(%s)?".formatted(commentExpression)))
```
tsaglam, line 72: Especially regex strings should be constants so they can be properly named. This ensures readability.

tsaglam: e.g. LINE_COMMENT

jplag/src/main/java/de/jplag/options/LanguageOption.java
```diff
@@ -13,6 +13,7 @@ public enum LanguageOption {
     PYTHON_3("python3", "de.jplag.python3.Language"),
     C_CPP("cpp", "de.jplag.cpp.Language"),
     C_SHARP("csharp", "de.jplag.csharp.Language"),
+    R("R", "de.jplag.R.Language"),
```
tsaglam, line 16: Every other display name is in lower case. For consistency, we should do the same here.

tsaglam (changes requested):
Looks good! Just a few minor things...

jplag.frontend.R/src/main/java/de/jplag/R/RParserAdapter.java
robinmaisch: This is what I could find out:
The RFilter accommodates for an oddity in the R syntax that is discussed [here](https://stackoverflow.com/a/13724135). 
Usually, `newline` tokens are not meaningful in R; they do not alter the AST. But when parsing if-else expressions, the common R parser is essentially not greedy enough to read the `else` at the end of the `if` clause when the following conditions are met:

- The if-else expression is top-level, i.e. not enclosed by any kind of block.
- The `else` keyword is not in the same line as the last symbol of the `if` clause.

So in that case, the `if` clause is parsed and then the following `else` keyword leads to an error.

The RFilter "hides" all non-top-level `newline` tokens to make sure that all ends of expressions are correctly detected.

jplag/src/main/java/de/jplag/options/LanguageOption.java
robinmaisch: I see. In that case, I think it would be helpful to make the CLI language argument case-insensitive by changing `equals` to `equalsIgnoreCase` in `LanguageOption.fromDisplayName(String)`, what do you think?

tsaglam: Sounds good!

jplag.frontend.R/src/main/java/de/jplag/R/RParserAdapter.java
tsaglam: Thanks for the explanation!

jplag.frontend.R/src/main/java/de/jplag/R/RTokenConstants.java
robinmaisch: AFAICT, R code is mostly just a sequence of function definition containing function calls and arithmetic operations. Syntactically, there really does not seem to be very much to do, and I would say that the current token set covers the elements given by the grammar quite well.

It turns out, though, that _some_ new syntactic elements have been introduced in R v4.0.0 and up, which are not represented in the [R grammar](https://github.com/antlr/grammars-v4/blob/master/r/R.g4) in the antlr4 repository yet, e.g. lambda functions `\(x) x*x` or the pipe operator `|>`. So, it is actually the grammar that should be expanded first if we want to continue to rely on that.

jplag.frontend.R/src/main/java/de/jplag/R/JplagRListener.java
```diff
@@ -0,0 +1,373 @@
...
+        parserAdapter.addToken(targetType, start.getLine(), start.getCharPositionInLine() + 1, end.getStopIndex() - start.getStartIndex() + 1);
+    }
+
+    @Override
+    public void enterProg(RParser.ProgContext ctx) {
+
+    }
+
+    @Override
+    public void exitProg(RParser.ProgContext ctx) {
```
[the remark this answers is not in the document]
robinmaisch: Done.

jplag.frontend.R/src/test/java/de/jplag/R/RFrontendTest.java
robinmaisch: Done.

jplag.frontend.R/README.md
```diff
@@ -0,0 +1,9 @@
+# JPlag R language frontend
+
+The JPlag R frontend allows the use of JPlag with submissions in R. 
+
+### R specification compatibility
+The underlying [grammar definition](https://github.com/antlr/grammars-v4/tree/master/r) was first created in June 2013, when R 3.0.1 was current. The latest commit is from April 2018, when R 3.5.0 was just released. Whether the grammar has been made to comply with any specific version of the R specification is unclear. Even if some parsing errors occur, the parser should be able to recover and still produce a valid analysis.
+
+### Usage
```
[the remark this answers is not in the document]
robinmaisch: Done.

tsaglam (approved):

dfuchss (changes requested):
To be discussed: We should maybe update the packages & artifact id to "rlang" or something similar.

jplag.frontend.R/src/main/antlr4/de/jplag/R/grammar/RFilter.g4
```diff
@@ -0,0 +1,83 @@
...
+
+parser grammar RFilter;
+
+options { tokenVocab=R; }
+
+@members {
+protected int curlies = 0;
+}
+
+// TODO: MAKE THIS GET ONE COMMAND ONLY
```
nestabentum, line 39: What is our stance on todos in checked in code? @tsaglam

jplag.frontend.R/src/main/java/de/jplag/R/RToken.java
```diff
@@ -0,0 +1,47 @@
+package de.jplag.R;
+
+/**
+ * This class represents the occurrence of an R Token in the source code. Based on an R frontend for JPlag v2.15 by Olmo
+ * Kramer, see their <a href="https://github.com/CodeGra-de/jplag/tree/master/jplag.frontend.R">JPlag fork</a>.
+ * @author Robin Maisch
```
nestabentum, line 6: What is our stance on Author annotations @tsaglam?

jplag.frontend.R/src/test/java/de/jplag/R/RFrontendTest.java
```diff
@@ -0,0 +1,112 @@
...
+    private void testTokenCoverage(TokenList tokens, String fileName) {
+        var foundTokens = StreamSupport.stream(tokens.allTokens().spliterator(), true).mapToInt(Token::getType).sorted().distinct().toArray();
+        // Exclude SEPARATOR_TOKEN, as it does not occur
+        var allTokens = IntStream.range(0, RTokenConstants.NUM_DIFF_TOKENS).filter(i -> i != TokenConstants.SEPARATOR_TOKEN).toArray();
+
+        if (allTokens.length > foundTokens.length) {
+            var diffLine = IntStream.range(0, allTokens.length)
+                    .dropWhile(lineIdx -> lineIdx < foundTokens.length && allTokens[lineIdx] == foundTokens[lineIdx]).findFirst();
+            diffLine.ifPresent(lineIdx -> fail("Token type %s was not found in the complete code example '%s'."
+                    .formatted(new RToken(allTokens[lineIdx], fileName, -1, -1, -1).type2string(), fileName)));
```
nestabentum, line 103: Magic numbers: -1 probably has a special meaning like e.g. undefined. Using variables with fitting names could convey this meaning.

nestabentum (commented):
Only some minor comments

jplag.frontend.R/src/main/java/de/jplag/R/RCombinedBaseListener.java
```diff
@@ -0,0 +1,342 @@
...
+import org.antlr.v4.runtime.ParserRuleContext;
+import org.antlr.v4.runtime.tree.ErrorNode;
+import org.antlr.v4.runtime.tree.TerminalNode;
+
+import de.jplag.R.grammar.*;
+
+/**
+ * Empty base implementation for {@link RListener} and {@link RFilterListener}.
+ */
+public class RCombinedBaseListener implements RListener, RFilterListener {
```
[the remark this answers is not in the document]
robinmaisch: It is a dummy class full of empty methods, so that in "real" implementations, empty methods do not need to be reimplemented and so the code is more readable, see [above](#discussion_r901315237).

jplag.frontend.R/src/main/antlr4/de/jplag/R/grammar/RFilter.g4
robinmaisch: This code is taken from [here](https://github.com/antlr/grammars-v4/blob/master/r/RFilter.g4). I figure it is best to leave it as is to make it more clear that I used that version specifically, if that is okay. Valid point, though!

jplag.frontend.R/src/test/java/de/jplag/R/RFrontendTest.java
robinmaisch: You are right. I need to create any random `RToken` of the given type to be able to access the `type2string` method. That shows a weakness of the Token interface, I think. Refactoring TokenConstants to enums may solve that.

So, the -1's are literally just random; the `RToken` is only temporarily created for the sake of this error message. That's not very nice, but now I'm not sure what to do about it.

tsaglam: Just use a constant `NOT_SET` for now, if we move to enums instead of constants for the tokens, this can be implemented a bit nicer.

Also, I just noticed: Please use `line` or `lineIndex` instead of `lineIdx`.

jplag.frontend.R/src/main/java/de/jplag/R/RToken.java
tsaglam: Currently, we have no rules that we enforce. Historically, some classes have author tags and many don't. I use it for new classes or classes I rewrite from scratch. However, we cannot really enforce contributors to put their full names in the code, as many of them do not have their names on GitHub. So for now, I think it is fine to add the tags if you want to. I find them useful to see which person and thus in which context a class was written. I would not add an author tag to existing classes that I just edited.

jplag.frontend.R/src/main/antlr4/de/jplag/R/grammar/RFilter.g4
tsaglam: We actually have TODOs in our code base, usually, we use name abbreviations (e.g. `// TODO TS: we should do XYZ here`). Ideally, we should use issues instead of TODOs, this slowly getting rid of the TODOs. However, the grammar taken from other sources can be left as is, as we do not enforce ANTLR formatting anyway.

jplag.frontend.R/src/main/java/de/jplag/R/RCombinedBaseListener.java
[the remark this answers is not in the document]
tsaglam: Let's make it abstract, as it should not be instantiated.

jplag.frontend.R/src/main/java/de/jplag/R/RToken.java
robinmaisch: You are right, I don't remember why I put that there. Removed.

jplag.frontend.R/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.antlr</groupId>
+                <artifactId>antlr4-maven-plugin</artifactId>
+                <version>4.10.1</version>
```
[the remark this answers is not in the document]
robinmaisch: I copied this part from the POM of the csharp-6 module. In the aggregator POM, the version is set to `${project.version}`, and yet, if I remove the version tag, maven cannot resolve the dependency anymore. Can you help me with this`, @dfuchss?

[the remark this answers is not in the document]
robinmaisch: Oh, I misread the code excerpt. Done!

robinmaisch:
> To be discussed: We should maybe update the packages & artifact id to "rlang" or something similar.

Should the language option for the CLI also be adapted?

tsaglam:
> Should the language option for the CLI also be adapted?

Yes, I would say so.

tsaglam (approved):

jplag/src/main/java/de/jplag/options/LanguageOption.java
```diff
@@ -34,7 +35,7 @@ public String getDisplayName() {
     }
 
     public static LanguageOption fromDisplayName(String displayName) {
-        return Arrays.stream(LanguageOption.values()).filter(languageOption -> languageOption.displayName.equals(displayName)).findFirst()
+        return Arrays.stream(LanguageOption.values()).filter(languageOption -> languageOption.displayName.equalsIgnoreCase(displayName)).findFirst()
```
dfuchss, line 38: Why this change ?

jplag.frontend.rlang/src/main/antlr4/de/jplag/R/grammar/R.g4
```diff
@@ -0,0 +1,216 @@
...
+not end-of-command. See TestR.java
+Usage:
+$ antlr4 R.g4 RFilter.g4
+$ javac *.java
+$ java TestR sample.R
+... prints parse tree ...
+*/
+
+/*
+Modified version of the original in https://github.com/antlr/grammars-v4/blob/master/r/R.g4 so that I can separate the most relevant tokens of R in
```
dfuchss, line 41: Please use the exact link to the commit and not the branch here

jplag.frontend.rlang/src/main/antlr4/de/jplag/R/grammar/R.g4
```diff
@@ -0,0 +1,216 @@
...
+ INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT
+ NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+/**
+derived from http://svn.r-project.org/R/trunk/src/main/gram.y
```
dfuchss, line 28: Same here if possible .. see below

dfuchss (changes requested):
some minor comments

jplag/src/main/java/de/jplag/options/LanguageOption.java
tsaglam: This came up in a discussion above. While this could be a separate PR. I think we can include this here. Maybe we should mention it in the PR description though.

jplag.frontend.rlang/src/main/antlr4/de/jplag/R/grammar/R.g4
tsaglam: See above, probably leave it as is.

jplag/src/main/java/de/jplag/options/LanguageOption.java
tsaglam: I updated the description.

jplag.frontend.rlang/src/main/antlr4/de/jplag/R/grammar/R.g4
robinmaisch: @tsaglam Thanks for taking care of these!

robinmaisch:
As it turns out, the "renaming R to rlang" left some inconsistencies, which is just terrible. Can the R branch be merged again from this PR, or do I need to open a new one?

tsaglam:
> As it turns out, the "renaming R to rlang" left some inconsistencies, which is just terrible. Can the R branch be merged again from this PR, or do I need to open a new one?

just open a new PR from that branch with a short description of what went wrong and how you fixed it.

PR #463, opened 2022-06-23
Dynamic Loading of Language Frontends & Fixed behavior of JavaC

dfuchss:
This PR resolves two issues:

## Dynamic Loading of Language Frontends
In order to use new language frontends in JPlag easily, this PR creates introduces the ServiceLoader API to JPlag (#272 introduced some first idea. This PR realizes the mentioned alternative)

1. Languages (entry points of the language frontends) have to be annotated with `@MetaInfServices` and need to have a default constructor.
2. By doing so we use the ServiceLoader API of Java to load all Language Frontends (in the classpath) to JPlag
3. for the CLI, we use `getShortName()` as key
4. The dependencies between the core algorithm and the frontends has been relaxed

## Fixed behavior of JavaC 
This PR resolves Issue #467 by disabling the annotation processing of the Java Compiler. This workaround originates from the discussion on stackoverflow.


dfuchss:
Its the same problem again as for log4j .. errors will not be generated anymore by Javac

dfuchss:
@tsaglam resolved issue regarding loading of invalid Java Files (first occurred in #250)
See comments in https://github.com/jplag/JPlag/pull/463/commits/541a0e17637a2b5a6dfc50ef1e8e4ca2cfa1d33e and https://stackoverflow.com/questions/72737445/system-java-compiler-behaves-different-depending-on-dependencies-defined-in-mave#

dfuchss:
@jplag/maintainer the duplication is only present due to the nature of the language interface :)

dfuchss:
@jplag/maintainer in a followup PR we could also remove the direct deps to the front ends (only the CLI should point to them. Other libraries should include the front ends they need, nothing more)

tsaglam:
> @jplag/maintainer in a followup PR we could also remove the direct deps to the front ends (only the CLI should point to them. Other libraries should include the front ends they need, nothing more)

The core project needs a dependency on the Java frontend for the tests. But for the others, we should be able to do that.

dfuchss:
> > @jplag/maintainer in a followup PR we could also remove the direct deps to the front ends (only the CLI should point to them. Other libraries should include the front ends they need, nothing more)
> 
> The core project needs a dependency on the Java frontend for the tests. But for the others, we should be able to do that.

For the tests, yes`.. but for more?

tsaglam:
> For the tests, yes`.. but for more?

Only for tests afaik.

dfuchss:
@jplag/maintainer I've merged all changes of the main branch :)

So it's finally ready for review.

Should we rename any `Language` to something like `JavaLanguage` or `LanguagJava`?

tsaglam:
> Should we rename any `Language` to something like `JavaLanguage` or `LanguagJava`?

I would wait with that until the practicals are over and we rename/refactor the frontends.

jplag/src/test/java/de/jplag/special/TokenPrinterTest.java
```diff
@@ -7,30 +7,31 @@
...
+import de.jplag.*;
 import de.jplag.exceptions.ExitException;
 import de.jplag.options.JPlagOptions;
-import de.jplag.options.LanguageOption;
 
 /**
  * Special test case the does not really test anything but prints the tokens and the corresponding line of code.
  * @author Timur Saglam
  */
+// TODO DF: We could move some tests to their language frontends.
```
tsaglam, line 18: This is not a test per se, this is more like a debug utility tool. We could create an abstract token printer test in the test-utils and then add subclasses to each frontend. But I am not sure if that is so much better.

tsaglam (approved):
Just a comment.

jplag.cli/src/main/java/de/jplag/CommandLineArgument.java
```diff
@@ -34,7 +27,7 @@ public enum CommandLineArgument {
     ROOT_DIRECTORY(new Builder("rootDir", String.class).nargs(NumberOfArgumentValues.ZERO_OR_MORE_VALUES)),
     NEW_DIRECTORY(new Builder("-new", String.class).nargs(NumberOfArgumentValues.ONE_OR_MORE_VALUES)),
     OLD_DIRECTORY(new Builder("-old", String.class).nargs(NumberOfArgumentValues.ONE_OR_MORE_VALUES)),
-    LANGUAGE(new Builder("-l", String.class).defaultsTo(LanguageOption.getDefault().getDisplayName()).choices(LanguageOption.getAllDisplayNames())),
+    LANGUAGE(new Builder("-l", String.class).defaultsTo(de.jplag.java.Language.SHORT_NAME).choices(LanguageLoader.getAllLanguageNames())),
```
JanWittler, line 30: This `de.jplag.java.Language.SHORT_NAME` is kind of a magic String here. At least later in the `LanguageTest.java` we again use it to test the default language selection. Though there is no `LanguageOption` anymore, why not refactor this into a `getDefaultLanguageIdentifier` in the options or similar?

jplag.frontend.kotlin/src/main/java/de/jplag/kotlin/Language.java
```diff
@@ -2,15 +2,18 @@
...
 
 /**
  * This represents the Kotlin language as a language supported by JPlag.
  */
+@MetaInfServices(de.jplag.Language.class)
 public class Language implements de.jplag.Language {
 
     private static final String NAME = "Kotlin Parser";
-    private static final String SHORT_NAME = "Kotlin";
+    public static final String SHORT_NAME = "kotlin";
```
JanWittler, line 16: Maybe we rather rename this to `IDENTIFIER`?

JanWittler (changes requested):
Just some minor comments

jplag.frontend-utils/src/main/java/de/jplag/LanguageLoader.java
```diff
@@ -0,0 +1,75 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class contains methods to load {@link Language Languages}.
+ * @author Dominik Fuchss
+ */
+public final class LanguageLoader {
+    private static final Logger logger = LoggerFactory.getLogger(LanguageLoader.class);
+
+    private static List<Language> loaded = null;
```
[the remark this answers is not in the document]
JanWittler: Or maybe a Map with Language.identifier -> Language?

jplag.cli/src/main/java/de/jplag/CommandLineArgument.java
dfuchss: I think that's a highly specific config for the CLI (JPlagOptions is defined in the core of jplag) Therefore, I would not put a method for it in JPlag.

jplag.frontend.kotlin/src/main/java/de/jplag/kotlin/Language.java
dfuchss: I agree; we could do that. I simply used the SHORT_NAME everywhere. But IDENTIFIER would be better (we could also rename the `getShortName()`)

What do you think @tsaglam

jplag.cli/src/main/java/de/jplag/CommandLineArgument.java
JanWittler: And what about a static method in `CommandLineArgument`?

dfuchss: Yes that would be fine for me. Could also be a constant in CLA

JanWittler: Fine with both :)

dfuchss: EDIT: No, not possible since CLA is an Enum :(

dfuchss: @JanWittler I've added a constant DEFAULT_LANGUGE to CommandLineArgument

jplag.frontend.kotlin/src/main/java/de/jplag/kotlin/Language.java
tsaglam: Identifier sounds good. The getter has the benefit that devs of new frontends are forced to implement such a method due to the contract of the Language interface. However, we can also remove it if the constant is directly accessed anyways.

jplag.cli/src/main/java/de/jplag/CommandLineArgument.java
JanWittler: If we use the `identifier` proposal, maybe adjust the name then to DEFAULT_LANGUAGE_IDENTIFIER

dfuchss: Agree 👍

jplag.frontend-utils/src/main/java/de/jplag/Language.java
```diff
@@ -20,7 +20,7 @@ public interface Language {
     /**
      * Short name of the language used for CLI options.
      */
```
JanWittler, line 22: Adapt documentation

dfuchss: Thx. Missed that :)

jplag.cli/src/main/java/de/jplag/CommandLineArgument.java
```diff
@@ -34,7 +33,9 @@ public enum CommandLineArgument {
     ROOT_DIRECTORY(new Builder("rootDir", String.class).nargs(NumberOfArgumentValues.ZERO_OR_MORE_VALUES)),
     NEW_DIRECTORY(new Builder("-new", String.class).nargs(NumberOfArgumentValues.ONE_OR_MORE_VALUES)),
     OLD_DIRECTORY(new Builder("-old", String.class).nargs(NumberOfArgumentValues.ONE_OR_MORE_VALUES)),
-    LANGUAGE(new Builder("-l", String.class).defaultsTo(LanguageOption.getDefault().getDisplayName()).choices(LanguageOption.getAllDisplayNames())),
+    LANGUAGE(
+            new Builder("-l", String.class).defaultsTo(de.jplag.java.Language.IDENTIFIER)
```
JanWittler, line 37: Why not `DEFAULT_LANGUAGE_IDENTIFIER` here?

jplag.frontend-utils/src/main/java/de/jplag/LanguageLoader.java
```diff
@@ -0,0 +1,79 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class contains methods to load {@link Language Languages}.
+ * @author Dominik Fuchss
+ */
+public final class LanguageLoader {
+    private static final Logger logger = LoggerFactory.getLogger(LanguageLoader.class);
+
+    private static Map<String, Language> loaded = null;
```
JanWittler, line 21: Currently, the methods do not use the benefit of having the languages indexed by identifier.
I would propose to have a private method that loads all languages. This can either directly return the Map or guarantee that `loadedLanguages` (you did not rename it to my suggestion, any reason?) is populated after calling. Having the Map will simplify `getLanguage` (Map.get) and `getAllAvailableLanguageIdentifiers` (getKeySet). Furthermore, `getAllAvailableLanguages` does not have to duplicate the conversion from map to stream.

jplag.frontend-utils/src/main/java/de/jplag/LanguageLoader.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private LanguageLoader() {
+        throw new IllegalAccessError();
+    }
+
+    /**
+     * Load all languages that are currently in the classpath.
+     * @return the languages
+     */
+    public static synchronized List<Language> getAllAvailableLanguages() {
```
JanWittler, line 31: This could also be a `Set`. Sorting alphabetically is not done anymore here and has no semantic meaning to the language provider. Rather, it should be done directly in the CLI.

jplag.frontend-utils/src/main/java/de/jplag/LanguageLoader.java
```diff
@@ -0,0 +1,79 @@
...
+        if (result.isEmpty())
+            logger.warn("Attempt to load Language {} was not successful", identifier);
+        return result;
+    }
+
+    /**
+     * Get a list of all available languages with their identifiers.
+     * @return the list of all languages
+     */
+    public static List<String> getAllAvailableLanguageIdentifiers() {
```
JanWittler, line 69: This could also be a `Set`. Sorting alphabetically is not done anymore here and has no semantic meaning to the language provider.

jplag/src/main/java/de/jplag/JPlag.java
```diff
@@ -106,22 +104,20 @@ private ComparisonStrategy initializeComparisonStrategy(final ComparisonMode com
...
-            this.options.setLanguage(language);
-            this.options.setLanguageDefaults(language);
-            logger.info("Initialized language {}", language.getName());
-            return language;
-        } catch (NoSuchMethodException | SecurityException | ClassNotFoundException | InstantiationException | IllegalAccessException
-                | IllegalArgumentException | InvocationTargetException e) {
-            logger.error(e.getMessage(), e);
-            throw new IllegalStateException("Language instantiation failed:" + e.getMessage(), e);
+    private static Language initializeLanguage(JPlagOptions options) {
+        String languageShortName = options.getLanguageIdentifier();
```
JanWittler, line 108:
```suggestion
        String languageIdentifier = options.getLanguageIdentifier();
```

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -20,13 +20,17 @@
...
-
     public static final ComparisonMode DEFAULT_COMPARISON_MODE = NORMAL;
     public static final float DEFAULT_SIMILARITY_THRESHOLD = 0;
     public static final int DEFAULT_SHOWN_COMPARISONS = 30;
 
     public static final Charset CHARSET = StandardCharsets.UTF_8;
 
+    /**
+     * The short name of the language used to parse the submissions.
+     */
```
JanWittler, line 31: Adapt documentation

JanWittler (commented):
I found some more minor things. Sorry, didn't see them in the first iteration.

jplag.frontend-utils/src/main/java/de/jplag/LanguageLoader.java
dfuchss: Forgot to rename it :)

jplag.frontend-utils/src/main/java/de/jplag/LanguageLoader.java
dfuchss: Language would have to implement equals/hashcode. Therefore I use a list here :)

jplag.frontend-utils/src/main/java/de/jplag/LanguageLoader.java
dfuchss: Agree. Set is also possible.

jplag/src/main/java/de/jplag/options/JPlagOptions.java
dfuchss: mmh seems that my "search" did not find all interesting occurrences of short.

I'll fix that

PR #468, opened 2022-06-27
Specified versions & cleanup duplicated versions

dfuchss:
This PR cleans up the main pom.xml. Duplicated versions are merged to properties. Missing versions are added. A `deployment` profile has been added to distinguish between cli and library (for deployment). This is needed since JPlag uses the `nexus-staging-maven-plugin`


tsaglam:
> I don't understand this part of the maven-jar-plugin. Is this needed for every project?
> 
> ```
> <executions>
> <execution>
>     <goals>
>         <goal>test-jar</goal>
>     </goals>
> </execution>
> </executions>
> ```

It seems I added that when I introduced the frontend-testutils in #400. Maybe it is required for the dependency on that. Or it is not required and it is left over from my testing regarding putting both tests and main code into the frontend-utils, which did not really work.

dfuchss:
> > I don't understand this part of the maven-jar-plugin. Is this needed for every project?
> > ```
> > <executions>
> > <execution>
> >     <goals>
> >         <goal>test-jar</goal>
> >     </goals>
> > </execution>
> > </executions>
> > ```
> 
> It seems I added that when I introduced the frontend-testutils in #400. Maybe it is required for the dependency on that. Or it is not required and it is left over from my testing regarding putting both tests and main code into the frontend-utils, which did not really work. Maybe we remove it and see if anything breaks?

Yes, its needed for the test project.

dfuchss:
@tsaglam after review, this PR should be ready for merge :)

tsaglam (approved):

PR #469, opened 2022-06-27
Fixes to the TokenPrinter

robinmaisch:
I would like to suggest some changes to the `TokenPrinter` to make it even more useful for testing frontends.
Specifically, these changes are intended to fix the alignment of Token labels to the code lines.

- Account for tab characters in code (which is not the same as activating the REPLACE_TABS option)
- Account for line numbers that are longer than one tab character

Tested in a console with various settings of tab lengths and REPLACE_TABS to ensure that the correct alignment is independent from the actual console tab length _and_ the constant `TAB_LENGTH`.

Assuming that these changes leave no faults, any remaining misalignment would indicate that the position saved in the Token is wrong, which could then be fixed in the respective frontend.

Also, personally, I am confused by the lower case token labels for "tiny tokens". Can someone explain why that setting could be useful? Otherwise, I suggest setting `INDICATE_TINY_TOKENS` to `false` by default.

### DIFF
--- a/jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
+++ b/jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
@@ -31,7 +31,7 @@ public final class TokenPrinter {
     private static final String TAB_REPLACEMENT = "        "; // might depend on files
 
     // Configuration:
-    private static final boolean INDICATE_TINY_TOKEN = true; // print token with length <= 1 in lowercase
+    private static final boolean INDICATE_TINY_TOKEN = false; // print token with length <= 1 in lowercase
     private static final boolean REPLACE_TABS = false;
 
     private TokenPrinter() {
@@ -62,6 +62,7 @@ public final class TokenPrinter {
         int lineIndex = 0;
         int columnIndex = 1;
         String file = null; // no file yet
+        String currentLine = null;
 
         for (Token token : tokens.allTokens()) {
             // New code file:
@@ -80,13 +81,15 @@ public final class TokenPrinter {
                 columnIndex = 1;
 
                 String fileName = token.getFile().isEmpty() ? root.getName() : token.getFile();
-                String currentLine = filesToLines.get(fileName).get(lineIndex - 1);
+                currentLine = filesToLines.get(fileName).get(lineIndex - 1);
                 if (REPLACE_TABS) {
                     currentLine = currentLine.replace(TAB, TAB_REPLACEMENT);
                 }
                 appendCodeLinePrefix(builder, lineIndex).append(currentLine);
                 appendTokenLinePrefix(builder);
             }
+
+            assert currentLine != null;
 
             // New line if already past token index:
             if (columnIndex > token.getColumn()) {
@@ -95,7 +98,7 @@ public final class TokenPrinter {
             }
             // Move to token index:
             while (columnIndex < token.getColumn()) {
-                builder.append(SPACE);
+                builder.append(currentLine.length() > columnIndex && currentLine.charAt(columnIndex - 1) == '\t'? TAB : SPACE);
                 columnIndex++;
             }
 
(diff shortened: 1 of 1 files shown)


jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
```diff
@@ -119,12 +119,42 @@ public static String printTokens(TokenList tokens, Collection<File> files, File
...
+        if (REPLACE_TABS) {
+            currentLine = currentLine.replaceAll(TAB, TAB_REPLACEMENT);
+        }
+        return builder.append(currentLine);
+    }
+
+    /**
+     * Adds padding to the given StringBuilder in order to reach the targetPosition.
+     */
+    private static int addPadding(StringBuilder builder, int currentPosition, String currentLine, int targetPosition) {
```
dfuchss, line 132: Could you provide more information on the parameters :) Do you have information regarding the index of "currentPosition" are these indices within the currentLine? Indexed by 0 ? Is that clear to you? If so, please add it to the docs .. this will help future development

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
```diff
@@ -152,4 +182,16 @@ private static List<String> linesFromFile(File file) {
         }
         return Collections.emptyList();
     }
+
+    /**
+     * Returns the number of digits (including a minus) of the given number.
+     */
+    private static int digitCount(int index) {
```
dfuchss, line 189: Rename index to number or something like that :)

robinmaisch: True that^^

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
robinmaisch: I agree. I hope the new documentation is more helpful.

tsaglam:
> Also, personally, I am confused by the lower case token labels for "tiny tokens". Can someone explain why that setting could be useful? Otherwise, I suggest setting `INDICATE_TINY_TOKENS` to `false` by default. 

Some languages have tokens that have length <=1 even though they represent a source code element that is longer than one character. To show this via the token printer, this option can be used. However, we can set it to false per default, as it can always be enabled.

tsaglam:
> Assuming that these changes leave no faults, any remaining misalignment would indicate that the position saved in the Token is wrong, which could then be fixed in the respective frontend.

Have you tested it with different frontends? E.g. C++ and C#?

robinmaisch:
Now, another small fix ended up here which was not intended.
It makes it so that in verbose mode, the longMessages are logged by the ErrorCollector, instead of `null`, which is currently the case.

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
```diff
@@ -152,4 +195,16 @@ private static List<String> linesFromFile(File file) {
...
+    /**
+     * Returns the number of digits (including a minus) of the given number.
+     */
+    private static int digitCount(int number) {
+        if (number == 0) {
+            return 1;
+        }
+        int minusLength = number < 0 ? 1 : 0;
+        // The 'log10' variant is supposedly faster than the 'toString' variant.
+        return (int) Math.log10(Math.abs(number)) + minusLength + 1;
```
tsaglam, line 208: Careful with premature optimization! As this is a testing tool, performance probably does not matter that much.

tsaglam (approved):
Looks good!

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
robinmaisch: I agree! To be clear, I created this method myself, so I just tried to optimize my very own code, not code that existed before. I thought I would leave this comment to explain why I chose this approach.

PR #471, opened 2022-06-28
Fix to R frontend

robinmaisch:
As it turns out, the "renaming R to rlang" left the frontend inconsistent: The LanguageOption currently uses the updated package name, but the package/directory itself is still called R. So currently, the R frontend cannot be accessed from the CLI.

Now with this PR, there should be no element called 'R' left in the file tree, and no invalid references.
Also, I added a test for the rlang frontend to the TokenPrinterTest.

Funnily enough, this sort of thing did not get called out by the tests. Maybe the `LanguageOptionTest` should instantiate a respective `Language` object for each `LanguageOption`, just to be safe. 
Fortunately, if I am not mistaken, hard-coded package names in the actual code will not be necessary anymore once #463 is merged.


tsaglam (approved):
Looks good!

PR #488, opened 2022-07-02
Feature: Own Logger for CLI with Error Collection

dfuchss:
After a short discussion yesterday with @jplag/maintainer , I've created a simple draft implementation of a collective logger. (Based on slf4j-simple)

* It collects all errors and prints them at the end
* SLF4J-simple has been removed from the runtime classpath as slf4j does not like multiple implementations of a logger :)

Feel free to adopt any logging level :)


jplag.cli/src/main/java/de/jplag/logger/CollectedLogger.java
```diff
@@ -0,0 +1,270 @@
...
+import org.slf4j.helpers.FormattingTuple;
+import org.slf4j.helpers.MarkerIgnoringBase;
+import org.slf4j.helpers.MessageFormatter;
+import org.slf4j.spi.LocationAwareLogger;
+
+/**
+ * This logger is able to collect errors and print them at the end. Mainly adopted from org.slf4j.impl.SimpleLogger
+ * @author Dominik Fuchss
+ */
+public final class CollectedLogger extends MarkerIgnoringBase {
```
tsaglam, line 19: Quite a few abbreviations in identifiers in this class.

tsaglam (approved):
Just a few minor things.

jplag.cli/src/main/java/de/jplag/logger/CollectedLogger.java
dfuchss: Yes, but as I've written above. The code is mainly adopted from the SimpleLogger. Therefore, I didn't changed any naming (just added a few lines of code and removed some option-related code lines :) )

Esp. I'm not sure whether some names are important or not (I guess not, but I'm not sure about it :) )

JanWittler: I think we should still adapt it to our naming conventions as for future maintenance nobody will know this reasoning. Furthermore, applying some renaming refactorings is not the hardest thing to do ;)

dfuchss: Ok. I'll work on it asap

dfuchss: Done

PR #500, opened 2022-07-10
Add Golang frontend

robinmaisch:
This PR introduces a Golang frontend to JPlag. Its structure is very similar to the C# and R frontends.
It comes with a README, some basic tests, and some test code in written in Go.

Also, the README of the top module is updated to reflect the new CLI language options.
Its grammar definition should comply to go1.17, the current version being go1.18.


robinmaisch:
The remaining indicated code smells have a point, but they are kind of by design/convention based on the rest of the project. I would argue they are acceptable for now.

jplag.frontend.golang/src/main/java/de/jplag/golang/GoTokenConstants.java
```diff
@@ -0,0 +1,98 @@
...
+    int SLICE_BODY_END = 21;
+    int NAMED_TYPE_BODY_BEGIN = 22;
+    int NAMED_TYPE_BODY_END = 23;
+    int MEMBER_DECLARATION = 24;
+
+    // FUNCTIONS AND METHODS
+
+    int FUNCTION_DECLARATION = 25;
+    int METHOD_DECLARATION = 26;
+    int RECEIVER = 27;
```
JanWittler, line 37: This appears to me to open a relatively easy attack vector by simply refactoring a method to a function or vice versa. Why do you think it is worth having this separation?

jplag.frontend.golang/src/main/java/de/jplag/golang/GoTokenConstants.java
```diff
@@ -0,0 +1,98 @@
...
+    int SELECT_STATEMENT = 43;
+    int SELECT_BLOCK_BEGIN = 44;
+    int SELECT_BLOCK_END = 45;
+    int CASE_BLOCK_BEGIN = 46;
+    int CASE_BLOCK_END = 47;
+
+    // Statements
+
+    int VARIABLE_DECLARATION = 48;
+    int CONSTANT_DECLARATION = 49;
```
JanWittler, line 65: Same here. Why the separation between variable and constant?

JanWittler (commented):
I added some comments regarding the token selection which seemed worth a discussion for me. 
I am not an expert with Go, so there might be totally valid reasons to select the tokens as you did.

jplag.frontend.golang/src/main/java/de/jplag/golang/GoTokenConstants.java
robinmaisch: That's right, this transformation is probably very easy. I was too concerned about covering all kinds of sub-ASTs that I missed this. I need to keep that in mind.
The transformation of the invocations might still mess up the matching sequence, I suppose.

This is the relevant piece of the grammar. The receiver denotes the type on which the method shall be invocated, unlike other languages, where the method is associated with the surrounding type.
```
functionDecl: FUNC IDENTIFIER (signature block?);
methodDecl: FUNC receiver IDENTIFIER (signature block?);
```

JanWittler: > The transformation of the invocations might still mess up the matching sequence, I suppose.
How would the tokenised invocations look like? I would assume independent of function or method that an invocation translates to an `INVOCATION` token.

If we want to prevent such attack, we had to unify the function and method declaration tokens as well as the receiver and function parameter tokens. However, I am not sure whether this is too much up-front optimisation to prevent one specific obfuscation attack.

robinmaisch: - Regardless of the function/method discussion, the arguments in the invocation may be any kind of expression, so if a argument expression produces a token, the order of the parameter can already easily be changed to avoid matches.
```golang
myFunc(getSomeObject(), []int{1})
//would produce INVOCATION ARG INVOCATION ARG SLICE() SLICE{ SLICE_ELEM }SLICE

myFunc([]int{1}, getSomeObject())
//would produce INVOCATION ARG SLICE() SLICE{ SLICE_ELEM }SLICE ARG INVOCATION
```

- I am not sure, either. At this point, generic type parameters are not supported by the grammar, so they are ignored. If the grammar and frontend are updated, the type arguments would sit right in the middle, so this optimization still would not work in that case, although that case may arguably be rare.
```golang
func (this myStruct) foo<T>(t T) {  }
//would produce FUNC PARAM TYPE_PARAM PARAM FUNC{ ...

func foo<T>(this myStruct, t T) {  }
//would produce FUNC TYPE_PARAM PARAM PARAM FUNC{ ...
```

robinmaisch:
@jplag/studdev @jplag/maintainer
The conflicting files are merged; the current commit is now ready for review.
Maybe put a thumbs up so that I know that you got the message, otherwise I'll write an e-mail.

dfuchss:
The sonar cloud issue is known. I'm working on it in #538

dfuchss:
@robinmaisch can you merge the current develop to your branch?

robinmaisch:
> @robinmaisch can you merge the current master branch to your branch?

Done. It works, thanks!

jplag.frontend.golang/src/main/java/de/jplag/golang/GoParserAdapter.java
```diff
@@ -0,0 +1,68 @@
...
+            getErrorConsumer().addError("Parsing Error in '%s': %s%s".formatted(fileName, File.separator, exception));
+            return false;
+        }
+        return true;
+    }
+
+    public void addToken(int tokenType, int line, int column, int length) {
+        tokens.addToken(new GoToken(tokenType, currentFile, line, column, length));
+    }
+}
```
SuyDesignz, line 68: While reading through the class I noticed that in the rlang module the RParserAdapter implements the RTokenConstants. Was that an exception or is there a reason you did it differently here?

jplag.frontend.golang/src/main/java/de/jplag/golang/Language.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.io.File;
+
+import de.jplag.ErrorConsumer;
+import de.jplag.TokenList;
+
+public class Language implements de.jplag.Language {
+
+    private static final String SHORT_NAME = "Go";
+    private static final String NAME = "Go Parser";
```
SuyDesignz, line 11: Order would be better to put the long name before the short. Simply because of the readability but is not a must

jplag.frontend.golang/src/main/java/de/jplag/golang/grammar/GoParserBase.java
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * Returns true if the current Token is a closing bracket (")" or "}")
+     */
+    protected boolean closingBracket() {
+        BufferedTokenStream stream = (BufferedTokenStream) _input;
+        int prevTokenType = stream.LA(1);
+
+        return prevTokenType == GoParser.R_CURLY || prevTokenType == GoParser.R_PAREN;
+    }
```
SuyDesignz, line 25: purely for my understanding, why is it necessary to already include a flag in the ParserBase class if it is a closing bracket?

jplag.frontend.golang/src/main/java/de/jplag/golang/grammar/GoTokenUtils.java
```diff
@@ -0,0 +1,21 @@
+package de.jplag.golang.grammar;
+
+import de.jplag.golang.GoToken;
+
+public class GoTokenUtils {
+
+    private static final int NOT_SET = -1;
+    private static final String NOT_SET_STRING = "";
```
SuyDesignz, line 8: Rename to "EMPTY_STRING"

jplag.frontend.golang/src/main/java/de/jplag/golang/grammar/GoTokenUtils.java
```diff
@@ -0,0 +1,21 @@
...
+        return new GoToken(type, NOT_SET_STRING, NOT_SET, NOT_SET, NOT_SET);
+    }
+
+    public static GoToken getDummyToken(int type, String fileName) {
+        return new GoToken(type, fileName, NOT_SET, NOT_SET, NOT_SET);
+    }
+
+    private GoTokenUtils() {
+        // Utility class, should not be instantiated.
+    }
```
SuyDesignz, line 20: the constuctor should rather be at the beginning of the class functions even if it is empty or private

jplag.frontend.golang/src/test/java/de/jplag/golang/GoFrontendTest.java
```diff
@@ -0,0 +1,178 @@
...
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import de.jplag.Token;
+import de.jplag.TokenConstants;
+import de.jplag.TokenList;
+import de.jplag.TokenPrinter;
+import de.jplag.golang.grammar.GoTokenUtils;
+import de.jplag.testutils.TestErrorConsumer;
```
SuyDesignz, line 28: pull the imports apart as you have done so far in the other classes.

jplag/src/test/java/de/jplag/special/TokenPrinterTest.java
```diff
@@ -43,7 +43,7 @@ void printJavaFiles() {
         });
     }
 
-    @Disabled
+    @Disabled("Not a meaningful test, used for designing the token set")
```
SuyDesignz, line 46: good idea with the comment! simplifies the reading or increases the comprehensibility

pom.xml
```diff
@@ -154,19 +154,24 @@
...
             <dependency>
                 <groupId>de.jplag</groupId>
-                <artifactId>rlang</artifactId>
+                <artifactId>cpp</artifactId>
                 <version>${revision}</version>
             </dependency>
             <dependency>
                 <groupId>de.jplag</groupId>
-                <artifactId>cpp</artifactId>
+                <artifactId>golang</artifactId>
```
SuyDesignz, line 162: have you sorted the modules alphabetically?

jplag.frontend.golang/README.md
```diff
@@ -0,0 +1,29 @@
...
+More syntactic elements of Go may turn out to be helpful to include in the future, especially those that are newly introduced.
+
+### Usage
+
+To use the Go frontend, add the `-l golang` flag in the CLI, or use a `JPlagOption` object set to `LanguageOption.GO_LANG` in the Java API as described in the usage information in the [readme of the main project](https://github.com/jplag/JPlag#usage) and [in the wiki](https://github.com/jplag/JPlag/wiki/1.-How-to-Use-JPlag).
+
+<br>
+
+#### Footnotes
+<section id="footnote-1"><sup>1 </sup>The grammar files are taken from grammar-v4, with the most recent modifiactions in <a href="https://github.com/antlr/grammars-v4/tree/51ecccf87b75e96177287367b96cfa99e9f304b8/golang">commit 51ecccf</a> from April 2022.</section>
```
SuyDesignz, line 29: confirmed

jplag.frontend.golang/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+                    <execution>
+                        <goals>
+                            <goal>antlr4</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+</project>
```
SuyDesignz, line 45: confirmed

jplag.frontend.golang/src/main/antlr4/de/jplag/golang/grammar/GoLexer.g4
```diff
@@ -0,0 +1,521 @@
...
+// Treat whitespace as normal
+WS_NLSEMI                     : [ \t]+             -> channel(HIDDEN);
+// Ignore any comments that only span one line
+COMMENT_NLSEMI                : '/*' ~[\r\n]*? '*/'      -> channel(HIDDEN);
+LINE_COMMENT_NLSEMI : '//' ~[\r\n]*      -> channel(HIDDEN);
+// Emit an EOS token for any newlines, semicolon, multiline comments or the EOF and 
+//return to normal lexing
+EOS:              ([\r\n]+ | ';' | '/*' .*? '*/' | EOF)            -> mode(DEFAULT_MODE);
+// Did not find an EOS, so go back to normal lexing
+OTHER: -> mode(DEFAULT_MODE), channel(HIDDEN);
```
SuyDesignz, line 521: confirmed

jplag.frontend.golang/src/main/antlr4/de/jplag/golang/grammar/GoParser.g4
```diff
@@ -0,0 +1,380 @@
...
+sourceFile:
+	packageClause eos (importDecl eos)* (
+		(functionDecl | methodDecl | declaration) eos
+	)* EOF;
+
+packageClause: PACKAGE packageName = IDENTIFIER;
+
+importDecl:
+	IMPORT (importSpec | L_PAREN (importSpec eos)* R_PAREN);
+
```
SuyDesignz, line 45: in go you can provide the imports with names.
import m ("math") as an example. is this also tested for something like this? because if you specify the name as "." this is seen as a "test" import. (point 5 Dot import [Import in GoLang](https://www.geeksforgeeks.org/import-in-golang/) )
I don't know much about GO and the parsers so this is just for my understanding of the code.

SuyDesignz (changes requested):

pom.xml
tsaglam: Could be the pom sorting from spotless.

jplag.frontend.golang/src/main/java/de/jplag/golang/GoParserAdapter.java
robinmaisch: The R frontend was refactored from [here](https://github.com/CodeGra-de/jplag/blob/master/jplag.frontend.R/src/main/java/jplag/R/RToken.java), and they did it as a trick so that all the RTokenConstants are accessible without importing them. I left it in but did not include it in the all-new frontends that I did myself.

jplag.frontend.golang/src/main/java/de/jplag/golang/grammar/GoParserBase.java
robinmaisch: This file is part of the Base classes delivered in the ANTLR4 grammar repository, [here](https://github.com/antlr/grammars-v4/tree/51ecccf87b75e96177287367b96cfa99e9f304b8/golang/Java). To be frank, I have never tried to understand why this might be necessary.

tsaglam (approved):
Looks good to me. Resolve the issues mentioned in the other reviews, and this is ready to merge!

jplag.frontend.golang/src/main/antlr4/de/jplag/golang/grammar/GoParser.g4
robinmaisch: This file, too, was imported from the ANTLR4 grammar repository.
Just below the quote, line 46 handles import aliases and "dot imports":
```antlr
importSpec: alias = (DOT | IDENTIFIER)? importPath;
```
The grammar treats all three types as essentially the same thing, and the frontend has only one TokenType for all of them: `IMPORT`. I did not test all types, but I considered the grammar to be working correctly.

pom.xml
robinmaisch: Yes, I saw that the module list in the aggregator POM was now sorted, so I figured it would be neat if the dependencies would be sorted as well. Now it looks really stupid in the diff.

jplag/src/test/java/de/jplag/special/TokenPrinterTest.java
robinmaisch: Yes! SonarCloud required me to do that, but it is actually very helpful.

jplag.frontend.golang/src/test/java/de/jplag/golang/GoFrontendTest.java
robinmaisch: Can you explain what you mean? I let the IDE handle the imports.

jplag/src/test/java/de/jplag/special/TokenPrinterTest.java
tsaglam: @dfuchss enabling SonarCloud already paid off! 😄

jplag.frontend.golang/src/main/java/de/jplag/golang/GoParserAdapter.java
SuyDesignz: Ok perfect thanks!

dfuchss (changes requested):
Please merge the current main branch to your branch :) error consumer has been removed. A logger is available now :)

jplag.frontend.golang/src/main/java/de/jplag/golang/Language.java
```diff
@@ -0,0 +1,69 @@
+package de.jplag.golang;
+
+import java.io.File;
+
+import de.jplag.ErrorConsumer;
+import de.jplag.TokenList;
+
+public class Language implements de.jplag.Language {
```
JanWittler, line 8: After merging the master branch, this file should be cleaned up due to #552

jplag.frontend.golang/src/main/java/de/jplag/golang/JPlagGoListener.java
```diff
@@ -0,0 +1,523 @@
...
+        }
+
+        int getEnd() {
+            return this.endTokenType;
+        }
+
+        public int getElement() {
+            return this.elementTokenType;
+        }
+    }
```
JanWittler, line 522: Could you add some documentation to this please? What it does, why it is needed

JanWittler (changes requested):

jplag.frontend.golang/src/main/java/de/jplag/golang/JPlagGoListener.java
robinmaisch: Done

jplag.frontend.golang/src/main/java/de/jplag/golang/Language.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public boolean hasErrors() {
+        return parserAdapter.hasErrors();
+    }
+
+    @Override
+    public boolean supportsColumns() {
+        return true;
+    }
```
JanWittler, line 52: can be removed, as equal to default value

jplag.frontend.golang/src/main/java/de/jplag/golang/Language.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public boolean isPreformatted() {
+        return true;
+    }
```
JanWittler, line 57: same

jplag.frontend.golang/src/main/java/de/jplag/golang/Language.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public int numberOfTokens() {
+        return GoTokenConstants.NUM_DIFF_TOKENS;
+    }
+}
```
JanWittler, line 62: In the other languages this was adapted to be not included in the Language anymore and the tests directly accessing the constant. So we should be consistent here too.

robinmaisch: Done

jplag.frontend.golang/src/main/java/de/jplag/golang/GoTokenConstants.java
JanWittler: Following your argumentation that obfuscation in the parameters is possible anyways, I think we can close this discussion as it would probably lead to over engineering.

JanWittler (approved):

jplag.cli/pom.xml
```diff
@@ -12,10 +12,6 @@
             <groupId>de.jplag</groupId>
             <artifactId>jplag</artifactId>
         </dependency>
-        <dependency>
-            <groupId>de.jplag</groupId>
-            <artifactId>jplag</artifactId>
-        </dependency>
         <dependency>
             <groupId>de.jplag</groupId>
             <artifactId>frontend-utils</artifactId>
```
dfuchss, line 17: Add dependency to golang here

dfuchss: But that's something we can also do later

dfuchss (approved):

PR #487, opened 2022-07-01
Add Max Metric To Report Viewer

nestabentum:
This PR builds upon #466 Misc Improvements to Report viewer and should not be merged before it. (Only the latest commit 6c49658eeacc6f970b1a7705a652cdfd46f8ec3d is actually new and has to be reviewed)

Implementation of the second bullet point from Issue #357.

Backend: 

- Distribution and top comparisons are calculated from the JPlagResult and then serialized in the ReportDto. 

UI:

-  User can now switch between viewing the max or the average metric. With corresponding changes to the DistributionDiagram and ComparisonList. 
- The design of the MetricButton has been improved: The threshold information has been removed and an explanation has been added to the question mark icon as a hover tooltip
Screenshot 2022-07-01 at 16 50 52
Screenshot 2022-07-01 at 16 51 03

.

PR #507, opened 2022-07-14
Draft: Optimize report viewer dto size 

nestabentum:
Previous discussion about the new file format can be found [here](https://github.com/jplag/JPlag/issues/357#issuecomment-1173168697).

This PR builds upon my changes introduced in #487 which yet have to be merged. Therefore, in this PR only the changes introduced by commit efe795e are relevant.

I implemented an approach to reduce the report's file size: Code lines in a comparison file are not saved as plain text any more, but as numbers. These numbers are indices to a lookup table. 
This lookup table is global, i.e., it contains an index -> code line mapping, for every code line across all report files. This lookup table is persisted in an additional file lookupTable.json and will have to be used by the report viewer to resolve code lines before displaying them.

@sebinside noted that a global lookup table could prove unhandily large for large submission sets.
We will look into this concern and might change the format to one lookup table per submission.


sebinside:
@nestabentum So, I might have good news: I tested the lookup-table version with 51 real submissions with the `-n -1` flag, so that it generates *all* comparisons (about 1300 files in total). The older version generates about 23MB of data, your version only about 7.5MB. Also, the lookup table for 51 submissions was only about 2 MB, so that should be fine.

Also, you might already solved the bug with the lag in the UI anonymization, at least I did not encounter it this time. However, the problem in the generation still is there, while comparing took like 3 seconds, JPlag freezed for about 15 minutes while saving the files, so I was unable to test it with larger sets. However, I will test it again with a bigger programming task (50 submissions again, but larger ones) and will let you know about the result.

sebinside:
I repeated the test with larger submissions, this time the look-up table is around 3.5MB. So, regarding file sizes, this could work. The only thing that I'm currently worried about is that this table already has around 50,000 entries. I don't know if this can become a problem with larger submission sizes. Once the saving problem is solved (this took around half an hour of freezed JPlag), I can test it with around 500 submissions.

Regarding the viewer, I did not encounter any problems. The initial loading of the 12MB zip file took around 3 seconds, afterwards, everything was smooth. However, this might change when you add the resolving of indices to real code lines in the comparison view, so that's something to keep an eye on. Are there any other opinions? @tsaglam @dfuchss

dfuchss:
I'll have a look at it next week. 30 min are a very long time. We may need to discuss what takes this huge amount of time.

jplag/src/main/java/de/jplag/JPlagResult.java
```diff
@@ -93,34 +96,49 @@ public JPlagOptions getOptions() {
...
 
     /**
      * Note: Before, comparisons with a similarity below the given threshold were also included in the similarity matrix.
      */
     private int[] calculateSimilarityDistribution(List<JPlagComparison> comparisons) {
-        int[] similarityDistribution = new int[10];
-
-        comparisons.stream().map(JPlagComparison::similarity).map(percent -> percent / 10).map(Float::intValue).map(index -> index == 10 ? 9 : index)
-                .forEach(index -> similarityDistribution[index]++);
+        return calculateDistributionFor(comparisons, (JPlagComparison::similarity));
```
dfuchss, line 134:
```suggestion
        return calculateDistributionFor(comparisons, JPlagComparison::similarity);
```

jplag/src/main/java/de/jplag/JPlagResult.java
```diff
@@ -93,34 +96,49 @@ public JPlagOptions getOptions() {
...
+    }
 
+    private int[] calculateDistributionFor(List<JPlagComparison> comparisons, Function<JPlagComparison, Float> similarityExtractor) {
```
dfuchss, line 137: Please make this lambda to a 'normal' loop. From my point of view, it's not that easy to read

jplag/src/main/java/de/jplag/reporting/reportobject/mapper/ComparisonReportMapper.java
```diff
@@ -0,0 +1,107 @@
...
+        return new ComparisonReportMapperResult(comparisons, lineLookUpTable);
+    }
+
+    private List<FilesOfSubmission> getFilesForSubmission(Submission submission) {
+        return submission.getFiles().stream().map(file -> new FilesOfSubmission(file.getName(), readFileLines(file))).collect(Collectors.toList());
+    }
+
+    private List<Long> readFileLines(File file) {
+        List<Long> lineIndices = new ArrayList<>();
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
```
dfuchss, line 52: Maybe scanner is faster. But not for sure

jplag/src/main/java/de/jplag/reporting/reportobject/mapper/ComparisonReportMapper.java
```diff
@@ -0,0 +1,107 @@
...
+            String line;
+            while ((line = bufferedReader.readLine()) != null) {
```
dfuchss, line 54: Why should we not simply store the start/end of a match. Why the string ?

dfuchss: Afterwards, we can simply compress the submissions of the students, and maybe more if needed

jplag/src/main/java/de/jplag/reporting/reportobject/mapper/ComparisonReportMapper.java
```diff
@@ -0,0 +1,107 @@
...
+                }
+            }
+        } catch (IOException exception) {
+            logger.error("Could not read file: " + exception.getMessage());
+        }
+        return lineIndices;
+    }
+
+    private Long getIndexOfLine(String line) {
+        return lineLookUpTable.entrySet().stream().filter(entry -> Objects.equals(entry.getValue(), line)).map(Map.Entry::getKey).findFirst()
```
dfuchss, line 70: A reversed map may be a lot faster here.

jplag/src/main/java/de/jplag/reporting/reportobject/mapper/ComparisonReportMapper.java
```diff
@@ -0,0 +1,107 @@
...
+        Token startTokenSecond = tokensSecond.getToken(match.getStartOfSecond());
+        Token endTokenSecond = tokensSecond.getToken(match.getStartOfSecond() + match.getLength() - 1);
+
+        int startFirst = usesIndex ? startTokenFirst.getIndex() : startTokenFirst.getLine();
+        int endFirst = usesIndex ? endTokenFirst.getIndex() : endTokenFirst.getLine();
+        int startSecond = usesIndex ? startTokenSecond.getIndex() : startTokenSecond.getLine();
+        int endSecond = usesIndex ? endTokenSecond.getIndex() : endTokenSecond.getLine();
+        int tokens = match.getLength();
+
+        return new Match(startTokenFirst.getFile(), startTokenSecond.getFile(), startFirst, endFirst, startSecond, endSecond, tokens);
```
dfuchss, line 101: Simply store this information should be enough if I store the student submissions as well, or am I wrong ?

jplag/src/test/java/de/jplag/reporting/reportobject/mapper/ComparisonReportMapperTest.java
```diff
@@ -0,0 +1,107 @@
...
+            Map<String, List<String>> fileNameToFileLinesExpected = getAllFilesOfSubmission(submissionName, jPlagResult); // collect all files in a
+                                                                                                                          // map [fileName]->
+                                                                                                                          // <linesOfFile>
+
+            var unresolvedLinesFilesActual = actualFilesOfSubmission(submissionName, actualMapperResult); // get all files of submission from mapping
+                                                                                                          // result.
+            // their lines are unresolved, so they are only numbers until we looked them up with the lineLookUpTable
+
+            for (FilesOfSubmission fileOfSubmissionActual : unresolvedLinesFilesActual) { // per mapped submissionFile
+                var resolvedLinesOfFile = fileOfSubmissionActual.lines().stream().map(actualMapperResult.lineLookupTable()::get).toList(); // resolve
```
dfuchss, line 53: Formatter :)

jplag/src/test/java/de/jplag/reporting/reportobject/mapper/MetricMapperTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+public class MetricMapperTest {
+    private final MetricMapper metricMapper = new MetricMapper();
+
+    @Test
+    public void test_getAverageMetric() {
+        // given
+        JPlagResult jPlagResult = createJPlagResult(MockMetric.AVG, distribution(2, 3, 5, 7, 11, 13, 17, 19, 23, 29),
+                comparison(submission("1"), submission("2"), .7f), comparison(submission("3"), submission("4"), .3f));
+        // when
```
dfuchss, line 25: I don't get the comments here :)

jplag/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
dfuchss, line 1: This is for static mocks ?

dfuchss (changes requested):
I just had a look at the Java code :)

We can discuss possibilities for compression / serialization :)

sebinside:
> I'll have a look at it next week. 30 min are a very long time. We may need to discuss what takes this huge amount of time.

No no no, the 30 minutes are only there due to a bug - the real question discussed here is whether there should be one global lookup table per result or one lookup table per submission

dfuchss:
> > I'll have a look at it next week. 30 min are a very long time. We may need to discuss what takes this huge amount of time.
> 
> No no no, the 30 minutes are only there due to a bug - the real question discussed here is whether there should be one global lookup table per result or one lookup table per submission

Ah ok :)

dfuchss:
I would save the start/end info and not try to save lines of code in a map, or did I misunderstand anything?

How you save the code depends on the task type. Still, I think I would also save the source code per submission, then you don't have to load everything for comparison.

nestabentum:
@sebinside I pushed a (very) quick fix for the serialisation issues. can you have look at it and run it?

jplag/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
nestabentum: for final classes. I think it was necessary to mock an instance of a record

jplag/src/test/java/de/jplag/reporting/reportobject/mapper/MetricMapperTest.java
nestabentum: https://martinfowler.com/bliki/GivenWhenThen.html

jplag/src/main/java/de/jplag/reporting/reportobject/mapper/ComparisonReportMapper.java
nestabentum: I'm not really sure if I know what you mean.. Are you asking why we are reading the code from the submission files and writing it to the report at all? If so, I do think this is the way to go to enable to the UI to have as little complexity as possible.

Also, we do only store beginning and end of matches (see `ComparisonReportMapper` line 101)

nestabentum:
I am currently trying another file format that does not use a lookup table at all: 
[ImmediateSerialization.zip](https://github.com/jplag/JPlag/files/9178459/ImmediateSerialization.zip)


I realised that while it is convenient for the UI that the submissions are in the comparison model, (even with a lookup approach) it is redundant. So the format I'm currently trying does not have Submission files in the comparison model at all, but stores the submissions in `Submission/<SubmissionName>/`.
This again shifts complexity in to the UI but reduces the memory footprint even more.

sebinside:
@nestabentum That look very intersting, a good approach to the problem. I would argue that its not that bad to have some level of complexity in the UI as long as the peak work load does not become higher, which is probably the case here. 

Unfortunately I have some other deadlines so I'm not sure whether I will able to test out the new generation before our next meeting.

jplag/src/main/java/de/jplag/reporting/reportobject/mapper/ComparisonReportMapper.java
dfuchss: I want to include the submission code directly in the jsons.
I'm not sure why the lookup table is needed. 

I think:
a) store the source code to the jsons
b) save the start/end lines of a match

should contain the information we need. Why do we need the indirection (map)?

nestabentum: That’s exactly the format I‘ve switched to and I‘m currently implementing: https://github.com/jplag/JPlag/pull/507#issuecomment-1193628178

nestabentum:
Closed as there is now a cleaner version of this PR (#535 )

nestabentum:
Closed as there is now a cleaner version of this PR (#535)

PR #534, opened 2022-07-27
Core algorithm refactoring

JanWittler:
This PR aims to refactor the core algorithm to make it more readable and maintainable.
The following refactorings were performed:
- move state `Token.marked` into algorithm
- remove optimisations to reduce RAM usage in form of reversed hash tables only being generated when necessary
- remove `TokenHashMap` which only differed from a normal map by using a modulo on the keys which had no benefit for the algorithm
- encapsulate `Token.hash`, `TokenList.tokenHashes`, `TokenList.hashLength` into separate class `SubsequenceHashLookupTable` with improved naming and documentation
- remove `TokenList` and replace it with `List<Token>`
- move state `Token.isBaseCode` into algorithm
  - refactored base code tagging to occur outside of the core algorithm
  - only tag base code for compared submission, not for the base code submission
- refactors `JPlagComparison` to record

#### `TokenList` removal
With the removal of `TokenList` come two minor changes. 
First, the performed string optimisation to point to the same String object in case the file name for multiple tokens is the same was removed. I don't think this optimisation is necessary with modern-day computers anymore.
Second, a token list adjusted the tokens to have monotonically non-decreasing line numbers. This is only relevant for the visualisation, as the matching does not use the line number information. With some minor testing, the adjustment behaviour seemed to trigger only for FILE_END tokens which was fixed by assigning a NO_VALUE line to any FILE_END token. Since FILE_END token cannot appear in any match anyways, it should not make any difference for the visualisation of results.

#### Choice of actual `Set` / `Map` class
The refactorings introduce various variables of type `Set` or `Map`. To tune performance, we could test different types of these classes. An initial discussion was already started [here](https://github.com/jplag/JPlag/pull/534#discussion_r931576763). However, I would postpone this tuning until all changes (including follow-up PRs) are integrated.

#### Choice of subsequence hash constants

In `SubsequenceHashLookupTable`, we have the `MAX_HASH_LENGTH` and the `TOKEN_HASH_MODULO` parameters which are needed to restrict the computed hash to not exceed `Integer.MAX_VALUE`. Currently, the length is 25 and the modulo is 64. This means, that if a minimum match length of >25 is passed, the hashes are still only computed for 25, which may result in too many subsequence suggestions. The modulo implies that if a token type with a value of >64 is encountered, it is mapped to `value % 64`, which results in hash collision with another token type and thus may result in too many suggestions too. As the subsequences are only suggestions and verified afterwards, this is not a problem for functionality but for performance. Thus, we should re-evaluate whether a smaller `MAX_HASH_LENGTH` in favour of a larger `TOKEN_HASH_VALUE` is to be preferred (as we currently don't have anywhere a minimum token match of over 20). Exemplary, we could take the value pair `23 / 257`. We could also include multiple value pairs and select the appropriate one based on the passed `windowSize`. However, in local tests with the available example submissions, I could not identify a large benefit which is why I didn't change the values. We should test this with a large submission set. Also, this will be influenced by #623 as this PR will greatly decrease the times a token has a value larger than the MODULO. 

### Testing

As the test coverage of the core is not too high, we should validate the refactorings on additional data sets (and maybe try to integrate some of those in form of plain token sequences into the core). The refactorings should not change functionality, thus the similarity values should remain identical for any submission.
Furthermore, as the algorithm is highly optimised for performance, we should compare the performance of the refactored implementation to the previous one with a large data set.
Unfortunately, I don't have a comparison set at hand, so it would be beneficial if some of the @jplag/maintainer could help out here.

#### Tokens
I refactored all state related information of tokens into the algorithm. In a future PR, we should refactor tokens to be records instead of classes. However, for this to be possible, we need to avoid subclassing tokens for every language. This will probably not be needed anymore as soon as token constants are refactored to enums.

#### Naming
We use `firstSubmission` and `secondSubmission` throughout the application. From my perspective, these names indicate that there might be some kind of ordering which is not. Thus, I renamed the submissions in the core algorithm to `leftSubmission` and `rightSubmission`. Another option would have been `patternSubmission` and `textSubmission` which would be closer to the original paper. I chose the former naming as the clear separation between pattern (baseline file) and text (file to detect matches in) is not given in our application context. If left / right naming is accepted, in a following PR we should adjust the naming throughout the application.

#### Parallel Comparison Strategy
Since tokens are now state-free, it should be possible to entirely remove the locking the parallel comparison strategy. Will be done in #624.


JanWittler:
https://github.com/jplag/JPlag/blob/0780fc49c5d37c70f172f9d8289ddc158b6de934/jplag.frontend-utils/src/main/java/de/jplag/TokenList.java#L34-L45

I would like to remove the `TokenList` data structure entirely for which we have to get rid of this custom add functionality.
The first part (ll37-39) with the file name can probably be removed.
For the second part (ll40-42) I am not sure whether we require a token sequence to have monotonically non-decreasing lines. I would guess if anywhere then in the report viewer. @nestabentum do you by accident know more about that?

nestabentum:
@JanWittler all I know is that the `TokenList` is currently used to determine the matches of a comparison: https://github.com/jplag/JPlag/blob/0780fc49c5d37c70f172f9d8289ddc158b6de934/jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java#L156-L171

JanWittler:
> @JanWittler all I know is that the `TokenList` is currently used to determine the matches of a comparison:
> 
> https://github.com/jplag/JPlag/blob/0780fc49c5d37c70f172f9d8289ddc158b6de934/jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java#L156-L171
> 
> The information on matches is then further used to highlight them in the comparison view.

Alright, thanks for the pointer. I will have a look at that.

jplag/src/main/java/de/jplag/GreedyStringTiling.java
```diff
@@ -217,18 +177,23 @@ private void addMatchIfNotOverlapping(List<Match> matches, int startA, int start
...
-     * @param isBaseCodeComparison Whether the {@link Token#basecode} matches should be enabled for matching.
-     */
-    private void markTokens(TokenList tokenList, boolean isBaseCodeComparison) {
-        for (Token token : tokenList.allTokens()) {
-            if (isBaseCodeComparison) {
-                token.setMarked(token.type == FILE_END || token.type == SEPARATOR_TOKEN);
-            } else {
-                token.setMarked(token.type == FILE_END || token.type == SEPARATOR_TOKEN || (token.isBasecode() && options.hasBaseCode()));
+    private Set<Token> initiallyMarkedTokens(List<Token> tokens, boolean isBaseCodeComparison) {
+        Set<Token> markedTokens = new HashSet<Token>();
```
dfuchss, line 181: Consider an identity HashSet (otherwise we rely on equals and hashcode of the token, not sure whether we want this.

`Collections.newSetFromMap(new IdentityHashMap<>())` .. but not sure here

jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -154,12 +154,12 @@ private static List<FilesOfSubmission> getFilesForSubmission(Submission submissi
      * @return A Match DTO.
      */
     private static Match convertMatchToReportMatch(JPlagComparison comparison, de.jplag.Match match, Boolean usesIndex) {
-        TokenList tokensFirst = comparison.getFirstSubmission().getTokenList();
-        TokenList tokensSecond = comparison.getSecondSubmission().getTokenList();
-        Token startTokenFirst = tokensFirst.getToken(match.getStartOfFirst());
-        Token endTokenFirst = tokensFirst.getToken(match.getStartOfFirst() + match.getLength() - 1);
-        Token startTokenSecond = tokensSecond.getToken(match.getStartOfSecond());
-        Token endTokenSecond = tokensSecond.getToken(match.getStartOfSecond() + match.getLength() - 1);
+        List<Token> tokensFirst = comparison.getFirstSubmission().getTokenList();
```
dfuchss, line 157: maybe we want that getTokenList() returns immutable lists or copies .. just to make the impl clean.

dfuchss (commented):
Just some minor comments :)

jplag.frontend.cpp/src/main/java/de/jplag/cpp/Scanner.java
```diff
@@ -1,15 +1,16 @@
...
 
 import de.jplag.AbstractParser;
 import de.jplag.ErrorConsumer;
-import de.jplag.TokenList;
 
 public class Scanner extends AbstractParser {
     private String currentFile;
 
-    private TokenList tokens;
+    private List<de.jplag.Token> tokens;
```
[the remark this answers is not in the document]
JanWittler: Not possible due to naming conflicts. There is another `Token` class in the package (I think generated from the grammar).

jplag.frontend.python-3/src/main/java/de/jplag/python3/Parser.java
```diff
@@ -14,14 +16,13 @@
...
 import de.jplag.ErrorConsumer;
-import de.jplag.TokenList;
 import de.jplag.python3.grammar.Python3Lexer;
 import de.jplag.python3.grammar.Python3Parser;
 import de.jplag.python3.grammar.Python3Parser.File_inputContext;
 
 public class Parser extends AbstractParser {
 
-    private TokenList tokens = new TokenList();
+    private List<de.jplag.Token> tokens;
```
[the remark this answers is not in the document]
JanWittler: Not possible due to naming conflicts. There is another Token class in the package (I think generated from the grammar).

jplag.frontend.scheme/src/main/java/de/jplag/scheme/Parser.java
```diff
@@ -1,15 +1,16 @@
...
 
 import de.jplag.AbstractParser;
 import de.jplag.ErrorConsumer;
-import de.jplag.TokenList;
 
 public class Parser extends AbstractParser {
     private String currentFile;
 
-    private TokenList tokens;
+    private List<de.jplag.Token> tokens;
```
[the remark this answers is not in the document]
JanWittler: Not possible due to naming conflicts. There is another Token class in the package (I think generated from the grammar).

jplag.frontend.text/src/main/java/de/jplag/text/Parser.java
```diff
@@ -21,7 +22,7 @@ public class Parser extends AbstractParser {
     protected Hashtable<String, Integer> table = new Hashtable<>();
     protected int serial = 1; // 0 is FILE_END token
 
-    private TokenList tokens;
+    private List<de.jplag.Token> tokens;
```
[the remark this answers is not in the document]
JanWittler: Not possible due to naming conflict with `antlr.Token`

jplag/src/main/java/de/jplag/GreedyStringTiling.java
JanWittler: I am aiming to refactor `Token` to a record class thus `equals` and `hashCode` should get generated correctly.
However, it might still makes sense to use an identity HashSet as reference equality suffices for our use case. Do you know how the performance between HashSet and IdentityHashSet compares?

dfuchss: HashSet uses equals, IdentityHashMap (so also the set) uses ==

jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
JanWittler: Done.

JanWittler:
The aim of this PR is to make the core algorithm understandable by looking at the code, without reading the paper. From my perspective this is now given, thus the PR is ready for review. If you spot any parts of the algorithm that you still don't understand or would like to see more documentation for, please comment.

tsaglam:
> For the second part (ll40-42) I am not sure whether we require a token sequence to have monotonically non-decreasing lines. I would guess if anywhere then in the report viewer. @nestabentum do you by accident know more about that?

I think these lines ensure that EOF tokens have a proper line number. As EOF tokens are specified without a line index, this might be the here to set the line index to the one of the second to last token of the file. EOF tokens cannot easily given a line number in the frontend, as the line indexing is done by the parser/scanner/compiler. Also, this might be an optimization for tokens that do not use lines and columns but only a single index. However, this is less likely.

jplag/src/main/java/de/jplag/GreedyStringTiling.java
JanWittler: Will be overhauled in #623 anyways where we can't use identity anymore. Thus is it fine if we keep the `HashSet` here? We can later still do performance optimisation.

JanWittler:
@jplag/maintainer 
I think there is no other PR anymore blocking this one, so it is ready to merge from my side. The end2end tests all run successfully, so I would assume that the changes do not change any results (as intended). However, if you like to validate this on a larger submission set, feel free to do so.

jplag.frontend-testutils/src/test/java/de/jplag/testutils/TokenUtils.java
```diff
@@ -12,25 +11,23 @@ private TokenUtils() {
     }
 
     /**
-     * Returns the type of all tokens in a {@link TokenList} that belong to a file.
-     * @param tokenList is the {@link TokenList}.
+     * Returns the type of all tokens that belong to a certain file.
+     * @param tokenList is the list of {@link Token}s.
```
dfuchss, line 15:
```suggestion
     * @param tokenList is the list of {@link Token Tokens}.
```

jplag.frontend-testutils/src/test/java/de/jplag/testutils/TokenUtils.java
```diff
@@ -12,25 +11,23 @@ private TokenUtils() {
...
-        return tokens.stream().map(Token::getType).toList();
+    public static List<Integer> tokenTypesByFile(List<Token> tokens, String name) {
+        return tokensByFile(tokens, name).stream().map(Token::getType).toList();
     }
 
     /**
-     * Returns the tokens in a {@link TokenList} that belong to a file.
-     * @param tokenList is the {@link TokenList}.
+     * Returns the tokens that belong to a certain file.
+     * @param tokenList is the list of {@link Token}s.
```
dfuchss, line 25:
```suggestion
     * @param tokenList is the list of {@link Token Tokens}.
```

jplag.frontend-utils/src/main/java/de/jplag/Language.java
```diff
@@ -33,7 +34,7 @@ public interface Language {
      * @param files are the names of the files to parse.
      * @return the list of parsed JPlag tokens.
      */
-    TokenList parse(File directory, String[] files);
+    List<Token> parse(File directory, String[] files);
```
dfuchss, line 37: Maybe we should think about `List<? extends Token>` as this would make it clear that the list is only readable.

jplag.frontend.rust/src/test/java/de/jplag/rust/RustFrontendTest.java
```diff
@@ -75,7 +74,7 @@ private void testSourceCoverage(String fileName, TokenList tokens) {
             // All lines that contain code
             var codeLines = new ArrayList<>(getCodeLines(lines));
             // All lines that contain token
-            var tokenLines = IntStream.range(0, tokens.size()).mapToObj(tokens::getToken).mapToInt(Token::getLine).distinct().boxed().toList();
+            var tokenLines = tokens.stream().mapToInt(Token::getLine).filter(line -> line != Token.NO_VALUE).distinct().boxed().toList();
```
dfuchss, line 77: how about .stream().map().distinct().toList()

jplag.frontend.scala/src/test/java/de/jplag/scala/ScalaFrontendTest.java
```diff
@@ -101,7 +100,7 @@ private void testSourceCoverage(String fileName, TokenList tokens) {
             // All lines that contain code
             var codeLines = new ArrayList<>(getCodeLines(lines));
             // All lines that contain token
-            var tokenLines = IntStream.range(0, tokens.size()).mapToObj(tokens::getToken).mapToInt(Token::getLine).distinct().boxed().toList();
+            var tokenLines = tokens.stream().mapToInt(Token::getLine).filter(line -> line != Token.NO_VALUE).distinct().boxed().toList();
```
dfuchss, line 103: same as above

jplag.frontend.scala/src/test/java/de/jplag/scala/ScalaFrontendTest.java
```diff
@@ -157,11 +156,11 @@ private List<Integer> getCodeLines(List<String> lines) {
...
     /**
      * Confirms that all Token types are 'reachable' with a complete code example.
-     * @param tokens TokenList which is supposed to contain all types of tokens
+     * @param tokens list of tokens which is supposed to contain all types of tokens
      * @param fileName The file name of the complete code example
      */
-    private void testTokenCoverage(TokenList tokens, String fileName) {
-        var foundTokens = tokens.allTokens().stream().parallel().mapToInt(Token::getType).distinct().boxed().toList();
+    private void testTokenCoverage(List<Token> tokens, String fileName) {
+        var foundTokens = tokens.stream().parallel().mapToInt(Token::getType).distinct().boxed().toList();
```
dfuchss, line 163: same as above

jplag/src/main/java/de/jplag/GreedyStringTiling.java
```diff
@@ -4,100 +4,56 @@
...
  * @see <a href=
  * "https://www.researchgate.net/publication/262763983_String_Similarity_via_Greedy_String_Tiling_and_Running_Karp-Rabin_Matching">
  * String Similarity via Greedy String Tiling and Running Karp−Rabin Matching </a>
  */
 public class GreedyStringTiling {
 
-    private final JPlagOptions options;
+    private final int minimumMatchLength;
+    private Map<Submission, SubsequenceHashLookupTable> hashLookupTables = new HashMap<>();
+    private Map<Submission, Set<Token>> baseCodeMarkings = new HashMap<>();
```
dfuchss, line 28: We should think about IdentityHashMaps here.

jplag/src/main/java/de/jplag/JPlagComparison.java
```diff
@@ -1,103 +1,20 @@
...
-        return firstSubmission;
-    }
-
-    /**
-     * @return all matches between the two submissions.
-     */
-    public List<Match> getMatches() {
-        return matches;
-    }
+public record JPlagComparison(Submission firstSubmission, Submission secondSubmission, List<Match> matches) {
```
dfuchss, line 11: Maybe use `List<? extends Match>` to indicate read only list

dfuchss (changes requested):
Code looks good. Some remarks to discuss,

jplag.frontend-utils/src/main/java/de/jplag/Language.java
JanWittler: Resolved by improving the documentation

jplag/src/main/java/de/jplag/JPlagComparison.java
JanWittler: Resolved by improving the documentation

jplag.frontend.scala/src/test/java/de/jplag/scala/ScalaFrontendTest.java
dfuchss: I think this is missing :)

dfuchss (approved):

jplag.frontend.emf-metamodel/src/main/java/de/jplag/emf/util/MetamodelTreeView.java
```diff
@@ -34,23 +37,29 @@ public MetamodelTreeView(String filePath) {
...
     /**
-     * Adds a token to the view, thus adding the index information to the token.
+     * Adds a token to the view, thus adding the index information to the token. Returns a new token enriched with the index
+     * metadata.
      * @param token is the token to add.
      * @param treeDepth is the current containment tree depth, required for the indentation.
      */
-    public void addToken(MetamodelToken token, int treeDepth, String prefix) {
-        token.getEObject().ifPresent(it -> {
+    public MetamodelToken convertToMetadataEnrichedTokenAndAdd(MetamodelToken token, int treeDepth, String prefix) {
```
tsaglam, line 45: Probably the best solution for now, but not really a nice solution, as this method deals with two very different concerns, which makes it a but confusing for outsiders.

jplag.frontend.emf-metamodel/src/main/java/de/jplag/emf/util/MetamodelTreeView.java
```diff
@@ -34,23 +37,29 @@ public MetamodelTreeView(String filePath) {
...
+        int length = Token.NO_VALUE;
+        int line = Token.NO_VALUE;
+        int column = Token.NO_VALUE;
+        Optional<EObject> optionalEObject = token.getEObject();
+        if (optionalEObject.isPresent()) {
+            EObject eObject = optionalEObject.get();
```
tsaglam, line 51: Should be `optionalEObject.ifPresent(eObject -> {...})`

jplag/src/main/java/de/jplag/GreedyStringTiling.java
```diff
@@ -4,100 +4,56 @@
...
-            }
-        }
-        tokenList.hashLength = hashLength;
+        this.minimumMatchLength = options.minimumTokenMatch();
     }
 
     public final JPlagComparison compare(Submission firstSubmission, Submission secondSubmission) {
-        return swapAndCompare(firstSubmission, secondSubmission, false);
+        return swapAndCompare(firstSubmission, secondSubmission);
     }
```
tsaglam, line 35: Now this is no longer a convenience method and thus can be removed, right?

jplag/src/main/java/de/jplag/GreedyStringTiling.java
```diff
@@ -4,100 +4,56 @@
...
 
-    public final JPlagComparison compareWithBaseCode(Submission firstSubmission, Submission secondSubmission) {
-        return swapAndCompare(firstSubmission, secondSubmission, true);
+    public final JPlagComparison compareWithBaseCode(Submission submission, Submission baseCodeSubmission) {
```
tsaglam, line 37: This change is confusing me a bit. Previously, there were two methods. One for comparison without factoring in the basecode and one for comparison where basecode matches were ignored. The second one was required for the main comparison, whereas the first one was required to calculate matching sections with the basecode for each submission in firsthand. Does it now mean after the refactoring that `compareWithBaseCode()` is the first one and `compare()` is the the second one?

jplag/src/main/java/de/jplag/GreedyStringTiling.java
```diff
@@ -4,100 +4,56 @@
...
+        JPlagComparison comparison = swapAndCompare(submission, baseCodeSubmission);
+        // Remove the hashLookupTable as the isBaseCode tagging for the tokens changed which will affect the computed hashes.
```
tsaglam, line 39: I don't really understand the sentence...

jplag/src/main/java/de/jplag/GreedyStringTiling.java
```diff
@@ -4,100 +4,56 @@
...
 /**
  * This class implements the Greedy String Tiling algorithm as introduced by Michael Wise. However, it is very specific
- * to the classes {@link TokenList}, {@link Token}, and {@link Match}. While this class was reworked, it still contains
- * some quirks from the initial version.
+ * to the classes {@link Token}, and {@link Match}.
  * @see <a href=
  * "https://www.researchgate.net/publication/262763983_String_Similarity_via_Greedy_String_Tiling_and_Running_Karp-Rabin_Matching">
  * String Similarity via Greedy String Tiling and Running Karp−Rabin Matching </a>
  */
 public class GreedyStringTiling {
```
tsaglam, line 23: The changes here are hard to review, we should really verify the behavior is the same. We should also test this for a run with basecode.

jplag/src/main/java/de/jplag/JPlagComparison.java
```diff
@@ -1,103 +1,32 @@
...
-
-    /**
-     * @return all matches between the two submissions.
-     */
-    public List<Match> getMatches() {
-        return matches;
+    public JPlagComparison(Submission firstSubmission, Submission secondSubmission, List<Match> matches) {
+        this.firstSubmission = firstSubmission;
+        this.secondSubmission = secondSubmission;
+        this.matches = Collections.unmodifiableList(matches);
```
tsaglam, line 22: I do not like the use of an unmodifiable list here! Upon using this record, there is no indication that the list of matches is unmodifiable. If misused, a crash will only occur at runtime. This is a badly designed API, in my opinion.

jplag/src/main/java/de/jplag/JPlagComparison.java
```diff
@@ -139,44 +68,15 @@ public final float similarityOfFirst() {
...
-     * @return Similarity of the first submission to the basecode in percent rounded down to the nearest tenth.
-     */
-    public final float basecodeSimilarityOfFirst() {
-        return ((int) (firstBasecodeSimilarity() * ROUNDING_FACTOR)) / (float) ROUNDING_FACTOR;
-    }
-
-    /**
-     * @return Similarity of the second submission to the basecode in percent rounded down to the nearest tenth.
-     */
-    public final float basecodeSimilarityOfSecond() {
```
tsaglam, line 162: Whats the reason for the removal of these methods?

jplag/src/main/java/de/jplag/Submission.java
```diff
@@ -147,10 +149,10 @@ public int getSimilarityDivisor(boolean subtractBaseCode) {
     }
 
     /**
-     * @return Parse result of the submission.
+     * @return unmodifiable list of tokens generated by parsing the submission.
      */
-    public TokenList getTokenList() {
-        return tokenList;
+    public List<Token> getTokenList() {
+        return tokenList == null ? null : Collections.unmodifiableList(tokenList);
```
tsaglam, line 155: As already mentioned, I am not a fan of an unmodifiable list. I would prefer to return a read-only view.

jplag/src/main/java/de/jplag/Submission.java
```diff
@@ -147,10 +149,10 @@ public int getSimilarityDivisor(boolean subtractBaseCode) {
     }
```
tsaglam, line 149: This is not part of your PR; but getSimilarityDivisor() should be package-private. There is no reason for it to be part of the public API.

tsaglam (changes requested):
First review, no tests sanity checks done so far.

jplag.frontend-utils/src/main/java/de/jplag/Language.java
```diff
@@ -31,9 +32,9 @@ public interface Language {
      * Parses a set files in a directory.
      * @param directory is the directory where the files are located.
      * @param files are the names of the files to parse.
-     * @return the list of parsed JPlag tokens.
+     * @return the unmodifiable list of parsed JPlag tokens.
```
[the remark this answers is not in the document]
dfuchss: I mostly assume that I can only read from lists/collections from returns. Otherwise, it would get access to internals.

jplag/src/main/java/de/jplag/JPlagComparison.java
dfuchss: You could use List<? extends Match> but that does not guarantee that methods like remove(Object o) will be invoked.
Java has no real alternative. Always copying isn't the best option in my opinion and records should act as some kind of value type :)

jplag.frontend-utils/src/main/java/de/jplag/Language.java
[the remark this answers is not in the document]
JanWittler: All these things regarding unmodifiable were introduced due to [a previous review](https://github.com/jplag/JPlag/pull/534#discussion_r961579657). I don't have a strong opinion on that (except that I don't like that `List<? extends Token>` workaround) so we can also undo that.

jplag/src/main/java/de/jplag/GreedyStringTiling.java
JanWittler: I would rather say that the previous API was confusing 😄.
In the old state, the `isBaseCodeMarking` parameter was used to identify that a base code comparison is taking place and thus the token hashes are computed without considering the base code (which is only needed as the base code submission itself is tagged as base code and this has to be ignored) and then the matches are used to identify what was a base code. Otherwise, the hashes for subsequences containing a baseCode marked token were set to `-1` to avoid matches. So there was actually never the intention to "ignore" the base code markings but rather only bad separation of concerns which integrated marking of base code tokens into the core algorithm.

With the new approach, the base code submission tokens are never tagged as base code such that their subsequence hashes are always computed correctly, and the base code markings are created outside of the core algorithm.

Besides that, the `compare` and `compareWithBasecode` have the same semantics as the ones in the old version. But I will add documentation to the methods.

jplag/src/main/java/de/jplag/GreedyStringTiling.java
JanWittler:
```suggestion
        // Remove the lookup table for the current submission to trigger a regeneration as hashes will change due to the new baseCodeMarking.
```

Better like that? What I want to say is that we remove the computed hash lookup table for the submission we just computed the base code markings for, in order to trigger a regeneration of that table which considers the base code markings. As then said, this is done to increase performance, as subsequences with baseCode-marked tokens receive the `NO_VALUE` hash to ignore them during selection of possibly matching subsequences. Obviously, before computing the base code markings, the hash lookup table can't consider base code tokens and thus the hash lookup table computed during the comparison with the base code itself is not optimal.

jplag/src/main/java/de/jplag/JPlagComparison.java
JanWittler: So what is your suggestion? 😅

jplag/src/main/java/de/jplag/JPlagComparison.java
JanWittler: 1) They are not needed
2) The values regarding base code can be accessed using the `Submission#getBaseCodeComparison#similarity`

jplag.frontend-utils/src/main/java/de/jplag/Token.java
```diff
@@ -6,15 +6,14 @@
  * The language parsers decide what is a token and what is not.
  */
 public abstract class Token {
+    /** Indicates that the requested field has no value. */
+    public static final int NO_VALUE = -1;
```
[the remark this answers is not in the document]
JanWittler: Good point. I expected it to work as there are tests with a FILE_END token but it seems the tests are not correctly implemented. So currently it is not working, I will fix it.

jplag.frontend.emf-metamodel/src/main/java/de/jplag/emf/util/MetamodelTreeView.java
JanWittler: If I do that, I have to use `AtomicInteger` to be able to assign values in the lambda expression to the outside element which seemed counter-intuitive to me.

jplag.frontend-utils/src/main/java/de/jplag/Language.java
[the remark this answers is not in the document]
dfuchss: i guess we should simply use a shallow-copy then

jplag/src/main/java/de/jplag/GreedyStringTiling.java
tsaglam: Ah okay, now I get it. That makes sense! +1 for the documentation.

jplag/src/main/java/de/jplag/GreedyStringTiling.java
tsaglam: Ah, now I get it! Way better, thanks! 😄

jplag/src/main/java/de/jplag/JPlagComparison.java
tsaglam: Either we override the getter, or we turn it into a class again. Either way, I am in favor of avoiding `Collections.unmodifiableList` as much as possible, especially if it concerns the public API. However, I am heavily biased here by my pet peeve regarding the Java devs disregard for Liskov! We can also discuss our policy regarding that internally today.

jplag/src/main/java/de/jplag/JPlagComparison.java
tsaglam: Sounds good to me. If the Artemis crowd complains (as these methods are part of the public interface), we can always adjust.

jplag.frontend-utils/src/main/java/de/jplag/Token.java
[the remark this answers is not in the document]
JanWittler: Fixed in https://github.com/jplag/JPlag/pull/534/commits/01e3c582da8918f0f8b72d7d977c8d214affac0f.

I was not able to make it work in such a nice functional way as before without introducing an additional data structure (for the tuple `(line, token)`) which I found a bit over engineered. If there is one, feel free to suggest it 😄 .

jplag.frontend-utils/src/main/java/de/jplag/Language.java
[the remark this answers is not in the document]
JanWittler: As this is an interface which can be implemented by anyone, we cannot control what is returned and thus cannot force a shallow copy here. So for this case I would simply suggest to remove the `unmodifiable` word in the documentation again. Fine like that?

jplag.frontend.emf-metamodel/src/main/java/de/jplag/emf/util/MetamodelTreeView.java
JanWittler: True, but this "two concerns" behaviour was already present before the refactoring. In the old state, the token was modified directly as a side effect instead of creating a new instance. So I would argue that the changes don't make things worse than they already were 😅 .

jplag/src/main/java/de/jplag/JPlagComparison.java
JanWittler: Agreed in maintainer meeting to use unmodifiable lists with an according documentation. Thus this resolves without further action.

jplag/src/main/java/de/jplag/Submission.java
JanWittler: Agreed in maintainer meeting to use unmodifiable lists with an according documentation. Thus this resolves without further action.

dfuchss (approved):

jplag.frontend-utils/src/main/java/de/jplag/Language.java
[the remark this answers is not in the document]
JanWittler: Documentation adapted in https://github.com/jplag/JPlag/pull/534/commits/566b736026a745fdd7473779f36bac8382450191

PR #538, opened 2022-07-29
Create a workaround for sonarcloud on forks

dfuchss:
Since external (forks) cannot use sonarcloud so far, I try workarounds from https://community.sonarsource.com/t/how-to-use-sonarcloud-with-a-forked-repository-on-github/7363/30


dfuchss:
Also interesting: https://github.com/matrix-org/sonarcloud-workflow-action

PR #552, opened 2022-08-01
Overhaul language interface

tsaglam:
The goal of this PR is to clean up the language interface:

- Remove the `usesIndex()` method as it just returns `!supportsColumns()`
- Introduce default methods for `supportsColumns()` and `isPreformatted()`
- Makes the JDoc comments of the language interface a bit more precise
- Adapts all existing frontends accordingly


robinmaisch (approved):
Looks good to me!

jplag.frontend-utils/src/main/java/de/jplag/Language.java
```diff
@@ -41,22 +41,23 @@ public interface Language {
...
 
     /**
-     * Number of defined tokens in the scanner of the language.
+     * Determines whether a fixed-width font should be used to display that language.
      */
-    int numberOfTokens();
+    default boolean isPreformatted() {
+        return true;
+    }
 }
```
JanWittler, line 63: This field is never used, maybe we just drop it too?

JanWittler: No, the `isPreformatted`

tsaglam: This should be used by the report viewer in future versions. Currently, the viewer always uses a fixed-width font.

JanWittler (approved):

PR #514, opened 2022-07-16
Cleanup some minor code issues

dfuchss:
I had some time for searching for code issues again. A code review is needed for these changes :)

* Made classes to records if possible
* Resolved issues found by SCA

PR #553, opened 2022-08-01
Introduce a cleaner way to use suppress warnings

dfuchss:
As mentioned by @JanWittler in #514 , we should document our suppress warning better :)

Here's a simple proposal for that


dfuchss:
Ok Sonarlint ignores it. Locally it works :(

dfuchss:
I think we have to investigate more here.

tsaglam:
As an alternative, we can just merge #555.

JanWittler:
> As an alternative, we can just merge #555.

If that solves all suppressed warnings, I would definitely go with that solution and thus close this PR. (But thanks for the effort to solve my complaints :D )

PR #555, opened 2022-08-02
Overhaul text frontend to improve code quality and reduce complexity

tsaglam:
This PR cleans up the text frontend, thus resolving some Sonar issues. It improves code quality and detangles some odd design choices. In detail:

- `Parser` is now called `ParserAdapter` (because it is an an adapter to the antlr parser)
- `ParserToken` is now called `AntlrParserToken` (because these are only the token provided by antlr)
- Fix visibilities by proper encapsulation
- Detangle "serial"/token type creation by moving it from the token class into the parser adapter
- Improve JDoc
- Minor code style fixes & formatting
- Add a very minimal test case (which also passes with the unchanged frontend)

This PR is an alternative to #553.

### DIFF
--- a/jplag.frontend.text/src/main/java/de/jplag/text/InputState.java
+++ b/jplag.frontend.text/src/main/java/de/jplag/text/InputState.java
@@ -10,31 +10,48 @@ import antlr.LexerSharedInputState;
  * This object contains the data associated with an input stream of characters. Multiple lexers share a single
  * LexerSharedInputState to lex the same input stream.
  */
-@SuppressWarnings({"java:S1104", "java:S2387"})
 public class InputState extends LexerSharedInputState {
-    public int column = 0;
-    public int tokenColumn = 0;
+    private int columnIndex;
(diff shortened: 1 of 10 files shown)


jplag.frontend.text/src/main/java/de/jplag/text/InputState.java
```diff
@@ -10,31 +10,48 @@
  * This object contains the data associated with an input stream of characters. Multiple lexers share a single
  * LexerSharedInputState to lex the same input stream.
  */
-@SuppressWarnings({"java:S1104", "java:S2387"})
 public class InputState extends LexerSharedInputState {
-    public int column = 0;
-    public int tokenColumn = 0;
+    private int columnIndex;
+    private int tokenColumnIndex;
```
JanWittler, line 15: default value to 1 and then remove the `initialize` method?

JanWittler (approved):
Looks fine, thanks for the effort. One minor comment

PR #566, opened 2022-08-03
Fix warnings of Sonarcloud

JanWittler:
Fixes the warnings of Sonarcloud such that the build is not marked as _failing_ anymore.
- Fixes potential NullpointerException in `SubmissionSetBuilder`
- Validate input in `BayesianOptimization` such that later on NullpointerException is not possible
- Implement `hashCode` and `equals` for `Submission` and `JPlagComparison`
  - Remove conformance to `Comparable` for `JPlagComparison` as it was not consistent with `equals` (thus resolving #233)

Additionally fixed some warnings for the generated `CPPScanner`.

### DIFF
--- a/jplag/src/main/java/de/jplag/Submission.java
+++ b/jplag/src/main/java/de/jplag/Submission.java
@@ -83,10 +83,11 @@ public class Submission implements Comparable<Submission> {
     public boolean equals(Object obj) {
         if (obj == this) {
             return true;
-        } else if (!(obj instanceof Submission)) {
+        }
+        if (!(obj instanceof Submission otherSubmission)) {
             return false;
         }
-        return ((Submission) obj).getName().equals(name);
+        return otherSubmission.getName().equals(name);
     }
 
     @Override
--- a/jplag/src/main/java/de/jplag/Submission.java
+++ b/jplag/src/main/java/de/jplag/Submission.java
@@ -5,6 +5,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collection;
+import java.util.Objects;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


jplag/src/main/java/de/jplag/Submission.java
```diff
@@ -75,10 +76,24 @@ public Submission(String name, File submissionRootFile, boolean isNew, Collectio
...
     }
 
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        } else if (!(obj instanceof Submission)) {
+            return false;
+        }
+        return ((Submission) obj).getName().equals(name);
```
dfuchss, line 89:
```suggestion
        if (obj == this) {
            return true;
        } 
        if (!(obj instanceof Submission otherSubmission)) {
            return false;
        }
        return otherSubmission.getName().equals(name);
```

jplag/src/main/java/de/jplag/SubmissionSetBuilder.java
```diff
@@ -158,7 +158,7 @@ private Optional<Submission> loadBaseCode(Set<File> submissionDirectories, Set<F
                 logger.info("Submission \"{}\" is the specified basecode, it will be skipped during comparison.", removed.getName());
             }
         }
-        return baseCodeSubmission;
+        return Optional.of(baseCode);
```
dfuchss, line 161: ofNullable here?

dfuchss (commented):
Sounds good. Some minor suggestion

jplag/src/main/java/de/jplag/SubmissionSetBuilder.java
JanWittler: Thanks for spotting that. Had some other refactoring before where `ofNullable` was not necessary anymore.

tsaglam:
Note that this PR does not really fix #233, as the `hashCode()` method is not consistent with the `equals()` method.

JanWittler:
> Note that this PR does not really fix #233, as the `hashCode()` method is not consistent with the `equals()` method.

Regarding which class? I intended to implement them consistent.

tsaglam:
> Regarding which class? I intended to implement them consistent.

Ah, you are right. You just hashing the name, not the whole class. Alright, my mistake.

jplag/src/main/java/de/jplag/Submission.java
dfuchss: Yes :)

dfuchss (approved):
Code lgtm

Issue #354, opened 2022-04-07
Crash writing json files while using multiple root directories


Alberth289346:
The new reporting backend breaks on trying to write to non-existing directories.
``` sh
# using 2 directories, each containing 1 `merged.cif` file.
$ ls -1 1DH/merged/2018-101/* 1DH/merged/2018-102/*
1DH/merged/2018-101/merged.cif
1DH/merged/2018-102/merged.cif

# Delete all files in result.
$ rm result/*

# 38173b1 without all frontends + our own cif language frontend
$ java -jar cifjplag.jar -bc 1DH/merged/_base 1DH/merged/2018-101 1DH/merged/2018-102 
Initialized language CIF
JPlag initialized
Basecode directory "_base" will be used.
2 submissions parsed successfully!
0 parser errors!
0 too short submissions!

----- Parsing basecode submission: _base
Basecode submission parsed!
Comparing 2018-101/merged.cif-2018-102/merged.cif: 34.77527

Total time for comparing submissions: 0 sec
Failed to save json files: result/2018-101/merged.cif-2018-102/merged.cif.json (No such file or directory)

# ls results
overview.json
```
Backtracking to cause:

``` java
// JsonFactory.java
// line 47
            for (ComparisonReport report : jPlagReport.getComparisons()) {
                String name = report.getFirstSubmissionId().concat("-").concat(report.getSecondSubmissionId()).concat(".json");
                mapper.writeValue(Path.of(folderPath, name).toFile(), report);
            }
        } catch (IOException e) {
            System.out.println("Failed to save json files: " + e.getMessage());
            return false;
        }
```

``` java
// ComparisonReport.java
// line 9
    @JsonProperty("first_submission_id")
    private final String firstSubmissionId;

    @JsonProperty("second_submission_id")
    private final String secondSubmissionId;

// line 27
    public ComparisonReport(String firstSubmissionId, String secondSubmissionId, float matchPercentage,
            List<FilesOfSubmission> filesOfFirstSubmission, List<FilesOfSubmission> filesOfSecondSubmission, List<Match> matches) {

        this.firstSubmissionId = firstSubmissionId;
        this.secondSubmissionId = secondSubmissionId;
```

``` java
// ReportObjectFactory.java
// line 64
                new ComparisonReport(comparison.getFirstSubmission().getName(), //
                        comparison.getSecondSubmission().getName(), //
                        comparison.similarity(), //
                        getFilesForSubmission(comparison.getFirstSubmission()), //
[...]

PR #569, opened 2022-08-05
Enable Multiroot for Report Viewer

nestabentum:
This is still WIP but I wanted to share my approach already to receive feedback

This PR intends to fix two major issues (both noted in issue #354):

1. Submission names were concatenated to produce the comparison file name. When JPlag is run with multiple roots, the submission names contain backward slashes causing file creation to crash.
2. As Submission names were concatenated  duplicate names could easily be produced in the single source case ({`A, A-B, C, B-C}` would result in two `A-B-C` )

My approach to fixing this:

 * Differentiation between the submission name, submission id, and submission display name.
      * `Submission name` simply refers to Submission#name
      * `Submission id` is newly introduced and used only in the context of the report viewer creation. Generation of this id is the topic of another (our next 😄 ) discussion, but currently, it is the sanitized version of the submission name (File separator replaced by underscore). This submission id serves as an id for internal handling in the report viewer. 
      * `Display name` is the name of a submission to be displayed by the report viewer. The `overview.json` file contains a map <`submission id` -> `display name`> the report viewer will use to resolve the name for presentation purposes.
 * Two be able to abolish name concatenation of submission names for the comparison file name, I introduced a nested lookup table to the `overview.json`. It has the structure <`submission id `-> <<`submission id`> -><`comparison file name`>>>. With this data structure, there will be no duplicate comparison file names and additionally, the comparison file name can be chosen arbitrarily as the frontend will only have to execute two gets two receive the file name and will not have to rely on string concatenation.

Currently, the nested lookup table contains all submission ids in its first level, and in the second level, it contains each comparison for the corresponding submission id. This redundancy could be eliminated and actually halve the table's size but this is an optimization for later.
The UI isn't fully compatible with this approach yet, but will be with my next commits.

EDIT: 
The report viewer is now fully compatible with the updated file format.
I realised that the problem in 2. persisted, as I still used id concatenation. Therefore, I changed the comparison file name generation to simply be a number that is counted up for each file. While this worsens human readability, it solves the aforementioned issue. As always, suggestions on how to solve this issue while maintaining readability are welcome.


jplag/src/main/java/de/jplag/reporting/jsonfactory/DirectoryCreator.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Creates a directory.
+     * @param path The path under which the new directory ought to be created
+     * @param name The name of the new directory
+     * @return The created directory
+     */
+    public static File createDirectory(String path, String name) {
+        File directory = new File(path.concat("/").concat(name));
+        if (!directory.exists() && !directory.mkdirs()) {
+            logger.error("Failed to create dir.");
```
dfuchss, line 20: If recovery is not possible. You should use exceptions

jplag/src/test/java/de/jplag/reporting/reportobject/mapper/ComparisonReportWriterTest.java
```diff
@@ -0,0 +1,31 @@
...
+import de.jplag.JPlagResult;
+import de.jplag.Submission;
+import de.jplag.TestBase;
+import de.jplag.exceptions.ExitException;
+import de.jplag.reporting.jsonfactory.ComparisonReportWriter;
+
+public class ComparisonReportWriterTest extends TestBase {
+
+    @Test
+    public void test() throws ExitException {
```
dfuchss, line 17: Maybe use a better name for the test

dfuchss (commented):
Very minor comments :)

sebinside (approved):
I only had a quick look over the code, so there is a more in-depth review from @jplag/studdev required. But it tested the PR with two directories with 200 and 50 submissions, and everything worked fine.

Probably not related to this PR, but sometimes, the comparison viewer still breaks, but not predictable with the warning `Could not find comparison file. indes.js:43` and then with the error:
```
templateLoader.js?e847:50 
        
       Uncaught (in promise) TypeError: _ctx.isAnonymous is not a function
    at Proxy.render (templateLoader.js?e847:50:1)
    at renderComponentRoot (runtime-core.esm-bundler.js?d2dd:896:1)
    at ReactiveEffect.componentUpdateFn [as fn] (runtime-core.esm-bundler.js?d2dd:5580:1)
    at ReactiveEffect.run (reactivity.esm-bundler.js?89dc:185:1)
    at instance.update (runtime-core.esm-bundler.js?d2dd:5694:1)
    at setupRenderEffect (runtime-core.esm-bundler.js?d2dd:5708:1)
    at mountComponent (runtime-core.esm-bundler.js?d2dd:5490:1)
    at processComponent (runtime-core.esm-bundler.js?d2dd:5448:1)
    at patch (runtime-core.esm-bundler.js?d2dd:5038:1)
    at ReactiveEffect.componentUpdateFn [as fn] (runtime-core.esm-bundler.js?d2dd:5660:1)
```

jplag/src/main/java/de/jplag/reporting/jsonfactory/DirectoryCreator.java
```diff
@@ -0,0 +1,33 @@
+package de.jplag.reporting.jsonfactory;
+
+import java.io.File;
+import java.io.IOException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DirectoryCreator {
```
tsaglam, line 9: JDoc

jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -19,85 +19,64 @@
...
+        submissionToIdFunction = (Submission submission) -> submissionNameToIdMap.get(submission.getName());
     }
 
-    private static void copySubmissionFilesToReport(String path, JPlagResult result) {
+    private void copySubmissionFilesToReport(String path, JPlagResult result) {
         List<JPlagComparison> comparisons = result.getComparisons(result.getOptions().getMaximumNumberOfComparisons());
         var submissions = getSubmissions(comparisons);
-        var submissionsPath = createDirectory(path, SUBMISSIONS_FOLDER);
+        var submissionsPath = createSubmissionsDirectory(path);
+        if (submissionsPath == null)
```
tsaglam, line 73: Avoid if statements without curly braces.

jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -109,52 +88,78 @@ private static void copySubmissionFilesToReport(String path, JPlagResult result)
...
-        return submissions;
+    private File createSubmissionDirectory(String path, File submissionsPath, Submission submission) {
+        File directory;
+        try {
+            directory = createDirectory(submissionsPath.getPath(), submissionToIdFunction.apply(submission));
+        } catch (IOException e) {
+            logger.error("Could not create directory " + path + " for report viewer generation", e);
+            return null;
+        }
+        return directory;
```
tsaglam, line 99:
```suggestion
        try {
            return createDirectory(submissionsPath.getPath(), submissionToIdFunction.apply(submission));
        } catch (IOException e) {
            logger.error("Could not create directory " + path + " for report viewer generation", e);
            return null;
        }
```

jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -109,52 +88,78 @@ private static void copySubmissionFilesToReport(String path, JPlagResult result)
...
-        return result.getComparisons(numberOfComparisons);
+    private File createSubmissionsDirectory(String path) {
+        File submissionsPath;
+        try {
+            submissionsPath = createDirectory(path, SUBMISSIONS_FOLDER);
+        } catch (IOException e) {
+            logger.error("Could not create directory " + path + " for report viewer generation", e);
+            return null;
+        }
+        return submissionsPath;
```
tsaglam, line 110: As above.

jplag/src/main/java/de/jplag/reporting/reportobject/mapper/SubmissionNameToIdMapper.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.File;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import de.jplag.JPlagComparison;
+import de.jplag.JPlagResult;
+import de.jplag.Submission;
+
+public class SubmissionNameToIdMapper {
```
tsaglam, line 12: JDoc

jplag/src/main/java/de/jplag/reporting/reportobject/mapper/SubmissionNameToIdMapper.java
```diff
@@ -0,0 +1,35 @@
...
+        HashMap<String, String> idToName = new HashMap<>();
+        getComparisons(result).forEach(comparison -> {
+            idToName.put(comparison.getFirstSubmission().getName(), sanitizeNameOf(comparison.getFirstSubmission()));
+            idToName.put(comparison.getSecondSubmission().getName(), sanitizeNameOf(comparison.getSecondSubmission()));
+        });
+        return idToName;
+    }
+
+    private static String sanitizeNameOf(Submission comparison) {
+        return comparison.getName().replace(File.separator, "_");
```
tsaglam, line 28: Ideally `"_"` should be a constant.

tsaglam (changes requested):
Just a few minor things...

tsaglam (approved):
Looks good! @nestabentum ready to merge from your end?

nestabentum:
@tsaglam this is good to go

robinmaisch:
Sorry, I'm late to the party, but I noticed that the `ComparisonReportWriterTest` writes test report files to `${baseDir}\jPlag` and does not remove them afterwards. I suggest that they are created in a special output directory and/or removed after the test.

tsaglam:
> 

Should be already fixed by #607

Issue #568, opened 2022-08-03
Scala frontend in development -- question


robinmaisch:
I cannot quite seem to remember the solution that we came up with for this problem, so I'll just quickly put it up for discussion again.

In Scala, operators are just standard methods. They can be called in infix notation or standard notation.

```Scala
val sum = 1 + 2   // evaluates to (1).+(2)
```

In other frontends, operations are not assigned tokens, but merthod calls are. The task would be to try to distingush between the two.
I came up with these solutions, all of which come with a downside.

- treat all methods as such, even operators
  -> more tokens are generated than by other frontends, danger of cluttering; but is consistent

- filter out the operators and only assign a token to "real" methods
   - by a list: `+, -, *, /, %, ..`.
   - by a pattern, e.g. `[\w_]+`
 -> both versions allow to hide method calls, as even standard operators can be overloaded. Is this a risk worth taking?

- decide whether an operator is just standard or a custom implementation
-> that requires type information not available to the parser

- "reject" solutions that use invalid method identifiers
-> this is not the responsibility of JPlag as I understand it

I lack the experience with typical attack patterns and the sensitivity of the algorithm to judge what's best. Please help!

tsaglam:
I would say we ignore the risk that someone overloads operators. Furthermore, I would say custom operators should be treated as methods, while the standard operators are filtered out (by list preferably), even if that includes filtering out overloaded standard operators.

sebinside:
Yes, this is what we discussed. The risk of attack is mitigated by only filtering a defined and limited number of operators (`&&`, `+`, and so on) and treating all other combinations of special characters like methods. In this way, both attack surfaces of A) hiding behind overwritten operators and B) hiding behind fake method calls are minimized.

robinmaisch:
Great! This should apply regardless of whether these standard operators are called in infix or dot notation, right?

tsaglam:
@robinmaisch 
> Great! This should apply regardless of whether these standard operators are called in infix or dot notation, right? 

Yes!

> EDIT: It might not be easy to figure out the method name, because the complete term up to the opening parenthesis is treated as the called function.

Can't you just split the name with a dot and take the last sequence? Or are there cases where this does not work?

PR #570, opened 2022-08-06
Add Scala frontend

robinmaisch:
This PR adds a Scala frontend to JPlag. It comes with some test files and a README which explains various design choices that can be considered for maintenance.
The frontend only supports Scala 2.12 so far. Maybe it can be upgraded to Scala 3 easily in a collective effort, I certainly failed to do it! In that case, a few more token types might be necessary.

Besides the difficulties with the syntax discussed in the README, the token set and the abstraction are designed along the lines of other frontends. 

As the parser takes a different approach that reduces code repetition, it should be considered if the parser mechanism is redesigned in the future.

Addressing #345, #568


jplag.frontend.scala/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+        <artifactId>aggregator</artifactId>
+        <version>${revision}</version>
+    </parent>
+    <artifactId>scala</artifactId>
+
+    <properties>
+        <encoding>UTF-8</encoding>
+        <scala.version>2.12.6</scala.version>
+        <scala.compat.version>2.12</scala.compat.version>
+        <spec2.version>4.2.0</spec2.version>
```
dfuchss, line 16: Where is "spec2.version" used?

jplag.frontend.scala/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+
+    <parent>
+        <groupId>de.jplag</groupId>
+        <artifactId>aggregator</artifactId>
+        <version>${revision}</version>
+    </parent>
+    <artifactId>scala</artifactId>
+
+    <properties>
+        <encoding>UTF-8</encoding>
```
dfuchss, line 13: Why encoding? Managed by parent

jplag.frontend.scala/src/test/java/de/jplag/scala/ScalaFrontendTest.java
```diff
@@ -0,0 +1,181 @@
...
+            if (coverage == 1) {
+                logger.info("All lines covered.");
+            } else {
+                logger.info("Coverage: %.1f%%.".formatted(coverage * 100));
+                logger.info("Missing lines {}", codeLines);
+                if (coverage < 0.9) {
+                    fail("Source coverage is unsatisfactory");
+                }
+            }
+
```
dfuchss, line 119: Comparisons of double should use a Epsilon :)

jplag.frontend.scala/src/test/resources/de/jplag/scala/Parser.scala
```diff
@@ -0,0 +1,376 @@
+package de.jplag.scala
```
dfuchss, line 1: Why is the parser a resource ?

dfuchss (commented):
Some first comments.

jplag.frontend.scala/src/test/resources/de/jplag/scala/Parser.scala
robinmaisch: I used a duplicate of the parser file as a test file :)

jplag.frontend.scala/pom.xml
robinmaisch: This is a leftover from the revised frontend. I'll remove it

jplag.frontend.scala/pom.xml
robinmaisch: Same as above.

jplag.frontend.scala/src/test/java/de/jplag/scala/ScalaFrontendTest.java
robinmaisch: Yes, I suppose an epsilon comparison won't hurt even if 0.9 is a somewhat arbitrary number. This will only fail if something went very wrong.

robinmaisch:
It turns out the IDE settings kept me from successfully upgrading to Scala 2.13, which I now succeeded in doing. Scalameta, the parser library used here, is in development for Scala 3, but not released yet, so it seems that we have to wait for that until Scala 3 can be fully supported.

jplag.frontend.scala/src/main/scala/de/jplag/scala/ScalaToken.scala
```diff
@@ -0,0 +1,36 @@
...
+import scala.meta.inputs.Position
+
+class ScalaToken(tType: Int,
+                 file: java.lang.String,
+                 var line: Int,
+                 var column: Int,
+                 var length: Int)
+  extends Token(tType, file, line, column, length) {
+
+    def this(tType: ScalaTokenConstants.Value, file: String, pos: Position, fromEnd: Boolean) {
```
sebinside, line 14: I don't know about the rules in frontends, but abbreviations like `pos` for `position` are generally viewed as bad practice (*write once, read many*)

jplag.frontend.scala/src/test/resources/de/jplag/scala/Complete.scala
```diff
@@ -0,0 +1,162 @@
+package mypackage
```
sebinside, line 1: I like how comprehensive this test already is, it uses a lot of custom scala features. What I'm currently missing are custom control structures. A method `method(args)` can be either called directly `method(args)` or as control structure, where the body becomes the argument `method { args }`. This could be a potential attack as every function call can be rewritten this way.

sebinside: For more information on that, this ancient blog post might be relevant: https://github.com/sebinside/sebinside-blog/blob/master/blog/md/scalafeatures-teil2.md#9-eigene-kontrollstrukturen

jplag.frontend.scala/src/main/scala/de/jplag/scala/ScalaToken.scala
robinmaisch: Absolutely! Coming right up

jplag.frontend.scala/src/test/java/de/jplag/scala/ScalaFrontendTest.java
```diff
@@ -0,0 +1,182 @@
...
+            // All lines that contain code
+            var codeLines = new ArrayList<>(getCodeLines(lines));
+            // All lines that contain token
+            var tokenLines = IntStream.range(0, tokens.size()).mapToObj(tokens::getToken).mapToInt(Token::getLine).distinct().boxed().toList();
+
+            // Keep only lines that have no tokens
+            codeLines.removeAll(tokenLines);
+
+            double coverage = 1.d - (codeLines.size() * 1.d / (codeLines.size() + tokenLines.size()));
```
SuyDesignz, line 110: you switch between dyamic typing with var and specifying the exact type. var is convenient but makes the code less readable. But this is only my opinion but it should be used constantly.

jplag.frontend.scala/src/main/scala/de/jplag/scala/ScalaToken.scala
robinmaisch: Absolutely! I gave the identifiers a good look again, hopefully I got them all.

jplag.frontend.scala/src/test/resources/de/jplag/scala/Complete.scala
robinmaisch: That's a nice feature! I included an example. The parser luckily treats them the same.

jplag.frontend.scala/src/test/java/de/jplag/scala/ScalaFrontendTest.java
robinmaisch: You are probably right. I still need to get a better feeling for when the type is obvious or distracting, and when the reader will have to guess the type to fully understand the code.

tsaglam (approved):

PR #572, opened 2022-08-06
Convert JPlagOption to Record (→ Read-Only)

### LINKED ISSUE #274: JPlagOptions are modified during the execution

dfuchss:
This PR converts the JPlag Options to a record including a builder pattern (`withX` methods).

Resolves parts of
 #274


jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,323 +1,261 @@
...
+ * This record defines the options to configure {@link JPlag}.
+ * @param languageOption Language to use when parsing the submissions.
+ * @param comparisonMode Determines which strategy to use for the comparison of submissions.
+ * @param debugParser If true, submissions that cannot be parsed will be stored in a separate directory.
+ * @param fileSuffixes List of file suffixes that should be included.
+ * @param similarityThreshold Percentage value (must be between 0 and 100). Comparisons (of submissions pairs) with a
+ * similarity below this threshold will be ignored. The default value of 0 allows all matches to be stored. This affects
+ * which comparisons are stored and thus make it into the result object. See also {@link #similarityMetric()}.
+ * @param maximumNumberOfComparisons The maximum number of comparisons that will be shown in the generated report. If
+ * set to -1 all comparisons will be shown.
```
JanWittler, line 34: Can you refactor this into a constant?

JanWittler (approved):
Thanks for the improvement.
Following the comment in #274, I would also say we should consider (in a future PR) to split it into separate instances. `comparisonMode` and `clusteringOptions` are only relevant for some specific classes, and then we mix debug logging, input description, and some language overrides. Maybe it is hard to completely separate all concepts as we need to store those user options somewhere, we could at least make it hierarchical to avoid this kind-of god-class.

JanWittler (approved):

jplag.cli/src/main/java/de/jplag/CLI.java
```diff
@@ -125,21 +154,27 @@ public JPlagOptions buildOptionsFromArguments(Namespace namespace) {
...
+        JPlagOptions options = new JPlagOptions(language, submissionDirectories, oldSubmissionDirectories) //
+                .withBaseCodeSubmissionName(BASE_CODE.getFrom(namespace)) //
+                .withVerbosity(Verbosity.fromOption(VERBOSITY.getFrom(namespace))) //
+                .withDebugParser(DEBUG.getFrom(namespace)) //
+                .withSubdirectoryName(SUBDIRECTORY.getFrom(namespace)) //
+                .withFileSuffixes(Arrays.stream(fileSuffixes).toList()) //
+                .withExclusionFileName(EXCLUDE_FILE.getFrom(namespace)) //
+                .withMinimumTokenMatch(MIN_TOKEN_MATCH.getFrom(namespace)) //
+                .withSimilarityThreshold(SIMILARITY_THRESHOLD.getFrom(namespace)) //
+                .withMaximumNumberOfComparisons(SHOWN_COMPARISONS.getFrom(namespace));
```
JanWittler, line 167: For brevity (and also to avoid constructing n new option instances) we could also use the full constructor here, right?

dfuchss: depends. It may be the case that we do not want to set all options via CLI. (or we may add more to JPlagOptions later)

JanWittler: I would say it makes the code more compact (and marginally more performant) and we don't use readability as the arguments are identified anyways by their `<Name.getFrom()` syntax.

jplag.cli/src/main/java/de/jplag/CLI.java
```diff
@@ -125,21 +154,15 @@ public JPlagOptions buildOptionsFromArguments(Namespace namespace) {
...
+        var language = LanguageLoader.getLanguage(LANGUAGE.getFrom(namespace)).orElseThrow();
+        var comparisonMode = ComparisonMode.fromName(COMPARISON_MODE.getFrom(namespace));
+        if (comparisonMode.isEmpty()) {
+            logger.warn("Unknown comparison mode, using default mode!");
+            comparisonMode = Optional.of(JPlagOptions.DEFAULT_COMPARISON_MODE);
+        }
 
         ClusteringOptions.Builder clusteringBuilder = new ClusteringOptions.Builder();
-        Optional.ofNullable(!(Boolean) CLUSTER_DISABLE.getFrom(namespace)).ifPresent(clusteringBuilder::enabled);
+        Optional.ofNullable((Boolean) CLUSTER_DISABLE.getFrom(namespace)).ifPresent(enabled -> clusteringBuilder.enabled(!enabled));
```
JanWittler, line 165: naming should probably be `disabled`?

jplag.cli/src/main/java/de/jplag/CLI.java
```diff
@@ -169,9 +192,24 @@ public JPlagOptions buildOptionsFromArguments(Namespace namespace) {
             clusteringBuilder.preprocessor(Preprocessing.THRESHOLD);
             clusteringBuilder.preprocessorPercentile(threshold);
         });
-        options.setClusteringOptions(clusteringBuilder.build());
 
-        return options;
+        return new JPlagOptions( //
+                language, //
+                comparisonMode.orElseThrow(), //
```
JanWittler, line 198: If we use `orElse` when constructing the comparisonMode, we can obtain directly an unwrapped value and thus not need this anymore.

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,265 @@
...
+ * @param baseCodeSubmissionName Path name of the directory containing the base code.
+ * @param subdirectoryName Example: If the subdirectoryName is 'src', only the code inside submissionDir/src of each
+ * submission will be used for comparison.
+ * @param verbosity Level of output verbosity.
+ * @param clusteringOptions Clustering options
+ */
+public record JPlagOptions(Language language, ComparisonMode comparisonMode, boolean debugParser, List<String> fileSuffixes,
+        float similarityThreshold, Integer maximumNumberOfComparisons, SimilarityMetric similarityMetric, Integer minimumTokenMatch,
+        String exclusionFileName, List<String> submissionDirectories, List<String> oldSubmissionDirectories, String baseCodeSubmissionName,
+        String subdirectoryName, Verbosity verbosity, ClusteringOptions clusteringOptions) {
```
JanWittler, line 50: I would reorder the arguments to cluster similar functionality together
```Java
//language frontend
language
minimumTokenMatch

//submissions
submissionDirectories
oldSubmissionDirectories
baseCodeSubmissionName
subdirectoryName

//include / exclude submissions
fileSuffixes
exclusionFileName

//report generation
similarityMetric
similarityThreshold
maximumNumberOfComparisons

//clustering
clusteringOptions

//other
comparisonMode
verbosity
debugParser
```

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,265 @@
...
+    }
+
+    public JPlagOptions(Language language, ComparisonMode comparisonMode, boolean debugParser, List<String> fileSuffixes, float similarityThreshold,
+            Integer maximumNumberOfComparisons, SimilarityMetric similarityMetric, Integer minimumTokenMatch, String exclusionFileName,
+            List<String> submissionDirectories, List<String> oldSubmissionDirectories, String baseCodeSubmissionName, String subdirectoryName,
+            Verbosity verbosity, ClusteringOptions clusteringOptions) {
+        this.language = language;
+        this.comparisonMode = comparisonMode;
+        this.debugParser = debugParser;
+        this.fileSuffixes = fileSuffixes == null ? null : fileSuffixes.stream().toList();
```
JanWittler, line 87: Why `fileSuffixes.stream().toList()?

`Collections.unmodifiableList(fileSuffixes)`

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,265 @@
...
+        this.similarityThreshold = fixSimilarityThreshold(similarityThreshold);
+        this.maximumNumberOfComparisons = fixMaximumNumberOfComparisons(maximumNumberOfComparisons);
+        this.similarityMetric = similarityMetric;
+        this.minimumTokenMatch = minimumTokenMatch;
+        this.exclusionFileName = exclusionFileName;
+        this.submissionDirectories = submissionDirectories == null ? null : submissionDirectories.stream().toList();
+        this.oldSubmissionDirectories = oldSubmissionDirectories == null ? null : oldSubmissionDirectories.stream().toList();
```
JanWittler, line 94: Again why stream()?

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,265 @@
...
+        this.verbosity = verbosity;
+        this.clusteringOptions = clusteringOptions;
     }
 
-    public Optional<String> getBaseCodeSubmissionName() {
-        return Optional.ofNullable(baseCodeSubmissionName);
+    public JPlagOptions withLanguageOption(Language language) {
+        return new JPlagOptions(language, comparisonMode, debugParser, fileSuffixes, similarityThreshold, maximumNumberOfComparisons,
+                similarityMetric, minimumTokenMatch, exclusionFileName, submissionDirectories, oldSubmissionDirectories, baseCodeSubmissionName,
+                subdirectoryName, verbosity, clusteringOptions);
```
JanWittler, line 104: Do we need all of these factory methods? E.g. I would say that the language or also the submission details should be set only when initially instantiating the options.

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,265 @@
...
+     * conditions of the compatibility fallback listed above.
      */
-    public void setLanguageDefaults(Language language) {
-        if (!hasMinimumTokenMatch()) {
-            setMinimumTokenMatch(language.minimumTokenMatch());
+    @Override
+    public String baseCodeSubmissionName() {
+        return baseCodeSubmissionName;
+    }
+
```
JanWittler, line 224:
```suggestion
```

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,265 @@
...
-    public void setOldSubmissionDirectories(List<String> oldSubmissionDirectories) {
-        this.oldSubmissionDirectories = oldSubmissionDirectories;
-    }
-
-    public void setSimilarityMetric(SimilarityMetric similarityMetric) {
-        this.similarityMetric = similarityMetric;
     }
 
-    public void setSimilarityThreshold(float similarityThreshold) {
+    private static float fixSimilarityThreshold(float similarityThreshold) {
```
JanWittler, line 250:
```suggestion
    private static float normalizeSimilarityThreshold(float similarityThreshold) {
```

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,265 @@
...
         } else {
-            this.similarityThreshold = similarityThreshold;
+            return similarityThreshold;
         }
     }
 
-    public void setSubdirectoryName(String subdirectoryName) {
-        // Trim problematic file separators.
-        this.subdirectoryName = (subdirectoryName == null) ? null : subdirectoryName.replace(File.separator, "");
+    private Integer fixMaximumNumberOfComparisons(Integer maximumNumberOfComparisons) {
```
JanWittler, line 262:
```suggestion
    private Integer normalizeMaximumNumberOfComparisons(Integer maximumNumberOfComparisons) {
```

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,265 @@
...
-        if (!hasMinimumTokenMatch()) {
-            setMinimumTokenMatch(language.minimumTokenMatch());
+    @Override
+    public String baseCodeSubmissionName() {
+        return baseCodeSubmissionName;
+    }
+
+    @Override
+    public Integer minimumTokenMatch() {
+        Integer fixedMinimumTokenMatch = (minimumTokenMatch != null && minimumTokenMatch < 1) ? Integer.valueOf(1) : minimumTokenMatch;
```
JanWittler, line 227: For other values that have to be within a certain range, you normalize them upon construction. Why not do that here too?

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,265 @@
...
+    }
+
+    private Set<String> readExclusionFile(final String exclusionFileName) {
+        try (BufferedReader reader = new BufferedReader(new FileReader(exclusionFileName, JPlagOptions.CHARSET))) {
+            final var excludedFileNames = reader.lines().collect(Collectors.toSet());
+            if (verbosity() == LONG) {
+                logger.info("Excluded files:");
+                for (var excludedFilename : excludedFileNames) {
+                    logger.info(excludedFilename);
+                }
```
JanWittler, line 241:
```suggestion
                logger.info("Excluded files:\n" + String.join("\n", excludedFileNames);
```

JanWittler (changes requested):

jplag.cli/src/main/java/de/jplag/CLI.java
dfuchss:
```suggestion
        Optional.ofNullable((Boolean) CLUSTER_DISABLE.getFrom(namespace)).ifPresent(disabled -> clusteringBuilder.enabled(!disabled));
```

dfuchss: good catch :)

jplag/src/main/java/de/jplag/options/JPlagOptions.java
dfuchss: I would like to have factories for all options. Otherwise, we have to discuss use cases for any option.

jplag/src/main/java/de/jplag/options/JPlagOptions.java
dfuchss: This is just for the documentation of the getter. I don't have a better idea here

jplag/src/main/java/de/jplag/options/JPlagOptions.java
dfuchss: good catch. I'll fix it

jplag/src/main/java/de/jplag/options/JPlagOptions.java
JanWittler: Documented in https://github.com/jplag/JPlag/issues/625

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,252 @@
...
+                subdirectoryName, fileSuffixes, exclusionFileName, similarityMetric, similarityThreshold, maximumNumberOfComparisons,
+                clusteringOptions, comparisonMode, verbosity, debugParser);
     }
 
-    public SimilarityMetric getSimilarityMetric() {
-        return similarityMetric;
+    public JPlagOptions withSimilarityThreshold(float similarityThreshold) {
+        return new JPlagOptions(language, minimumTokenMatch, submissionDirectories, oldSubmissionDirectories, baseCodeSubmissionName,
+                subdirectoryName, fileSuffixes, exclusionFileName, similarityMetric, normalizeSimilarityThreshold(similarityThreshold),
+                maximumNumberOfComparisons, clusteringOptions, comparisonMode, verbosity, debugParser);
```
JanWittler, line 115: normalisation is already done in constructor, thus not needed here

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,252 @@
...
     }
 
-    public float getSimilarityThreshold() {
-        return similarityThreshold;
+    public JPlagOptions withMaximumNumberOfComparisons(Integer maximumNumberOfComparisons) {
+        return new JPlagOptions(language, minimumTokenMatch, submissionDirectories, oldSubmissionDirectories, baseCodeSubmissionName,
+                subdirectoryName, fileSuffixes, exclusionFileName, similarityMetric, similarityThreshold,
+                normalizeMaximumNumberOfComparisons(maximumNumberOfComparisons), clusteringOptions, comparisonMode, verbosity, debugParser);
```
JanWittler, line 121: normalisation is already done in constructor, thus not needed here

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,252 @@
...
-        if (baseCodeSubmissionName == null || baseCodeSubmissionName.isEmpty()) {
-            this.baseCodeSubmissionName = null;
-        } else {
-            this.baseCodeSubmissionName = baseCodeSubmissionName;
-        }
+    public JPlagOptions withBaseCodeSubmissionName(String baseCodeSubmissionName) {
+        return new JPlagOptions(language, minimumTokenMatch, submissionDirectories, oldSubmissionDirectories,
+                (baseCodeSubmissionName == null || baseCodeSubmissionName.isBlank()) ? null : baseCodeSubmissionName, subdirectoryName, fileSuffixes,
+                exclusionFileName, similarityMetric, similarityThreshold, maximumNumberOfComparisons, clusteringOptions, comparisonMode, verbosity,
+                debugParser);
```
JanWittler, line 158: Normalisation should be done in constructor and then removed here

jplag/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -1,320 +1,251 @@
...
+    public JPlagOptions withClusteringOptions(ClusteringOptions clusteringOptions) {
+        return new JPlagOptions(language, minimumTokenMatch, submissionDirectories, oldSubmissionDirectories, baseCodeSubmissionName,
+                subdirectoryName, fileSuffixes, exclusionFileName, similarityMetric, similarityThreshold, maximumNumberOfComparisons,
+                clusteringOptions, comparisonMode, verbosity, debugParser);
     }
 
-    public void setExclusionFileName(String exclusionFileName) {
-        this.exclusionFileName = exclusionFileName;
+    public boolean hasBaseCode() {
+        return baseCodeSubmissionName != null && !baseCodeSubmissionName.isBlank();
```
JanWittler, line 179:
```suggestion
        return baseCodeSubmissionName != null;
```

Since we normalize the attribute, the 2nd check is not needed here anymore

JanWittler (approved):

Issue #548, opened 2022-07-31
End to end testing - "comparative values"


SuyDesignz:
@jplag/studdev 
@jplag/maintainer

The following issue will refer to the issue ([endToEndTesting-Issue](https://github.com/jplag/JPlag/issues/193) #193) and should be about the comparison values for the end-to-end testing.
Suitable comparison values for endToEndTesting are to be evaluated as well as their persistence in the project.

First of all, I wanted to list the possible properties and explain which values are available for the tests at all.
The endToEnd tests are performed using the JPlag API:
``` java
JPlagOptions options = new JPlagOptions(submissionTestPaths, new ArrayList<String>(), LanguageOption.JAVA);  
JPlag jplag = new JPlag(options);  
JPlagResult result = jplag.run();  
//for the fine-granular testing strategy, the result object contains only one comparison.
for (JPlagComparison jPlagComparison : result.getComparisons()) {
     System.out.println("Comparison of the stored values and the current equality values");
     assertEquals(Float.parseFloat(resultFromXML), jPlagComparison.similarity());
}
```
In the code example above, the similarity value of the test is compared with the stored values.

## Access to the possible comparative values
The returned result object is [JPlagComparison](https://github.com/jplag/JPlag/blob/master/jplag/src/main/java/de/jplag/JPlagComparison.java) which can be used to access the values of the comparison.
```java
[...]
    private final Submission firstSubmission;
    private final Submission secondSubmission;

    private final List<Match> matches;

    public JPlagComparison(Submission firstSubmission, Submission secondSubmission) {
        this.firstSubmission = firstSubmission;
        this.secondSubmission = secondSubmission;
        matches = new ArrayList<>();
    }
[...]
```
This is built up from two [Submission-objects](https://github.com/jplag/JPlag/blob/master/jplag/src/main/java/de/jplag/Submission.java) and a List\<[Match](https://github.com/jplag/JPlag/blob/master/jplag/src/main/java/de/jplag/Match.java)\> 

## Possible comparative values
The similarity(float) is already used as a comparative value in the current tests. 
The following possible comparison values must be discussed:
1. Here, the values "maximalSimilarity()" and "minimalSimilarity()" are additionally available. 
2. In the Match object length(int), startOfFirst(int) and startOfSecond(int) can be accessed.
3. The compared submission objects contain other possible comparative values (for every submission)
3.1 length of the token list 
3.2 the token list contains the JavaTokens with type number, hash value, length of the token and the line number
[...]

PR #583, opened 2022-08-10
End To End Test Suite extended

SuyDesignz:
## Extension Of The End To End Module
The extension of the EndToEnd tests includes the following new features/Fixes:
- (Fix) Identification in json changed to SHA256 hash
- (feature) extension of the comparative values as discussed in the issue [End to end testing - "comparative values"](https://github.com/jplag/JPlag/issues/548)
- (Fix) made json more flexible to better store results per test function
- current json format looks like this :
``` json
[...]
 "function_name" : "normalizationLevelTest",
  "test_results" : [ {
    "result_similarity" : 100.0,
    "result_minimal_similarity" : 100.0,
    "result_maximum_similarity" : 100.0,
    "result_matched_token_number" : 56,
    "test_identifier" : "85FF00F531A497F002D40E9C8430CB159EFC40E618925FD757E35EB2A533227E"
  },
[...]
```

- (feature) Made adjustments to the use of paths in a mapper to be able to add languages flexibly `LanguageToPathMapper.java`
- (feature) Create the memory tests to insert the temporary results into the memory json file. This is necessary to be able to save changes made to languages automatically so that the endToEnd tests run successfully again after a change made to the recognition.
- (feature) Documentation written to the README.md. Here is a step by step guide how to create new endToEnd tests for languages
- (feature) The endToEnd test for java extended
- (feature) Made it possible to create a quick over all test with the help of `JPlagTestSuiteHelper.java`
``` java
[...]
    @Test
    void overAllTests() throws IOException, ExitException, NoSuchAlgorithmException {
        String[] testClassNames = jplagTestSuiteHelper.getAllTestFileNames();
        runJPlagTestSuite(testClassNames);
    }
[...]
```

For more necessary information please have a look at the README.md file.

### Meeting summary
**DONE:**
- [X] Make adjustment to identifiktaor and use the test data names instead of the SHA256 hash
- [X] Remove the "result_similarity" field in the json file as this is composed of the max and min similarity.
- [X] View JUnit DynamicTest and its use in the OverAllTests cases.
- [X] Extend with the possibility to run other test configurations
- [x] Modify the README.md file to add the new endToEndTests


SuyDesignz:
Regarding the buidl error I am already looking for the problem. Likewise, I have local problem to build the project, since the module of Scala throws reference error with me

SuyDesignz:
There seems to be problem with comparing the SHA256 hashes created on my windows machine and the one created by the unix build.

dfuchss:
Thats may be related to line endings (\r\n for windows \n for windows). How did you created the SHA? (In Java or Locally)

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
```diff
@@ -64,8 +75,103 @@ public TestCaseModel createNewTestCase(String[] classNames, String functionName)
...
+        String testFileNamesInFirstSubmission = new String();
+        String testFileNamesInSecondSubmission = new String();
+        for (File file : jPlagComparison.getFirstSubmission().getFiles()) {
+            testFileNamesInFirstSubmission += file.getName().toString();
+        }
+        for (File file : jPlagComparison.getSecondSubmission().getFiles()) {
+            testFileNamesInSecondSubmission += file.getName().toString();
+        }
+        MessageDigest digest = MessageDigest.getInstance("SHA-256");
+        byte[] encodedhash = digest.digest((testFileNamesInFirstSubmission + testFileNamesInSecondSubmission).getBytes(StandardCharsets.UTF_8));
```
dfuchss, line 133: Split the files into lines and aggregate them by a fixed line ending (e.g., \n)

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
```diff
@@ -64,8 +75,103 @@ public TestCaseModel createNewTestCase(String[] classNames, String functionName)
...
+
+        StringBuilder stringBuilder = new StringBuilder();
+        for (byte nextByte : encodedhash) {
```
dfuchss, line 136: new String(byte[], ..)  or similar should exist

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
SuyDesignz: Thanks for the quick reply! I change it as soon as possible and then push the changes

SuyDesignz:
### Meeting summary
- Make adjustment to identifiktaor and use the test data names instead of the SHA256 hash
- remove the "result_similarity" field in the json file as this is composed of the max and min similarity. 
- Extend with the possibility to run other test configurations.
- View JUnit DynamicTest and its use in the OverAllTests cases.

JanWittler:
> * View JUnit DynamicTest and its use in the OverAllTests cases.

https://www.baeldung.com/junit5-dynamic-tests

Here is an example. In best case the dynamic tests would be generated nested, but I didn't find this feature with a quick search. But I think it would be fine to have one test factory per test suite which simply generated the tests. If I add a test suite in the future, the overhead of adding that one method is acceptable low.
With generated nested, I mean a test structure like this
```
- Test Suite 1
 - Test A1 <-> A2
 - Test A1 <-> A3
 - Test A2 <-> A3
- Test Suite 2
 - Test B1 <-> B2
 - ... 
```

SuyDesignz:
### Meeting summary
**DONE:**
- [X] Make adjustment to identifiktaor and use the test data names instead of the SHA256 hash
- [X] Remove the "result_similarity" field in the json file as this is composed of the max and min similarity.

**TODO:**
- [ ] Extend with the possibility to run other test configurations.
- [ ] View JUnit DynamicTest and its use in the OverAllTests cases.

SuyDesignz:
I found something here about the nested junit5 dynamic tests which were mentioned in the meeting. I'll have a look at the topic and then make changes if necessary.

[Beyond the simple: An in-depth look at JUnit 5’s nested tests, dynamic tests, parameterized tests, and extensions](https://blogs.oracle.com/javamagazine/post/beyond-the-simple-an-in-depth-look-at-junit-5s-nested-tests-dynamic-tests-parameterized-tests-and-extensions)

SuyDesignz:
### The implemented code for the Dynamic tests could look something like this:

``` java
	@TestFactory
	Stream<DynamicTest> dynamicTestsWithIterator() {
		//test cases calculated with (n over k)
		var fileNames = jplagTestSuiteHelper.getAllTestFileNames();
		ArrayList<String[]> testCases = new ArrayList<>();
		int outerCounter = 1;
		for (String fileName : fileNames) {
			for (int counter = outerCounter; counter < fileNames.length; counter++) {
				testCases.add(new String[] { fileName, fileNames[counter] });
			}
			outerCounter++;
		}
		return testCases.stream().map(dom -> DynamicTest.dynamicTest("Testing: " + dom[0] + " " + dom[1], () -> {
			runJPlagTestSuite(dom);
			jplagTestSuiteHelper.clear();
		}));
	}
```

But:
[**The DynamicTests are executed differently than the standard @Tests and do not support lifecycle callbacks. Meaning, the @BeforeEach and the @AfterEach methods will not be called for the DynamicTests.**  ](https://www.baeldung.com/junit5-dynamic-tests)
- this means that the jplagTestSuiteHelper.clear() function must be executed on itself in the test case.
- Also the test time increases from about 10 seconds to over 30. If a language has even more test data, this could lead to a significantly increased test runtime.

SuyDesignz:
To avoid having to write a separate [TestExecutionListener](https://junit.org/junit5/docs/5.0.3/api/org/junit/platform/launcher/TestExecutionListener.html), the test case is now handled with cry-finally 

``` java 
 return testCases.stream().map(testCase -> DynamicTest.dynamicTest("Testing: " + testCase[0] + " " + testCase[1], () -> {
            try {
                runJPlagTestSuite(testCase, functionName);
            } finally {
                jplagTestSuiteHelper.clear();
            }
        }));
```

tsaglam:
> Also the test time increases from about 10 seconds to over 30. If a language has even more test data, this could lead to a significantly increased test runtime.

The end-to-end tests are only supposed to run for the Java language frontend. So I think this should be fine.

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JavaEndToEndTest.java
```diff
@@ -100,15 +143,28 @@ void tokenGenerationLevelTest(String[] testClassNames, int testIdentifier) throw
...
+        for (JPlagComparison jPlagComparison : currentJPlagComparison) {
+            String hashCode = jplagTestSuiteHelper.getTestIdentifier(jPlagComparison);
+            ResultModel resultModel = testCaseModel.getCurrentJsonModel().getResultModelById(hashCode);
+            assertNotNull(resultModel, "No stored result could be found for the identifier! " + hashCode);
+            assertEquals(resultModel.getMinimalSimilarity(), jPlagComparison.minimalSimilarity(),
+                    "The JPlag results [minimalSimilarity] do not match the stored values!");
+            assertEquals(resultModel.getMaximalSimilarity(), jPlagComparison.maximalSimilarity(),
+                    "The JPlag results [maximalSimilarity] do not match the stored values!");
+            assertEquals(resultModel.getNumberOfMatchedTokens(), jPlagComparison.getNumberOfMatchedTokens(),
+                    "The JPlag results [numberOfMatchedTokens] do not match the stored values!");
```
JanWittler, line 167: It would be great to first collect all incorrect results. For example a `List<String> validationErrors` could be filled with error descriptions for incorrect values. Then at the end that array is checked for not empty. This has the benefit that the test failure description logs all incorrect values instead of only the first incorrect one.
I.e. test failure description:
```
There were <count> validation error(s):
- min similarity was <a> but expected <b>
- max similarity was <c> but expected <d>
```

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
```diff
@@ -64,8 +76,120 @@ public TestCaseModel createNewTestCase(String[] classNames, String functionName)
...
+    public String getTestIdentifier(JPlagComparison jPlagComparison) {
+
+        String testFileNamesInFirstSubmission = getEnclosedFileNamesFromCollection(jPlagComparison.getFirstSubmission().getFiles());
+        String testFileNamesInSecondSubmission = getEnclosedFileNamesFromCollection(jPlagComparison.getSecondSubmission().getFiles());
+
+        int compaire = testFileNamesInFirstSubmission.compareTo(testFileNamesInSecondSubmission);
+
+        String returnIdentifier;
+
+        if (compaire != 0) {
```
JanWittler, line 129: File names should never be equal as we control the file names. Thus we can simply have `if (a < b) a + b else b + a`. Also I would suggest to introduce a separator between the file names for readability. Maybe a `-` as underscores are already used in the file names?

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JavaEndToEndTest.java
```diff
@@ -50,48 +59,82 @@ public void teardown() throws IOException {
...
     /**
      * Inserting comments or empty lines (normalization level) -> id = 0 Changing variable names or function names
      * (normalization level) -> id = 1 Inserting comments or empty lines (normalization level) -> id = 2
      * @param testClassNames Plagiarized classes names in the resource directorie which are needed for the test
-     * @param testIdentifier name of the testId to load and identify the stored results
      * @throws IOException is thrown in case of problems with copying the plagiarism classes
      * @throws ExitException in case the plagiarism detection with JPlag is preemptively terminated would be of the test.
+     * @throws NoSuchAlgorithmException when no hash algorithm could be found
+     * @throws NameNotFoundException if the no filenames cloud be found in the JPlagCOmparison object
      */
```
JanWittler, line 85: Good that you explicitly specify what kind of plagiarism actually happens in the files. But in the method header it is a bit hard to discover from my opinion. Maybe we add the kind of performed plagiarism as a comment at the start of each source file? If the original submission also has a comment that we could even say that it doesn't change the structure, but comments should not be tokenized anyways.

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/model/TestCaseModel.java
```diff
@@ -41,6 +42,11 @@ public JsonModel getCurrentJsonModel() {
      * @return current JPlag options for the created object
      */
     public JPlagOptions getJPlagOptionsFromCurrentModel() {
-        return new JPlagOptions(new ArrayList<>(Arrays.asList(submissionFolderPath)), new ArrayList<>(), languageOption);
+        JPlagOptions jplagOptions = new JPlagOptions(new ArrayList<>(Arrays.asList(submissionFolderPath)), new ArrayList<>(), languageOption);
+
+        jplagOptions.setComparisonMode(ComparisonMode.PARALLEL);
+        jplagOptions.setComparisonMode(ComparisonMode.NORMAL);
```
JanWittler, line 48: ?

JanWittler (commented):
There are two major points I want to address. These should probably have already come when reviewing the first PR but it seems we were a bit sloppy with that.

## Result JSON Structure

Currently, you identify the results based on the function name. However, the results are identical independent of the function (assuming the same configuration is used). Thus, there is no need to duplicate these configurations for each function. Instead, I would suggest to use one JSON per test suite. This should match the name of the test suite to be easily discoverable, i.e. right now `SortAlgo.json`.

Additionally, you use the test identifier to lookup the test. Thus, we should convert the array with the identifier as a test attribute to a map with the identifier as the key. We can also reduce the other names in the json for readability.
Considering #590 , I propose a structure like
```
[{ 
  "options": {"minimum_match_length" : <value>},
  "tests": [
    { 
      <identifier1> : {"min_similarity" : <value>, [...other values...] }
    }, {...}]
}, {...}]
```

## Test pipeline

Right now, intermediate files describing the similarity (`saveTemporaryResult`) are stored for each test run and removed on success. I guess this was done for being able to check the values when a test failed. However, as you already nicely upgraded to dynamic tests (which I would adapt for the other two tests), we can provide all detailed similarity deviation information in the test failure description. Thus, these intermediate results are not needed anymore. Then, we can also simplify the test suite generator to take the submissions, run JPlag on them and output the final JSON instead having the intermediate representation. As a consequence, the intermediate representation can be removed.

#### Performance

You stated that the tests now perform far slower (30s instead of 10s). I guess the problem is that with the dynamic test implementation, JPlag is executed pairwise, i.e. for each submission pair a new JPlag instance is created. This includes that each submission has to be parsed, tokenized and hashed multiple times. It might be possible to perform the JPlag comparison only once and then simply generate dynamic tests for each `JPlagComparison`. However, this is not necessary to be done as part of this PR. 30s is still an acceptable performance.

## Other minor things

- `JPlagTestSuiteHelper` currently contains methods for the end2end tests as well as for the test suite generation. This should be separated.
- All classes in `/model` can probably be refactored to records
- It should be possible to avoid the cost of copying all test files in `JPlagTestSuiteHelper.createNewTestCaseDirectory` by instead setting `JPlag.excludedFiles`.

Again, I am sorry that this major feedback comes so late. If you have questions or need support, don't hesitate to contact me directly.

SuyDesignz:
First of all, thank you for the detailed comments and remarks. Based on the comments I will have to come up with a new structure for the json data and the tests. I hope it is enough time for me to rebuild everything. If I have questions or comments I would simply create a new issue.

JanWittler:
I tried to help you save some time by already providing an adjusted JSON scheme

> ```
> [{ 
>   "options": {"minimum_match_length" : <value>},
>   "tests":
>     { 
>       <identifier1> : {"min_similarity" : <value>, [...other values...] }
>       <identifier2> : {...},
>       [...]
>     }
> }, {...}]
> ```

So the model would be something like (pseudo-code) this:
```
ResultDescription {
  Options options;
  Map<String, ExpectedResult> identifierToResultMap;

  Options {
    int minimumMatchLength;
  }

  ExpectedResult {
    int minSimilarity;
    int maxSimilarity;
    int numberOfMatchedTokens;
  }
}
```
Hope that helps.

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
SuyDesignz: Changed to:

``` java
/**
     * Creates a unique identifier from the submissions in the JPlagComparison object which is used to find the results in
     * the json files.
     * @param jPlagComparison object from which the hash should be generated
     * @return unique identifier for test case recognition
     */
    public static String getTestIdentifier(JPlagComparison jPlagComparison) {

        String testFileNamesInFirstSubmission = getEnclosedFileNamesFromCollection(jPlagComparison.getFirstSubmission().getFiles());
        String testFileNamesInSecondSubmission = getEnclosedFileNamesFromCollection(jPlagComparison.getSecondSubmission().getFiles());

        int compaire = testFileNamesInFirstSubmission.compareTo(testFileNamesInSecondSubmission);

        return compaire < 0 ? testFileNamesInFirstSubmission + "_" + testFileNamesInSecondSubmission
                : testFileNamesInSecondSubmission + "_" + testFileNamesInFirstSubmission;
    }
    
    /**
     * Merges all contained filenames together without extension
     * @param files whose names are to be merged
     * @return merged filenames
     */
    private static String getEnclosedFileNamesFromCollection(Collection<File> files) {
        StringBuilder stringBuilder = new StringBuilder(files.size());
        for (File file : files) {
            String fileName = file.getName();
            stringBuilder.append(fileName.substring(0, fileName.lastIndexOf('.')));
        }

        return stringBuilder.toString();
    }
```

JanWittler: As said, I would go with "-" as "_" is already used in the file names itself.
Also it is `compare` not `compaire`

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/model/TestCaseModel.java
SuyDesignz: Forgot to remove

SuyDesignz:
**Concept code only! has not been cleaned yet**

After I rewrote the whole module, you get such a json file with the name sortAlgo.json. Here it is important to say that now no new language tests must be added manually! As soon as a folder with the language string in the resoucren folder is present and there test data are, these are taken over with into the test. It is important that the upper folder is named exactly like the language option.toString. Also from these new folders new tests are generated. 

``` java
	public static Map<LanguageOption, Map<String, Path>> getAllLanguageResources() {
		String[] languageDirectoryNames = FileHelper
				.getAllDirectoriesInPath(TestDirectoryConstants.BASE_PATH_TO_LANGUAGE_RESOURCES);
		List<LanguageOption> languageInPathList = FileHelper.getLanguageOptionsFromPath(languageDirectoryNames);
		var returnMap = new HashMap<LanguageOption, Map<String, Path>>();

		for (LanguageOption languageOption : languageInPathList) {
			var tempMap = new HashMap<String, Path>();
			var allDirectoriesInPath = FileHelper.getAllDirectoriesInPath(Path
					.of(TestDirectoryConstants.BASE_PATH_TO_LANGUAGE_RESOURCES.toString(), languageOption.toString()));
			Arrays.asList(allDirectoriesInPath)
					.forEach(directory -> tempMap.put(Path.of(directory).toFile().getName(),
							Path.of(TestDirectoryConstants.BASE_PATH_TO_LANGUAGE_RESOURCES.toString(),
									languageOption.toString(), directory)));
			returnMap.put(languageOption, tempMap);
		}
		return returnMap;
	}
```

### result json for a test folder:
``` json
[ {
  "options" : {
    "minimum_token_match" : 1
  },
  "tests" : {
    "SortAlgo1_2_SortAlgo1_3" : {
      "result_minimal_similarity" : 82.89474,
      "result_maximum_similarity" : 84.0,
      "result_matched_token_number" : 63
    },
    "SortAlgo1_2_SortAlgo1_4" : {
      "result_minimal_similarity" : 76.31579,
      "result_maximum_similarity" : 98.305084,
      "result_matched_token_number" : 58
    },
    "SortAlgo1_3_SortAlgo1_4" : {
      "result_minimal_similarity" : 77.333336,
      "result_maximum_similarity" : 98.305084,
      "result_matched_token_number" : 58
    },
    "SortAlgo_SortAlgo1_2" : {
      "result_minimal_similarity" : 73.68421,
      "result_maximum_similarity" : 100.0,
      "result_matched_token_number" : 56
    },
    "SortAlgo_SortAlgo1_4" : {
      "result_minimal_similarity" : 94.91525,
      "result_maximum_similarity" : 100.0,
      "result_matched_token_number" : 56
    },
    "SortAlgo_SortAlgo1_3" : {
      "result_minimal_similarity" : 74.666664,
      "result_maximum_similarity" : 100.0,
      "result_matched_token_number" : 56
    }
  }
}, {
  "options" : {
    "minimum_token_match" : 15
  },
  "tests" : {
    "SortAlgo1_2_SortAlgo1_3" : {
      "result_minimal_similarity" : 52.63158,
      "result_maximum_similarity" : 53.333332,
      "result_matched_token_number" : 40
    },
    "SortAlgo1_2_SortAlgo1_4" : {
      "result_minimal_similarity" : 40.789474,
      "result_maximum_similarity" : 52.542374,
      "result_matched_token_number" : 31
    },
    "SortAlgo1_3_SortAlgo1_4" : {
      "result_minimal_similarity" : 32.0,
      "result_maximum_similarity" : 40.677967,
      "result_matched_token_number" : 24
    },
    "SortAlgo_SortAlgo1_2" : {
      "result_minimal_similarity" : 59.210526,
      "result_maximum_similarity" : 80.35714,
      "result_matched_token_number" : 45
    },
    "SortAlgo_SortAlgo1_4" : {
      "result_minimal_similarity" : 52.542374,
      "result_maximum_similarity" : 55.357143,
      "result_matched_token_number" : 31
    },
    "SortAlgo_SortAlgo1_3" : {
      "result_minimal_similarity" : 53.333332,
      "result_maximum_similarity" : 71.42857,
      "result_matched_token_number" : 40
    }
  }
} ]
```
### DynamicTest
As already mentioned, there is now only one testFactory for all tests and languages with the permutation of test data and options. The displayed name for the test results from the language, the configuration and the tested data (JAVA: (15) SortAlgo_SortAlgo1)
``` java 
@TestFactory
	Collection<DynamicTest> dynamicOverAllTest() {
		for (Entry<LanguageOption, Map<String, Path>> languageMap : LanguageToTestCaseMapper.entrySet()) {
			LanguageOption currentLanguageOption = languageMap.getKey();
			for (Entry<String, Path> languagePaths : languageMap.getValue().entrySet()) {
				String[] fileNames = FileHelper.loadAllTestFileNames(languagePaths.getValue());
				var testCases = JPlagTestSuiteHelper.getPermutation(fileNames, languagePaths.getValue());
				var testCollection = new ArrayList<DynamicTest>();
				for (Options option : options) {
					for (var testCase : testCases) {
						testCollection.add(DynamicTest
								.dynamicTest(getTestCaseDisplayName(option, currentLanguageOption, testCase), () -> {
									try {
										runJPlagTestSuite(languagePaths.getValue().getFileName().toString(), option,
												currentLanguageOption, testCase);
									} finally {
										JPlagTestSuiteHelper.clear();
									}
								}));
					}
				}
				return testCollection;
			}
		}
		return null;
	}
```
### Models
The models are as follows and are based on your examples.
``` java
public class ExpectedResult {

	@JsonProperty("result_minimal_similarity")
	private float resultSimilarityMinimum;
	@JsonProperty("result_maximum_similarity")
	private float resultSimilarityMaximum;
	@JsonProperty("result_matched_token_number")
	private int resultMatchedTokenNumber;

	public ExpectedResult(float resultSimilarityMinimum, float resultSimilarityMaximum,
			int resultMatchedTokenNumber) {
		this.resultSimilarityMinimum = resultSimilarityMaximum;
		this.resultSimilarityMaximum = resultSimilarityMaximum;
		this.resultMatchedTokenNumber = resultMatchedTokenNumber;
	}
```

``` java
public class Options  {
	@JsonProperty("minimum_token_match")
	private int minimumTokenMatch;
	
	public Options(int minimumTokenMatch)
	{
		this.minimumTokenMatch = minimumTokenMatch;
	}
	
	public Options(JPlagOptions jplagOptions)
	{
		this.minimumTokenMatch = jplagOptions.getMinimumTokenMatch();
	}
```

``` java
public class ResultDescription {
	@JsonIgnore
	private LanguageOption languageOption;
	@JsonProperty("options")
	Options options;
	@JsonProperty("tests")
	Map<String, ExpectedResult> identifierToResultMap;
	
	public ResultDescription(Options options, Map<String, ExpectedResult> identifierToResultMap)
	{
		this.options = options;
		this.identifierToResultMap = identifierToResultMap;
	}
	
	public ResultDescription(Options options , JPlagComparison jPlagComparison , LanguageOption languageOption)
	{
		this.languageOption = languageOption;
		this.options = options;
		identifierToResultMap = new HashMap<>();
		identifierToResultMap.put(JPlagTestSuiteHelper.getTestIdentifier(jPlagComparison), new ExpectedResult(jPlagComparison));
	}
```

### Summary
I'm not quite done with the changes yet but this could possibly be enough for the first feedbacks. to make the test suite runnable again I need some more time.
The most important changes are:
- Revision of the json and models
- only one dynamic test for all languages and for all test data
- to add new language endToEnd tests you only have to add the test data in `test/resources/LANGUAGE/...` 
-> important is that the top folder is named like the enum type LanguageOption you want to test

SuyDesignz:
There is only completely uncleaned code in this branch! it is only meant to be looked at, as I haven't had time to clean it up yet.

JanWittler:
Overall looks good to me :)
Some hints:
- The code of `dynamicOverAllTest` returns after having collected tests for one language instead of for all. However, as Timur said earlier, it is totally sufficient if the tests only support Java as of now. So if there are problems adjusting it, you could also simply remove the options for supporting dynamic languages.
- I would trimm the "result_" prefix from the JSON values for brevity
- All model classes should be [records](https://www.baeldung.com/java-record-keyword)
- I don't get the necessity for the 2nd constructor in both the `Options` and `ResultDescription` classes. From my opinion they should only be used for deserialising the JSON, why do I need to initialize them with the other values? Especially for the `ResultDescription` I think adding the language option and allowing initialization with a `JPlagComparison` object mixes concepts here.

SuyDesignz:
Thanks for the quick feedback! I'm not sure about the dynamic language loading removal. Currently it only supports java, but if you want to test more than one language endToEnd, I think it's handy to only have to create the testata. 
Timur what do you think about this?

JanWittler:
> The end-to-end tests are only supposed to run for the Java language frontend. So I think this should be fine.

[by Timur](https://github.com/jplag/JPlag/pull/583#issuecomment-1214846836)

If you can make it work for various languages without too much effort, it's a nice to have but don't spend too much time on it.

SuyDesignz:
It is already implemented and would be more effort to remove it again than to take it further.

tsaglam:
@SuyDesignz can you resolve all review comments that you addressed? If you are waiting for feedback in a comment just tag the reviewer again.

SuyDesignz:
I tried to change the classes of the json objects to records. Here I ran into some problems with the deserialization. Apparently there are already some entries about this :
[Java 14/15 records can not set final field #1794](https://github.com/google/gson/issues/1794)
[Properties naming strategy do not work with Record #2992](https://github.com/FasterXML/jackson-databind/issues/2992)
[@JsonNaming does not work with Java 16 record types #3102](https://github.com/FasterXML/jackson-databind/issues/3102)

my conclusion is that I will not change the classes in records. I haven't found a working WorkAround yet. Class is attached.

```java

@JsonDeserialize(as = ResultDescription.class)
public record ResultDescription(@JsonIgnore LanguageOption languageOption, 
		@JsonProperty("options") Options options,
		@JsonProperty("tests") Map<String, ExpectedResult> identifierToResultMap,
		@JsonIgnore JPlagComparison jplagComparison) {
	
    public ResultDescription(LanguageOption languageOption, Options options, Map<String, ExpectedResult> identifierToResultMap, JPlagComparison jplagComparison){
    	this.identifierToResultMap = identifierToResultMap;
    	this.options = options;
    	this.languageOption = languageOption;
    	this.jplagComparison = jplagComparison;
    }

	public ResultDescription(Options options, Map<String, ExpectedResult> identifierToResultMap)
	{
		this(null , options , identifierToResultMap , null);
	}

	@JsonIgnore
	public Map<String, ExpectedResult> getIdentifierResultMap()
	{
		return identifierToResultMap;
	}
	@JsonIgnore
	public ExpectedResult getExpectedResultByIdentifier(String identifier)
	{
		return identifierToResultMap.get(identifier);
	}
	
	@JsonIgnore
	public Options getOptions()
	{
		return options;		
	}
	@JsonIgnore
	public LanguageOption getLanguageOption()
	{
		return languageOption;
	}
	
	public void putIdenfifierToResultMap(String identifier, ExpectedResult expectedResult) {
		identifierToResultMap.put(identifier, expectedResult);
	}
	
	@JsonDeserialize(as = Options.class)
	public record Options(@JsonProperty("minimum_token_match") int minimumTokenMatch)
	{
		public Options(int minimumTokenMatch)
		{
			this.minimumTokenMatch = minimumTokenMatch;
		}

		@JsonIgnore
		public int getMinimumTokenMatch() {
			return minimumTokenMatch;
		}

		@JsonIgnore
		@Override
		public boolean equals(Object options) {
			if (options instanceof Options) {
				return minimumTokenMatch == ((Options) options).getMinimumTokenMatch();
			} else {
				return false;
			}

		}

		@JsonIgnore
		@Override
		public String toString() {
			 return "Options [minimumTokenMatch=" + minimumTokenMatch+ "]";
		}
	}
	
	@JsonDeserialize(as = ExpectedResult.class)
	public record ExpectedResult(@JsonProperty("minimal_similarity")float resultSimilarityMinimum,
	@JsonProperty("maximum_similarity") float resultSimilarityMaximum,
	@JsonProperty("matched_token_number") int resultMatchedTokenNumber,
	@JsonIgnore JPlagComparison jplagComparison)
	{
		public ExpectedResult(float resultSimilarityMinimum, float resultSimilarityMaximum,
				int resultMatchedTokenNumber , JPlagComparison jplagComparison)
		{
			this.resultSimilarityMinimum = resultSimilarityMaximum;
			this.resultSimilarityMaximum = resultSimilarityMaximum;
			this.resultMatchedTokenNumber = resultMatchedTokenNumber;
			this.jplagComparison = jplagComparison;
		}
		
		public ExpectedResult(JPlagComparison jplagComparison)
		{
			this(jplagComparison.minimalSimilarity() , jplagComparison.maximalSimilarity() , jplagComparison.getNumberOfMatchedTokens());
		}
		
		public ExpectedResult(float resultSimilarityMinimum, float resultSimilarityMaximum,
				int resultMatchedTokenNumber) {
			this(resultSimilarityMinimum, resultSimilarityMaximum, resultMatchedTokenNumber, null);
		}
		
		@JsonIgnore
		public float getResultSimilarityMinimum() {
			return resultSimilarityMinimum;
		}

		@JsonIgnore
		public float getResultSimilarityMaximum() {
			return resultSimilarityMaximum;
		}

		@JsonIgnore
		public int getResultMatchedTokenNumber() {
			return resultMatchedTokenNumber;
		}
	}
```

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JavaEndToEndTest.java
SuyDesignz: was removed

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JavaEndToEndTest.java
```diff
@@ -50,48 +59,82 @@ public void teardown() throws IOException {
...
+        // test cases calculated with (n over k)
+        var fileNames = jplagTestSuiteHelper.getAllTestFileNames();
+        ArrayList<String[]> testCases = new ArrayList<>();
+        int outerCounter = 1;
+        for (String fileName : fileNames) {
+            for (int counter = outerCounter; counter < fileNames.length; counter++) {
+                testCases.add(new String[] {fileName, fileNames[counter]});
+            }
+            outerCounter++;
+        }
```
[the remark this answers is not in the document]
SuyDesignz: was changed and outsourced to JPlagTestSuiteHelper.java

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/model/ExpectedResult.java
```diff
@@ -0,0 +1,74 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import de.jplag.JPlagComparison;
+
+/**
+ * The ExpectedResult contains all information and comparison values that are important for a test and that could be
+ * parsed from the json file.
+ */
+public class ExpectedResult {
```
dfuchss, line 12: This may be a record

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/model/Options.java
```diff
@@ -0,0 +1,70 @@
+package de.jplag.end_to_end_testing.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Class manages the test options and thus allows testing of different JPlag options for the test cases.
+ */
+public class Options {
```
dfuchss, line 9: This can be a record

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JPlagEndToEndTestingSuiteTest.java
```diff
@@ -0,0 +1,183 @@
...
+        options.add(new Options(1));
+        options.add(new Options(15));
+    }
+
+    @AfterAll
+    public static void tearDown() throws IOException {
+        for (var test : temporaryResultList.entrySet()) {
+            JsonHelper.writeJsonModelsToJsonFile(test.getValue(), test.getKey());
+        }
+        var test = "";
```
dfuchss, line 66: Delete debugging line :)

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JPlagEndToEndTestingSuiteTest.java
```diff
@@ -0,0 +1,183 @@
...
+     * @return dynamic test cases across all test data and languages
+     * @throws IOException is thrown for all problems that may occur while parsing the json file. This includes both reading
+     */
+    @TestFactory
+    Collection<DynamicTest> dynamicOverAllTest() throws IOException {
+        for (Entry<LanguageOption, Map<String, Path>> languageMap : LanguageToTestCaseMapper.entrySet()) {
+            LanguageOption currentLanguageOption = languageMap.getKey();
+            for (Entry<String, Path> languagePaths : languageMap.getValue().entrySet()) {
+                String[] fileNames = FileHelper.loadAllTestFileNames(languagePaths.getValue());
+                var testCases = JPlagTestSuiteHelper.getPermutation(fileNames, languagePaths.getValue());
```
dfuchss, line 80: maybe getTestCases()

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JPlagEndToEndTestingSuiteTest.java
```diff
@@ -0,0 +1,183 @@
...
+                        Optional<ResultDescription> currentResultDescription = tempResult.stream().filter(x -> x.getOptions().equals(option))
+                                .findFirst();
+
+                        testCollection.add(DynamicTest.dynamicTest(getTestCaseDisplayName(option, currentLanguageOption, testCase), () -> {
+                            try {
+                                ResultDescription currentResult = currentResultDescription.isPresent() ? currentResultDescription.get() : null;
+                                runJPlagTestSuite(directoryName, option, currentLanguageOption, testCase, currentResult);
+                            } finally {
+                                JPlagTestSuiteHelper.clear();
+                            }
```
dfuchss, line 95: Put that code to a method

dfuchss (commented):
Some minor comments

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/model/ExpectedResult.java
dfuchss: Jackson JSON should be able to use records.

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/model/Options.java
dfuchss: Jackson JSON should be able to use records.

tsaglam:
> I tried to change the classes of the json objects to records. Here I ran into some problems with the deserialization. Apparently there are already some entries about this:

AFAIK this should be possible as we do it for the report generation (e.g. [here](https://github.com/jplag/JPlag/blob/ac36d3d11b68a66f5f368f54b213e35dcd31cd34/jplag/src/main/java/de/jplag/reporting/reportobject/model/Match.java)). I just briefly glanced at the issues you linked, and there are workarounds mentioned in there by some commenters.  So either check the report generation or have a look at the workarounds.

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JPlagEndToEndTestingSuiteTest.java
SuyDesignz: removed :sweat_smile:

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JavaEndToEndTest.java
SuyDesignz: done
looks like this:
```
org.opentest4j.AssertionFailedError: 
There were <3> validation error(s):
minimalSimilarity was 99.0 but expected 100.0
maximalSimilarity was 99.0 but expected 100.0
numberOfMatchedTokens was 54 but expected 56
 ==> expected: <true> but was: <false>
```

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JPlagEndToEndTestingSuiteTest.java
SuyDesignz: done

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JPlagEndToEndTestingSuiteTest.java
SuyDesignz: done

SuyDesignz:
If I change the models into records i receive this error-message while deserializing:

```
com.fasterxml.jackson.databind.JsonMappingException: Can not set final int field de.jplag.end_to_end_testing.modelRecord.ResultDescription$Options.minimumTokenMatch to java.lang.Integer (through reference chain:
```

with this record:

``` java 
@JsonDeserialize(as = Options.class)
	public record Options(@JsonProperty("minimum_token_match") int minimumTokenMatch) {
		public Options(int minimumTokenMatch) {
			this.minimumTokenMatch = minimumTokenMatch;
		}

		@JsonIgnore
		public int getMinimumTokenMatch() {
			return minimumTokenMatch;
		}

		@JsonIgnore
		@Override
		public boolean equals(Object options) {
			if (options instanceof Options) {
				return minimumTokenMatch == ((Options) options).getMinimumTokenMatch();
			} else {
				return false;
			}

		}

		@JsonIgnore
		@Override
		public String toString() {
			return "Options [minimumTokenMatch=" + minimumTokenMatch + "]";
		}
	}
```

when I change the type to integer I get the same error message
```
 Can not set final java.lang.Integer field de.jplag.end_to_end_testing.modelRecord.ResultDescription$Options.minimumTokenMatch to java.lang.Integer 
```

SuyDesignz:
Maybe I need to follow this guide:

[Mapping Nested Values with Jackson](https://www.baeldung.com/jackson-nested-values)

dfuchss:
> If I change the models into records i receive this error-message while deserializing:
> 
> ```
> com.fasterxml.jackson.databind.JsonMappingException: Can not set final java.lang.int field de.jplag.end_to_end_testing.modelRecord.ResultDescription$Options.minimumTokenMatch to java.lang.Integer (through reference chain: java.lang.Object[][0]->de.jplag.end_to_end_testing.modelRecord.ResultDescription["options"])
> ```
> 
> with this record:
> 
> ```java
> @JsonDeserialize(as = Options.class)
> 	public record Options(@JsonProperty("minimum_token_match") int minimumTokenMatch) {
> 		public Options(int minimumTokenMatch) {
> 			this.minimumTokenMatch = minimumTokenMatch;
> 		}
> 
> 		@JsonIgnore
> 		public int getMinimumTokenMatch() {
> 			return minimumTokenMatch;
> 		}
> 
> 		@JsonIgnore
> 		@Override
> 		public boolean equals(Object options) {
> 			if (options instanceof Options) {
> 				return minimumTokenMatch == ((Options) options).getMinimumTokenMatch();
> 			} else {
> 				return false;
> 			}
> 
> 		}
> 
> 		@JsonIgnore
> 		@Override
> 		public String toString() {
> 			return "Options [minimumTokenMatch=" + minimumTokenMatch + "]";
> 		}
> 	}
> ```
> 
> when I change the type to integer I get the same error message
> 
> ```
> com.fasterxml.jackson.databind.JsonMappingException: Can not set final java.lang.Integer field de.jplag.end_to_end_testing.modelRecord.ResultDescription$Options.minimumTokenMatch to java.lang.Integer (through reference chain: java.lang.Object[][0]->de.jplag.end_to_end_testing.modelRecord.ResultDescription["options"])
> ```

Try to delete the constructor, and all methods (as records provide them by default), and also delete `@JsonDeserialize(as = Options.class)`

SuyDesignz:
Removing the constructor worked for me! 

``` java
public record ResultDescription(@JsonIgnore LanguageOption languageOption, @JsonProperty("options") Options options,
        @JsonProperty("tests") Map<String, ExpectedResult> identifierToResultMap) {

    @JsonIgnore
    public ExpectedResult getExpectedResultByIdentifier(String identifier) {
        return identifierToResultMap.get(identifier);
    }

    public void putIdenfifierToResultMap(String identifier, ExpectedResult expectedResult) {
        identifierToResultMap.put(identifier, expectedResult);
    }
}
```

``` java
public record Options(@JsonProperty("minimum_token_match") Integer minimumTokenMatch) {

    /**
     * Compares like inside values with the passed object. is necessary to find the correct results in the deserialized json
     * file.
     */
    @JsonIgnore
    @Override
    public boolean equals(Object options) {
        if (options instanceof Options optionsCaseted) {
            return minimumTokenMatch == optionsCaseted.minimumTokenMatch();
        } else {
            return false;
        }
    }

    /**
     * Creates the hashCode for the AKtuelle Options object
     */
    @JsonIgnore
    @Override
    public int hashCode() {
        final int prime = 31;
        int result = 1;
        result = prime * result + minimumTokenMatch;
        return result;
    }

    @JsonIgnore
    @Override
    public String toString() {
        return "Options [minimumTokenMatch=" + minimumTokenMatch + "]";
    }
}
```

``` java 
public record ExpectedResult(@JsonProperty("minimal_similarity") float resultSimilarityMinimum,
        @JsonProperty("maximum_similarity") float resultSimilarityMaximum, @JsonProperty("matched_token_number") int resultMatchedTokenNumber) {
}
```

tsaglam (changes requested):
Please also consider the [Sonar code smells](https://sonarcloud.io/project/issues?id=jplag_JPlag&pullRequest=583&resolved=false&types=CODE_SMELL)!

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/Options.java
```diff
@@ -0,0 +1,39 @@
...
+    public boolean equals(Object options) {
+        if (options instanceof Options optionsCaseted) {
+            return minimumTokenMatch == optionsCaseted.minimumTokenMatch();
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * Creates the hashCode for the AKtuelle Options object
```
tsaglam, line 23: The what?

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/Options.java
```diff
@@ -0,0 +1,39 @@
...
+     */
+    @JsonIgnore
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + minimumTokenMatch;
+        return result;
```
tsaglam, line 31: Why not use `Objects.hash()`?

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/ExpectedResult.java
```diff
@@ -0,0 +1,7 @@
+package de.jplag.end_to_end_testing.modelRecord;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record ExpectedResult(@JsonProperty("minimal_similarity") float resultSimilarityMinimum,
```
tsaglam, line 5: Records are still public types and thus need JavaDoc!

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/ResultDescription.java
```diff
@@ -1,129 +1,22 @@
...
+package de.jplag.end_to_end_testing.modelRecord;
+
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import de.jplag.options.LanguageOption;
+
+public record ResultDescription(@JsonIgnore LanguageOption languageOption, @JsonProperty("options") Options options,
```
tsaglam, line 10: Same here, as for all types!

tsaglam (changes requested):

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/Options.java
dfuchss: Delete hashcode. It's not needed for records :)

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/Options.java
dfuchss: Delete hashcode. It's not needed for records :)

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/Options.java
dfuchss: Also delete equals. It's created in a record implicitly

SuyDesignz:
TODO:

- [ ] add comments and documentation for public classes/records and functions
- [ ] create the new README file for the new sequence of the endToEnd test suite
- [ ] remove as many code smells as possible 
- [ ] review comments and documentation

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/Options.java
dfuchss: And toString as well . This should also be generated implicitly

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/Options.java
SuyDesignz: Was flagged as a bug by the sonarcloud because I overwrote equals() method

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/model/ExpectedResult.java
SuyDesignz: class was replaced by record

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/model/Options.java
SuyDesignz: class was replaced by record

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
SuyDesignz: was removed

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/Options.java
SuyDesignz: removed the above mentioned functions

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/ExpectedResult.java
SuyDesignz: added comments

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/modelRecord/ResultDescription.java
SuyDesignz: added comments

SuyDesignz:
Revised version is now available. Changes can be taken from the README.md as well as the documentation how to create new endToEnd tests for each language. 
the following has been changed:
1. json format
``` json 

  "options" : {
    "minimum_token_match" : 1
  },
  "tests" : {
    "SortAlgo-SortAlgo5" : {
      "minimal_similarity" : 82.14286,
      "maximum_similarity" : 82.14286,
      "matched_token_number" : 46
    },
[...]
```
2. Creating the tests runs completely dynamically, which means that even for new languages nothing has to be changed in the source code. Only test data must be made available. Can be found in the README under "Copying Plagiarism To The Resources".
3. json serialization and deserialization now runs via records 
4. it is now possible to run the tests in different JPlag test settings.  [JPlag - option variants for the endToEnd tests #590](https://github.com/jplag/JPlag/issues/590)
5. The comparisons in the tests themselves have been adjusted to provide more accurate information about the incorrect values. 

```
There were <3> validation error(s):
minimalSimilarity was 82.1 but expected 82.14286
maximalSimilarity was 82.1 but expected 82.14286
numberOfMatchedTokens was 41 but expected 46
 ==> expected: <true> but was: <false>
```

For all further information, the README should be a good source of information. Here everything concerning the basics of the TestSuite should be available and documented.

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JPlagEndToEndTestingSuiteTest.java
```diff
@@ -0,0 +1,258 @@
...
+import de.jplag.exceptions.ExitException;
+import de.jplag.options.JPlagOptions;
+import de.jplag.options.LanguageOption;
+
+/**
+ * Main test class for end-to-end testing in all language. The test cases aim to detect changes in the detection of
+ * plagiarism in the Java language and to be able to roughly categorize them. The plagiarism is compared with the
+ * original class. The results are compared with the results from previous tests and changes are detected.
+ */
+public class JPlagEndToEndTestingSuiteTest {
```
tsaglam, line 39: Convoluted name, rename it to `EndToEndTestSuite` or similar.

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
```diff
@@ -1,120 +1,90 @@
...
-import de.jplag.end_to_end_testing.model.JsonModel;
-import de.jplag.end_to_end_testing.model.TestCaseModel;
 import de.jplag.options.LanguageOption;
 
-/**
- * This helper class deals with creating the test cases as well as copying and deleting for the end-to-end tests. The
- * required plagiarisms are copied from the resource folder to a temporary location, thus creating a folder structure
- * that can be tested by JPlag. Models are instantiated here and required information for the tests is loaded.
- */
 public class JPlagTestSuiteHelper {
```
tsaglam, line 16: I would remove the JPlag prefix in the class name.

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
```diff
@@ -1,120 +1,90 @@
...
+     * @throws IOException
      */
-    public TestCaseModel createNewTestCase(String[] classNames, String functionName) throws IOException {
-        createNewTestCaseDirectory(classNames);
-        JsonModel resultJsonModel = resultModel.stream().filter(jsonModel -> functionName.equals(jsonModel.getFunctionName())).findAny().orElse(null);
-        return new TestCaseModel(TestDirectoryConstants.TEMPORARY_SUBMISSION_DIRECTORY_NAME, resultJsonModel, languageOption);
+    public static Map<LanguageOption, Map<String, Path>> getAllLanguageResources() throws IOException {
+        String[] languageDirectoryNames = FileHelper.getAllDirectoriesInPath(TestDirectoryConstants.BASE_PATH_TO_LANGUAGE_RESOURCES);
+        List<LanguageOption> languageInPathList = FileHelper.getLanguageOptionsFromPath(languageDirectoryNames);
+        var returnMap = new HashMap<LanguageOption, Map<String, Path>>();
```
tsaglam, line 34: Use enum map: https://sonarcloud.io/project/issues?pullRequest=583&issues=AYLDFWS4IezL5K6xxgcC&open=AYLDFWS4IezL5K6xxgcC&id=jplag_JPlag

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
```diff
@@ -1,120 +1,90 @@
...
+     * Loads all existing test data into a test structure. the complex structure consists of the LanguageOption and their
+     * data to be tested. These are divided into folders. for more information please read the README file
+     * @return mapped LanguageOption to the data under test
+     * @throws IOException
      */
-    public TestCaseModel createNewTestCase(String[] classNames, String functionName) throws IOException {
-        createNewTestCaseDirectory(classNames);
-        JsonModel resultJsonModel = resultModel.stream().filter(jsonModel -> functionName.equals(jsonModel.getFunctionName())).findAny().orElse(null);
-        return new TestCaseModel(TestDirectoryConstants.TEMPORARY_SUBMISSION_DIRECTORY_NAME, resultJsonModel, languageOption);
+    public static Map<LanguageOption, Map<String, Path>> getAllLanguageResources() throws IOException {
```
tsaglam, line 31: Remove throws declarations, as it is no longer required.

tsaglam (changes requested):
Besides the small changes below, I have some issues with the module/package naming.
However, this should be fixed in a separate (!!) pull request, as it would make the change overview of this one unreadable. I will make a small issue for this, and you should resolve it after this PR is merged.

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JPlagEndToEndTestingSuiteTest.java
SuyDesignz: Without the test at the end, the classes don't run in the pipeline, do they?

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
SuyDesignz: done

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
SuyDesignz: done

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/JPlagTestSuiteHelper.java
SuyDesignz: done

jplag.endToEndTesting/src/test/java/de/jplag/end_to_end_testing/JPlagEndToEndTestingSuiteTest.java
tsaglam: Good point, I am not sure. You can test that with a local maven run. If it is the case, maybe we just go for `EndToEndSuiteTest` for now.

SuyDesignz: maybe `EndToEndSuiteTest` ?

SuyDesignz: without the "test" at the end the tests don't run locally either. so i will rename it to` EndToEndSuiteTest`

SuyDesignz: done

tsaglam:
@SuyDesignz when I run `mvn clean package assembly:single` locally with your branch, it produces junk files in the `jplag/jplag` directory. This is probably unintended, right?
grafik

SuyDesignz:
This shouldn't really happen, but it only happened with the current master's degree. I still do not know what the reason is.

SuyDesignz:
This should not actually happen, but has also only come with the current master branch. I still do not know what the reason is.

SuyDesignz:
if you look in the A-C.json there are also completely different names as I use for the tests
``` json
{"file1":"GSTiling.java","file2":"GSTiling.java","start1":6,"end1":247,
```

tsaglam:
You are right, this is currently also happening on the master branch. This means it is not your responsibility!

tsaglam (approved):

SuyDesignz:
@jplag/studdev can someone look over the PR so that this can be completed?

tsaglam:
> @jplag/studdev can someone look over the PR so that this can be completed?

@robinmaisch maybe you can review this PR and @SuyDesignz reviews your PR 594?

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/TestSuiteHelper.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public static String getTestIdentifier(JPlagComparison jPlagComparison) {
+
+        String testFileNamesInFirstSubmission = FileHelper.getEnclosedFileNamesFromCollection(jPlagComparison.getFirstSubmission().getFiles());
+        String testFileNamesInSecondSubmission = FileHelper.getEnclosedFileNamesFromCollection(jPlagComparison.getSecondSubmission().getFiles());
+
+        int compare = testFileNamesInFirstSubmission.compareTo(testFileNamesInSecondSubmission);
+
+        return compare < 0 ? testFileNamesInFirstSubmission + "-" + testFileNamesInSecondSubmission
+                : testFileNamesInSecondSubmission + "-" + testFileNamesInFirstSubmission;
```
robinmaisch, line 61:
```suggestion
	return List.of(jPlagComparison.getFirstSubmission(), jPlagComparison.getSecondSubmission()).stream()
		.map(Submission::getFiles)
		.map(FileHelper::getEnclosedFileNamesFromCollection)
		.sorted()
		.collect(Collectors.joining("-"));
		
```

robinmaisch (commented):
I admit that I don't get the bigger picture from just this diff, but I tried to refactor the code to be more readable or more "common practice" (based on what I perceive as such), which more often than not was accomplished by working with Collections/Streams. I prefer them over Arrays any day, but again, that's just me.

I typed the code into this website untested, so there might be some parentheses missing or something like that. Hopefully, my suggestions are still helpful. Good job!

tsaglam:
@SuyDesignz, please finish this PR today so it is ready to merge.

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/FileHelper.java
```diff
@@ -0,0 +1,149 @@
...
+            Files.delete(folder.toPath());
+        }
+    }
+
+    /**
+     * Creates directory if it dose not exist
+     * @param directory to be created
+     * @throws IOException if the directory could not be created
+     */
+    public static void createDirectoryIfItDoseNotExist(File directory) throws IOException {
```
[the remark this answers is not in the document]
SuyDesignz: done

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/model/ResultDescription.java
```diff
@@ -0,0 +1,36 @@
...
+        return identifierToResultMap.get(identifier);
+    }
+
+    /**
+     * Adds expected results to the existing list of the object. These results are stored in a map with the given
+     * identifier.
+     * @param identifier under which the results should be stored
+     * @param expectedResult expected results belonging to the identifier
+     */
+    public void putIdenfifierToResultMap(String identifier, ExpectedResult expectedResult) {
```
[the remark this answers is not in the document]
SuyDesignz: done

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/FileHelper.java
```diff
@@ -0,0 +1,149 @@
...
+     */
+    public static String[] createNewTestCaseDirectory(String[] classNames) throws IOException {
+        // Copy the resources data to the temporary path
+        String[] returnSubmissionPath = new String[classNames.length];
+        for (int counter = 0; counter < classNames.length; counter++) {
+            Path originalPath = Path.of(classNames[counter]);
+            returnSubmissionPath[counter] = Path
+                    .of(TestDirectoryConstants.TEMPORARY_SUBMISSION_DIRECTORY_NAME.toString(), "submission" + (counter + 1)).toAbsolutePath()
+                    .toString();
+            Path copiePath = Path.of(TestDirectoryConstants.TEMPORARY_SUBMISSION_DIRECTORY_NAME.toString(), "submission" + (counter + 1),
```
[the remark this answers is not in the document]
SuyDesignz: done

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/FileHelper.java
```diff
@@ -0,0 +1,149 @@
...
+     * Load all possible language is resource path
+     * @param directoryNames folder names for which the language options should be listed.
+     * @return list of all LanguageOptions included in the resource path
+     */
+    public static List<LanguageOption> getLanguageOptionsFromPath(String[] directoryNames) {
+        List<LanguageOption> returnList = new ArrayList<>();
+        for (String languageDirectoryName : directoryNames) {
+            returnList.add(LanguageOption.valueOf(languageDirectoryName));
+        }
+        return returnList;
```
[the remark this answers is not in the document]
SuyDesignz: That's really helpful thanks! actually it shouldn't happen but you never know. especially when names change in the languages it could lead to what you described.

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/FileHelper.java
```diff
@@ -0,0 +1,149 @@
...
+     * @return merged filenames
+     */
+    public static String getEnclosedFileNamesFromCollection(Collection<File> files) {
+        StringBuilder stringBuilder = new StringBuilder(files.size());
+        for (File file : files) {
+            String fileName = file.getName();
+            stringBuilder.append(fileName.substring(0, fileName.lastIndexOf('.')));
+        }
+
+        return stringBuilder.toString();
```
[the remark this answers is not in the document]
SuyDesignz: replaced with:

``` java
return files.stream().map(File::getName).map(fileName -> fileName.substring(0, fileName.indexOf('.'))).collect(Collectors.joining());
```
thanks for the suggestion! my java knowledge is not the best so the code examples help a lot!

SuyDesignz:
I currently have problems to implement the new structure with the languages in the endToEnd tests. Since the languages are actually loaded automatically and this has now changed, I must first look at how I can adapt this now.

SuyDesignz:
Mapper and all other functions have to be adapted now.

SuyDesignz:
Done with the adjustments and suggested changes.

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/FileHelper.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Merges all contained filenames together without extension
+     * @param files whose names are to be merged
+     * @return merged filenames
+     */
+    public static String getEnclosedFileNamesFromCollection(Collection<File> files) {
+
+        return files.stream().map(File::getName).map(fileName -> fileName.substring(0, fileName.indexOf('.'))).collect(Collectors.joining());
```
dfuchss, line 32:
```suggestion
        return files.stream().map(File::getName).map(fileName -> fileName.substring(0, fileName.lastIndexOf('.'))).collect(Collectors.joining());
```

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/TestSuiteHelper.java
```diff
@@ -0,0 +1,90 @@
...
+        int outerCounter = 1;
+        for (String fileName : fileNames) {
+            for (int counter = outerCounter; counter < fileNames.length; counter++) {
+                testCases.add(new String[] {Path.of(path.toAbsolutePath().toString(), fileName).toString(),
+                        Path.of(path.toAbsolutePath().toString(), fileNames[counter]).toString()});
+            }
+            outerCounter++;
+        }
+        return testCases;
+    }
```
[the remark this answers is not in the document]
dfuchss: I agree here. Please change that.

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/model/ResultDescription.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public record ResultDescription(@JsonIgnore String languageIdentifier, @JsonProperty("options") Options options,
+        @JsonProperty("tests") Map<String, ExpectedResult> identifierToResultMap) {
+
+    @JsonIgnore
+    /**
+     * @param identifier for which the stored results are needed
+     * @return stored results as ExpectedResult object for the passed id
+     */
+    public ExpectedResult getExpectedResultByIdentifier(String identifier) {
```
dfuchss, line 21:
```suggestion
    /**
     * @param identifier for which the stored results are needed
     * @return stored results as ExpectedResult object for the passed id
     */
    @JsonIgnore
    public ExpectedResult getExpectedResultByIdentifier(String identifier) {
```

dfuchss (changes requested):
Some comments

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/model/ResultDescription.java
SuyDesignz: done

jplag.endToEndTesting/src/main/java/de/jplag/end_to_end_testing/helper/FileHelper.java
SuyDesignz: done

SuyDesignz:
I'm not sure right now what I should change in the function?

dfuchss:
> I'm not sure right now what I should change in the function?

What function?

SuyDesignz:
That I should change something about the function "getTestCases" in `/helper/TestSuiteHelper.java`
``` java 
 public static List<String[]> getTestCases(String[] fileNames, Path path) {
```

https://github.com/jplag/JPlag/pull/583#discussion_r955735898

dfuchss:
> That I should change something about the function "getTestCases" in `/helper/TestSuiteHelper.java`
> 
> ```java
>  public static List<String[]> getTestCases(String[] fileNames, Path path) {
> ```
> 
> [#583 (comment)](https://github.com/jplag/JPlag/pull/583#discussion_r955735898)

It's cleaner to use two for(int i = 0,... ) loops here; instead of using a enhanced for loop with index counter outside.

dfuchss (approved):

dfuchss:
@SuyDesignz did you resolved all open conversations? If so, just drop a comment here :)

SuyDesignz:
Yes I did :) not sure why I can't close https://github.com/jplag/JPlag/pull/583#discussion_r955735898 but the other comments have been edited

PR #585, opened 2022-08-10
Add Rust frontend

robinmaisch:
This PR introduces a Rust frontend to JPlag. Its design is similar to the other frontends based on ANTLR4.
Some challenges regarding the syntax of Rust and its formulation in the ANTLR4 grammar are discussed in the README.
To the best of my understanding, it now produces consistent, correct TokenLists. Still, this frontend may very well be subject to further redesign in the future.

Adressing #345 and completing the set of frontends proposed there.


jplag.frontend.rust/src/main/java/de/jplag/rust/JplagRustListener.java
```diff
@@ -0,0 +1,898 @@
...
+     * Implementation of Context for the Rust language
+     */
+    enum RustContext implements ParserState.Context {
+        /**
+         * This is used to make sure that the stack is not empty -> getCurrent() != null
+         **/
+        FILE(NONE, NONE),
+
+        /**
+         * These contexts are used to assign the correct tokens to '{' and '}' terminals.
```
tsaglam, line 826: Is that not possible via the exit methods?

tsaglam (changes requested):
Just a few issues and questions...

jplag.frontend.rust/src/test/java/de/jplag/rust/RustFrontendTest.java
```diff
@@ -0,0 +1,148 @@
...
+            if (coverage == 1) {
+                logger.info("All lines covered.");
+            } else {
+                logger.info("Coverage: %.1f%%.".formatted(coverage * 100));
+                logger.info("Missing lines {}", codeLines);
+                if (coverage - 0.9 <= EPSILON) {
+                    // TODO use fail() instead when frontend is ready
+                    logger.error("Source coverage is unsatisfactory");
+                }
+            }
```
JanWittler, line 94: If in the future something about the frontend changes, it will be almost impossible to detect what changed and why the coverage is now reduced. Thus, I would suggest to explicitly document the lines of the code that are not covered. One option might be to indicate block-wise uncovered sections with surrounding comment (like `//TOKENIZATION IGNORED SECTION START` / `//TOKENIZATION IGNORED SECTION END`) or for single lines a trailing comment.

jplag.frontend.rust/src/main/java/de/jplag/rust/RustTokenConstants.java
```diff
@@ -0,0 +1,118 @@
+package de.jplag.rust;
+
+import de.jplag.TokenConstants;
+
+public interface RustTokenConstants extends TokenConstants {
+    int NONE = -1;
```
JanWittler, line 6: This will cause problems with the core algorithm. Please only use >= 0 values. Why do you need that one?

jplag.frontend.rust/src/main/java/de/jplag/rust/ParserState.java
```diff
@@ -0,0 +1,74 @@
...
+    protected void enter(C context) {
+        blockContexts.push(context);
+    }
+
+    /**
+     * Leaves the current context, making sure that it is one of the given ones.
+     * @param contexts The contexts to expect to end here
+     */
+    @SafeVarargs
+    protected final void leave(C... contexts) {
```
JanWittler, line 34:
```suggestion
    protected final void leaveAsserted(C... contexts) {
```

jplag.frontend.rust/src/main/java/de/jplag/rust/ParserState.java
```diff
@@ -0,0 +1,74 @@
...
+        C topContext = blockContexts.pop();
+        assert Arrays.stream(contexts).anyMatch(context -> context == topContext);
+    }
+
+    /**
+     * Returns the current context.
+     * @return the current context
+     */
+    protected C getCurrent() {
```
JanWittler, line 43:
```suggestion
    protected C getCurrentContext() {
```

jplag.frontend.rust/src/main/java/de/jplag/rust/ParserState.java
```diff
@@ -0,0 +1,74 @@
...
+        return blockContexts.peek();
+    }
+
+    /**
+     * Leaves the current context if it is the given one.
+     * @param blockContext the context that may be expected to end here
+     */
+    protected void maybeLeave(C blockContext) {
```
JanWittler, line 51:
```suggestion
    protected void leaveIfInContext(C blockContext) {
```

JanWittler (commented):
General remarks without checking the actual tokenization

jplag.frontend.rust/src/main/java/de/jplag/rust/JplagRustListener.java
```diff
@@ -0,0 +1,898 @@
...
+        transformToken(STRUCT, context.getStart());
+        state.enter(RustContext.STRUCT_BODY);
+        super.enterStruct_(context);
+    }
+
+    @Override
+    public void exitStruct_(RustParser.Struct_Context context) {
+        state.leave(RustContext.STRUCT_BODY);
+        super.exitStruct_(context);
+    }
```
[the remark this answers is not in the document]
robinmaisch: It is supposed to make it easier to decide which token type to add when encountering symbols that are used in various different contexts. So, instead of
```java
// handle '{'
if (context.getParent() instanceof IfExpressionContext) transform(IF_BODY_START, ...);
else if (context.getParent() instanceof LoopExpressionContext) transform(LOOP_BODY_START, ...);
// many more cases...
```
the code can be changed to
```java
// handle '{'
int tokenType = state.getCurrent().getStartType();
transform(tokenType, ...);
```
and thus, many different uses of the same symbol can be handled concisely.

This worked very well in other frontends already, in this frontend there are more cases and exceptions to consider, so it took a lot more different states to make it work. I tried to document why each of the states seemed necessary to me.

jplag.frontend.rust/src/main/java/de/jplag/rust/JplagRustListener.java
robinmaisch: In some cases, yes. I'll make an example:
```rust
if (true) { println!("{}", 1); }
```
This statement is handled as follows:
1. Assign `IF_STATEMENT` in the `enter` method of the if statement
2. Traverse the condition and assign tokens as needed
3. Now, the `IF_BODY_START` token shall be added at the position of `{`. This can easily be done when the terminal `{` itself is traversed; another way would be to try to find its position in the `enter` method, but then the order of the tokens might turn out wrong.
4. Traverse the body and assign tokens as needed
5. Add the `IF_BODY_END` token. This could be done in the `exit` method of the if statement indeed.

In Rust, there are many instances of optional bodies (for `struct`s, for example) and cases where any type of bracket is permitted (e.g. macro calls: `()`, `[]`, `{}`), so it seemed easier to wait for the terminal to actually occur instead of trying to find out whether/where/which one occurs in the `enter` method.

jplag.frontend.rust/src/main/java/de/jplag/rust/RustTokenConstants.java
robinmaisch: Oh, my bad. I should say that no token with this type is ever created, instead, it is a flag used in ParserState.Context objects to indicate that no token should be added for `{`or `}` inside of this context.

I'll move it to the ParserState.Context interface where it is actually used and causes less confusion.

jplag.frontend.rust/src/main/java/de/jplag/rust/ParserState.java
robinmaisch: That's better, thanks!

robinmaisch: That's better, thanks!

jplag.frontend.rust/src/test/java/de/jplag/rust/RustFrontendTest.java
robinmaisch: The files used for the test are very long, so it would be hard to annotate it that way. 
Maybe I can add the output of the TokenPrinter for the test files somewhere so a diff analysis tool can reveal what changed.
Would that be a fair option? -- If yes, I need a hint where to put those output files.

jplag.frontend.rust/src/main/java/de/jplag/rust/JplagRustListener.java
```diff
@@ -0,0 +1,898 @@
...
+
+import org.antlr.v4.runtime.ParserRuleContext;
+import org.antlr.v4.runtime.RuleContext;
+import org.antlr.v4.runtime.Token;
+import org.antlr.v4.runtime.tree.*;
+
+import de.jplag.rust.grammar.RustParser;
+import de.jplag.rust.grammar.RustParserBaseListener;
+
+public class JplagRustListener extends RustParserBaseListener implements ParseTreeListener {
```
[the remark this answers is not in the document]
robinmaisch: Done, comes with next push.

jplag.frontend.rust/README.md
```diff
@@ -0,0 +1,84 @@
+# JPlag Scala language frontend
+
+The JPlag Scala frontend allows the use of JPlag with submissions in Scala. <br>
+It is based on the [Rust ANTLR4 grammar](https://github.com/antlr/grammars-v4/tree/master/rust), licensed under MIT.
```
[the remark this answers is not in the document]
robinmaisch: Ha, the text replacement tool did not wrap around. That's not good.

jplag.frontend.rust/src/test/java/de/jplag/rust/RustFrontendTest.java
JanWittler: Ah, I'm sorry. I confused it with the token coverage test where only the `complete.rs` file is relevant. Yes, for the other file it is probably too much.
Your proposed solution sounds like a good compromise. Maybe put it in `resources/de/jplag/rust/annotated`?

jplag.frontend.rust/src/main/java/de/jplag/rust/JplagRustListener.java
tsaglam: So there is only one enter and exit method for bodies, but not a specific one for if/for/... bodies?

robinmaisch: I was beginning to wonder, and looking at the other frontends again, I now remember why this sort of thing was easy before and is now hard.
There are no enter/exit methods for any bodies at all, because the bodies get no syntactical category of their own in the Rust grammar.

```antlr
// in C++ frontend
ifStatement: 
	IF OPEN_PARENS expression CLOSE_PARENS if_body (ELSE if_body)? ;
if_body: 
	block ;

// in Rust frontend
ifExpression:
	'if' expression blockExpression ( 'else' (blockExpression | ifExpression | ifLetExpression) )? ;
```
In the C++ grammar, the if-body gets a grammatical rule of its own, so that it can be easily distinguished from any other type of block.

In the Rust grammar however, the if-body is simply a blockExpression, but a blockExpression can occur in many different contexts. So, the grammar will not help to tell blockExpressions apart.

So, it is really the general formulation of the Rust grammar that does not play in my favor here. Your reply describes the problem perfectly.

tsaglam: Ah, now that makes sense! Thank you for your insights. We may add a sentence or two regarding that in the class JDoc or the frontend readme.

JanWittler: Probably out of scope for this PR, but if this is the entire problem, refactoring the grammar could also be an option. Based on your experience, what would you estimate how difficult this is? I would assume it to not be too hard as it should only introduce an intermediate rule. Or do I oversee something there?

robinmaisch: Yes, that should be entirely possible. I did that for another frontend but did not think to do it this time.

jplag.frontend.rust/src/main/java/de/jplag/rust/grammar/RustLexerBase.java
```diff
@@ -0,0 +1,94 @@
...
+            if (_input.LA(2) == '6' && _input.LA(3) == '4')
+                return true;
+            return false;
+        }
+        if (next >= 'a' && next <= 'z')
+            return false;
+        if (next >= 'A' && next <= 'Z')
+            return false;
+        return true;
+    }
```
dfuchss, line 53: Is this class generated? I'm not sure about the naming of the methods here.

dfuchss (commented):
Only one minor comment at one class

jplag.frontend.rust/src/main/java/de/jplag/rust/grammar/RustLexerBase.java
robinmaisch: It's not generated, but it's taken from the Antlr4 grammar library. I agree about the naming.

tsaglam: I think we shold refactor it, as it really is in bad condition.

tsaglam: @robinmaisch, I think this is the only open issue here before we can merge.

JanWittler: I think we are also waiting for the new `TokenPrinter` such that the stored annotated files can look good.

tsaglam: Good call, I missed that.

robinmaisch: Now it should look a bit better. I cannot quite grasp what floatLiteralPossible and floatDotPossible really do, so I just tried to make them more readable without altering their behaviour.

tsaglam: I think this is the best way to treat this code. From glancing at it, I am not sure either what it does (probably something regarding parsing [differently encoded float values](https://doc.rust-lang.org/std/primitive.f64.html)), but that just shows us how important well-written and well-documented code is.

tsaglam (approved):
Looks good, now this PR only depends on the token printer changes.

robinmaisch:
The code smells are acceptable for my part. Should be ready to merge.

tsaglam (commented):
@robinmaisch one final request: Could you update the newly [added language table](https://github.com/jplag/JPlag/blob/master/README.md#supported-languages) in the project readme? Either here or in a separate PR.

robinmaisch:
> @robinmaisch one final request: Could you update the newly [added language table](https://github.com/jplag/JPlag/blob/master/README.md#supported-languages) in the project readme? Either here or in a separate PR.

Ah, sure! There you go.

PR #588, opened 2022-08-12
Zip Report Viewer Files

nestabentum:
Builds upon PR #569 and should not be merged before it. Only the changes from commit 79813908d9c2f3d2fb2cf2bb8664f64345668b84 are relevant.

Zips the folder containing the report viewer files. Deletes the unzipped version, given zipping was successful.


jplag/src/main/java/de/jplag/reporting/jsonfactory/DirectoryManager.java
```diff
@@ -0,0 +1,88 @@
...
+                }
+
+                @Override
+                public FileVisitResult visitFileFailed(Path file, IOException exc) {
+                    logger.error("Unable to zip " + file, exc);
+                    return FileVisitResult.CONTINUE;
+                }
+            });
+        } catch (IOException e) {
+            return false;
```
dfuchss, line 84:
```suggestion
            logger.error(e.getMessage(), e);
            return false;
```

dfuchss (commented):
Some minor remark

jplag/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -41,26 +41,37 @@ public class ReportObjectFactory {
...
 
+            zipAndDelete(path);
         } catch (IOException e) {
             logger.error("Could not create directory " + path + " for report viewer generation", e);
         }
 
     }
 
+    private void zipAndDelete(String path) {
+        var zipWasSuccessful = zipDirectory(path);
```
sebinside, line 67: another var where the type is not clear

sebinside (approved):
I tested this on windows 10 with a set of **40.000** comparisons, and it worked fine (although it took some time). Hint: Don't try to use the Report Viewer with 40.00 files, it obviously crashes :)

jplag/src/main/java/de/jplag/reporting/jsonfactory/DirectoryManager.java
```diff
@@ -0,0 +1,88 @@
...
+        }
```
JanWittler, line 85: The documentation was a bit incomplete for me, so my suggestions may already be covered by the current implementation.
1) If visiting a file failed, I would say the zipping failed entirely. Otherwise, we lose information. Thus, I would guess we need to return `FileVisitResult.TERMINATE` here. However, I did not find in the documentation whether that triggers an exception or how to otherwise get informed that zipping was terminated.
2) If the file walking fails after already having visited some files, are there already files written on disk or is some /tmp/ intermediate file used? Otherwise, we should (try to) cleanup when catching the `IOException`.

nestabentum: >If visiting a file failed, I would say the zipping failed entirely. Otherwise, we lose information. Thus, I would guess we need to return FileVisitResult.TERMINATE here. However, I did not find in the documentation whether that triggers an exception or how to otherwise get informed that zipping was terminated.

As the report viewer currently does not handle missing files very well, I agree. But once the report viewer is able to recognise that files are missing and indicate it to the user, we could very well use `CONTINUE` instead of `TERMINATE`.
Using `TERMINATE` does not trigger an exception but simply stops the visiting process.

> If the file walking fails after already having visited some files, are there already files written on disk or is some /tmp/ intermediate file used? Otherwise, we should (try to) cleanup when catching the IOException.

I could not locate any hidden files. But an incomplete zip folder remains, so I chose to throw an exception when a visit fails, so that the catch code is triggered and the zip folder is deleted. The unzipped version will remain.

tsaglam:
@nestabentum, some conflicts need to be solved here before we can proceed.

nestabentum:
Clean and ready to merge version is #608

PR #594, opened 2022-08-16
Add more config options to TokenPrinter

robinmaisch:
I'd like to propose two very small additions to the TokenPrinter that made it more helpful in my process of designing frontends. As it has been proposed to upload the TokenPrinter output for some sample files, this is the format that I would suggest for those output files.

1. Print code lines that contain no tokens if `PRINT_EMPTY_LINES` is active -- in case there are some lines where tokens are missing, or simply to get the complete context
2. Put an empty line after the last "token line" of a nonempty line if `SPACIOUS` is active -- this makes it easier on the eyes for me personally.

These flags are active by default in this PR, but the original behaviour can be restored by setting `PRINT_EMPTY_LINES` and `SPACIOUS` to false.

Sample:
```java
22      public final class TokenPrinter {
        |CLASS{

23          private static final Logger logger = LoggerFactory.getLogger(TokenPrinter.class);
            |VARDEF                                                                        |
                                                 |apply

24          // Representation:
25          private static final String BAR = "|";
            |VARDEF                             |

26          private static final String TAB = "\t";
            |VARDEF                              |

27          private static final String SPACE = " ";
            |VARDEF                               |
```
Before:
```java
22      public final class TokenPrinter {
        |CLASS{
23          private static final Logger logger = LoggerFactory.getLogger(TokenPrinter.class);
            |VARDEF                                                                        |
                                                 |apply
25          private static final String BAR = "|";
            |VARDEF                             |
26          private static final String TAB = "\t";
            |VARDEF                              |
27          private static final String SPACE = " ";
            |VARDEF                               |
```


jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
```diff
@@ -104,10 +106,15 @@ public static String printTokens(TokenList tokens, Collection<File> files, File
...
             // New code line:
-            if (token.getLine() > lineIndex) {
-                lineIndex = token.getLine();
+            boolean previousLineEmpty = false;
+            while (token.getLine() > lineIndex) {
+                lineIndex++;
                 columnIndex = 1;
 
+                boolean currentLineEmpty = token.getLine() > lineIndex;
+                if (currentLineEmpty && !PRINT_EMPTY_LINES)
```
tsaglam, line 115: Avoid if conditions without curly braces.

tsaglam (changes requested):
Just one small thing.

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
```diff
@@ -104,10 +106,15 @@ public static String printTokens(TokenList tokens, Collection<File> files, File
             }
 
             // New code line:
-            if (token.getLine() > lineIndex) {
-                lineIndex = token.getLine();
+            boolean previousLineEmpty = false;
+            while (token.getLine() > lineIndex) {
```
JanWittler, line 110: I get that you tried to modify as little as possible with this PR and thus opted for this while loop. Before, the logic was to get the line for a token. However, now you are implicitly iterating over the lines and get the tokens for it, i.e. the logic changed to get all tokens for the current line. Thus, the code should be adapted to iterate over the lines which would also remove the need for a nested loop which is always difficult to read.

To keep the scope of this PR small, we can also make an issue and refactor the token printer in a subsequent PR. Probably @tsaglam should have the decided opinion on that 😄 .

tsaglam: Both fixing it in here or making an issue is fine for me. To be fair, the token printer never was that readable to begin with. However, that's not a reason to leave it as is!

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
```diff
@@ -104,10 +106,15 @@ public static String printTokens(TokenList tokens, Collection<File> files, File
...
+                lineIndex++;
                 columnIndex = 1;
 
+                boolean currentLineEmpty = token.getLine() > lineIndex;
+                if (currentLineEmpty && !PRINT_EMPTY_LINES)
+                    continue;
+
                 String fileName = token.getFile().isEmpty() ? root.getName() : token.getFile();
                 fileName = fileName + suffix.orElse("");
                 if (!filesToLines.containsKey(fileName)) {
```
robinmaisch, line 120: @jplag/maintainer 
I'm having a go at a refactoring, but I'm having trouble understanding why using the root folder as a replacement file name would be valid. Can someone explain?

JanWittler: Don't know neither but if you refactor it anyways, I would opt for doing it properly, i.e.
```Java
for (File file: files) {
  List<String> lines = getLinesForFile(file); //invented method but probably better than loading all lines upfront
  builder.append(System.lineSeparator());
  builder.append(file.getName());
  for (String line: lines) {
    / print lines and tokens
  }
}
```
and then you don't need the fallback anymore.

tsaglam: Whenever your submission folder contains single files instead of folders, the root is the name of the file, and `Token.getFile()` is `""`.

This is a valid folder structure:
```
input/
|- submissionA.java
|- submissionB.java
|- submissionC.java
```

robinmaisch:
This is my attempt at refactoring the TokenPrinter, and some small additions to go with it:

- Extracted everything concerning appending some output and managing the current position in the output to `PrinterOutputBuilder` (name open for debate)
- Extracted all data needed for the output of one line to `LineData`
   -> printTokens method is now much more concise

- Added method `stream` to `TokenList`, as that is used quite often in tests
- Added "filesAsSubmission" test to TokenPrinterTest

jplag.frontend-utils/src/main/java/de/jplag/TokenList.java
```diff
@@ -63,6 +65,14 @@ public Token getToken(int index) {
         return tokens.get(index);
     }
 
+    /**
+     * Returns a sequential Stream of the tokens.
+     * @return a stream of the tokens
+     */
+    public final Stream<Token> stream() {
```
tsaglam, line 72: I would rather change the return type of `allTokens()` to Collection rather than add a stream method. Collections can be directly streamed.

jplag.frontend.golang/src/test/java/de/jplag/golang/GoFrontendTest.java
```diff
@@ -158,7 +157,7 @@ private List<Integer> getCodeLines(List<String> lines) {
      * @param fileName The file name of the complete code example
      */
     private void testTokenCoverage(TokenList tokens, String fileName) {
-        var foundTokens = StreamSupport.stream(tokens.allTokens().spliterator(), true).mapToInt(Token::getType).sorted().distinct().boxed().toList();
+        var foundTokens = tokens.stream().parallel().mapToInt(Token::getType).sorted().distinct().boxed().toList();
```
tsaglam, line 160: Why parallel? Is there really a performance issue here?

jplag.frontend.kotlin/src/test/java/de/jplag/kotlin/KotlinFrontendTest.java
```diff
@@ -154,7 +153,7 @@ private int[] getCodeLines(List<String> lines) {
      * @param fileName The file name of the complete code example
      */
     private void testTokenCoverage(TokenList tokens, String fileName) {
-        var foundTokens = StreamSupport.stream(tokens.allTokens().spliterator(), true).mapToInt(Token::getType).sorted().distinct().toArray();
+        var foundTokens = tokens.stream().parallel().mapToInt(Token::getType).sorted().distinct().toArray();
```
tsaglam, line 156: Also here.

jplag.frontend.rlang/src/test/java/de/jplag/rlang/RFrontendTest.java
```diff
@@ -92,7 +92,7 @@ private void testSourceCoverage(String fileName, TokenList tokens) {
      * @param fileName The file name of the complete code example
      */
     private void testTokenCoverage(TokenList tokens, String fileName) {
-        var foundTokens = StreamSupport.stream(tokens.allTokens().spliterator(), true).mapToInt(Token::getType).sorted().distinct().toArray();
+        var foundTokens = tokens.stream().parallel().mapToInt(Token::getType).sorted().distinct().toArray();
```
tsaglam, line 95: Same here.

jplag.frontend.scala/src/test/java/de/jplag/scala/ScalaFrontendTest.java
```diff
@@ -162,7 +161,7 @@ private List<Integer> getCodeLines(List<String> lines) {
      * @param fileName The file name of the complete code example
      */
     private void testTokenCoverage(TokenList tokens, String fileName) {
-        var foundTokens = StreamSupport.stream(tokens.allTokens().spliterator(), true).mapToInt(Token::getType).distinct().boxed().toList();
+        var foundTokens = tokens.stream().parallel().mapToInt(Token::getType).distinct().boxed().toList();
```
tsaglam, line 164: And here.

jplag/src/test/java/de/jplag/special/TokenPrinterTest.java
```diff
@@ -64,6 +65,21 @@ void printKotlinFiles() {
...
+            for (Submission submission : result.getSubmissions().getSubmissions()) {
+                printSubmission(submission);
+            }
+            System.out.println("JPlag printed " + result.getSubmissions().numberOfSubmissions() + " valid submissions!");
+        } catch (ExitException exception) {
+            System.err.println("JPlag threw Error: " + exception.getMessage());
+            fail();
+        }
+    }
+
```
tsaglam, line 82: This is not ideal, as this class is not a test class in the traditional sense. If we want to unit test the token printer, we should actually do that in the module where the token printer resides. Also, no need to run JPlag to print tokens, you can just use the frontend itself and pass the tokens directly to the printer. This makes the unit test more fine-grained.

tsaglam (changes requested):
I found a few things to discuss here.

JanWittler:
Unfortunately, I can't annotate the lines that you didn't change but there are some minor things that should be fixed in the `TokenPrinter` while you are on it anyways.
- `printTokens(TokenList tokens, File directory, Collection<String> fileNames)` should just call `printTokens(tokens, directory, fileNames, Optional.empty())` instead of duplicating the string to file transformation
- the `root` / `directory` parameter in all `printTokens` methods should be renamed to `rootDirectory` consistently

JanWittler (commented):
Code looks way better than before 👍🏼  (sorry Timur 😜 )

tsaglam:
> Code looks way better than before 👍🏼 (sorry Timur 😜 )

I refer to this comment: https://github.com/jplag/JPlag/pull/594#discussion_r947773342 😄

jplag.frontend.golang/src/test/java/de/jplag/golang/GoFrontendTest.java
robinmaisch: The only reason is that it was set to parallel before, and that the order does not matter here. But it is surely not necessary performance-wise.

jplag.frontend-utils/src/main/java/de/jplag/TokenList.java
robinmaisch: Ah, sure, that's a good alternative! Coming right up.

jplag.frontend.golang/src/test/java/de/jplag/golang/GoFrontendTest.java
tsaglam: Ah, I missed that, then it is a mistake that was already in there. Let's remove `parallel()` then.

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
```diff
@@ -227,14 +213,204 @@ private static List<String> linesFromFile(File file) {
...
+        }
+
+        /**
+         * Convenience method to apply a LineDataConsumer to all lineDatas in the given collection.
+         * @param lineDatas the LineData objects to apply the consumer to
+         * @param consumer the LineDataConsumer to apply
+         */
+        public static void forEachApply(Collection<LineData> lineDatas, LineDataConsumer consumer) {
+            lineDatas.forEach(lineData -> lineData.apply(consumer));
+        }
```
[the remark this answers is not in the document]
robinmaisch: Yes, but I thought it looked complicated with two `->` in  the same line.
```java
// TokenPrinter.java::96
lineDatas.forEach(lineData -> lineData.apply((String file, Integer lineNumber, String currentLine, List<Token> tokens) -> {
```
What do you think? Just to reiterate from the documentation, the purpose of the `apply` method is to avoid this:
```java
lineDatas.forEach(lineData -> {
	String file = lineData.file();
	int lineNumber = lineData.lineNumber();
	String currentLine = lineData.currentLine();
	List<Tokens> tokens = lineData.tokens();
	
	// Now, this function really begins.
	// ...
}
```
That's not wrong either, but it distracts from what's really happening.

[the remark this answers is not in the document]
JanWittler: As you access the values each only once (except for the current line), you could even avoid all these variables up-front and just use the accessors where needed. I would prefer that over the current solution as it is just simpler but that is just a subjective opinion. If you prefer your current solution, I am fine with that too.

jplag/src/test/java/de/jplag/special/TokenPrinterTest.java
robinmaisch: So, should TokenPrinterTest.java just be moved to the package de.jplag, or should a separate test class be created?

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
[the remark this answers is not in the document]
robinmaisch: Ah, it doesn't look half as bad as I thought. I'll do just that.

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
```diff
@@ -212,6 +146,58 @@ private static Map<String, List<String>> readFiles(Collection<File> allFiles) {
...
+     * of the file].
+     * @param allFiles a collection of files
+     * @param allTokens a TokenList containing all tokens of the files
+     * @param suffix the optional view file suffix
+     * @param root the root element of the submission
+     * @return map of file -> [LineData]
+     */
+    private static Map<String, List<LineData>> getLineData(Collection<File> allFiles, TokenList allTokens, Optional<String> suffix, File root) {
+        Map<String, List<Token>> filesToTokens = allTokens.stream().collect(Collectors.groupingBy(Token::getFile));
+        Map<String, List<String>> filesToLines = readFiles(allFiles);
```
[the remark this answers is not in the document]
robinmaisch: Now, that makes this method much less intimidating. Great!
Weirdly enough, that makes the `file` parameter of `printTokens` unnecessary.

jplag/src/test/java/de/jplag/special/TokenPrinterTest.java
robinmaisch: Ah, got it. This new test class with the test I mentioned will make the frontend-utils dependent on at least one frontend (during testing only). But that is not an issue, I suppose.

tsaglam: Yes, that should lead to a cyclic dependency. From a software engineering standpoint, the best solution would be to make this test case frontend independent! Basically, create your own mock tokens with only the interfaces provided by the frontend-utils. The references file by the tokens could be anything, e.g., pseudo-code or just Java code. This has the advantage that the unit test does not depend on additional modules, thus keeping the dependencies clean. Moreover, this makes the unit test as fine-grained as possible, ensuring that it will only fail if the tested code changes and code that is not the subject under test, for example, a language frontend.

A sufficient test input could be two files, each containing one line. For those, you could just two tokens in the test case with trivial tokens.

robinmaisch: Ah yes, this is surely the best way to test the functionality of the TokenPrinter. I've come up with some fun test cases*.
The only question left for me is whether I should actively check for the correct output (e.g. by comparing to a hard-coded reference output file) or just accept the output as long as no exceptions occur. The correctness could still be checked "by eye" anytime.

*I realize that the original `printFilesAsSubmissions` test from above points to an issue separate from the TokenPrinter. Is it really intentional that the `file` attribute of tokens may be empty depending on the CLI arguments? This may change also, I think, so I'd rather not add a test that _requires_ the TokenPrinter to deal with the inconsistency.

tsaglam: You could just check if the printed string contains the lines of the file and the string representation of the token. This would be a relatively easy assertion to implement.

I think all tokens should point to a file, but maybe this is a weird edge case for single file submissions?

tsaglam: Also, as a comment: Don't put too much time into the token printer testing, as it is obviously not as important as other parts of our codebase. A simple test or a few simple tests should be enough for now.

tsaglam:
@robinmaisch could you resolve all review comments that are already addressed?

jplag/src/test/java/de/jplag/special/TokenPrinterTest.java
robinmaisch: That's right. Done! I'll go over the SonarCloud issues and that should be it.

tsaglam (approved):

jplag.frontend-testutils/src/test/java/de/jplag/testutils/TokenUtils.java
```diff
@@ -25,12 +24,12 @@ public static List<Integer> tokenTypesByFile(TokenList tokenList, String name) {
...
     /**
      * Returns the tokens in a {@link TokenList} that belong to a file.
-     * @param tokens is the {@link TokenList}.
+     * @param tokenList is the {@link TokenList}.
      * @param name is the name of the target file.
      * @return the immutable list of tokens.
      */
     public static List<Token> tokensByFile(TokenList tokenList, String name) {
-        var tokens = StreamSupport.stream(tokenList.allTokens().spliterator(), false);
+        var tokens = tokenList.allTokens().stream();
```
SuyDesignz, line 32: renaming was a good idea

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
```diff
@@ -24,19 +24,21 @@
...
 public final class TokenPrinter {
+    public static final String NON_WHITESPACE = "\\S";
+    public static final int MIN_PADDING = 1;
     private static final Logger logger = LoggerFactory.getLogger(TokenPrinter.class);
     // Representation:
     private static final String BAR = "|";
     private static final String TAB = "\t";
     private static final String SPACE = " ";
-    public static final String NON_WHITESPACE = "\\S";
-    public static final int MIN_PADDING = 1;
```
SuyDesignz, line 33: i'm not sure right now if moving the lines of code here makes sense. Was there a reason to move them?

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
```diff
@@ -45,157 +47,93 @@ private TokenPrinter() {
...
-                lineIndex = 0;
-            }
+    public static String printTokens(TokenList tokenList, File rootDirectory, Optional<String> suffix) {
+        PrinterOutputBuilder builder = new PrinterOutputBuilder();
+        Map<String, List<Token>> fileToTokens = groupTokensByFile(tokenList);
+
+        fileToTokens.forEach((String fileName, List<Token> fileTokens) -> {
+            builder.append(fileName);
+
+            var lineDatas = getLineData(fileTokens, rootDirectory, suffix);
```
SuyDesignz, line 71: you have always used explicit typing except here. Best to keep consistent

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
```diff
@@ -227,14 +158,179 @@ private static List<String> linesFromFile(File file) {
...
+                return 1;
+            }
+            int minusLength = number < 0 ? 1 : 0;
+            // The 'log10' variant is supposedly faster than the 'toString' variant.
+            return (int) Math.log10(Math.abs(number)) + minusLength + 1;
+        }
+
+        private void resetLinePosition() {
+            columnIndex = 1;
+        }
```
SuyDesignz, line 194: great idea!

jplag.frontend-utils/src/test/java/de/jplag/TestToken.java
```diff
@@ -0,0 +1,19 @@
...
+
+    @Override
+    protected String type2string() {
+        return switch (type) {
+            case STRING -> "STRING";
+            case FILE_END -> "<EOF>";
+            default -> "<UNKNOWN%d>".formatted(type);
+        };
+    }
+}
```
SuyDesignz, line 19: good :+1:

jplag.frontend-utils/src/test/java/de/jplag/TokenPrinterTest.java
```diff
@@ -0,0 +1,98 @@
...
+                        fail("Expected token %s, but found %s".formatted(currentToken, lineToken));
+                    }
+                }
+            }
+        }
+        if (!seenFileName) {
+            fail("Expected file name");
+        }
+    }
+}
```
SuyDesignz, line 98: good :+1:

jplag.frontend-utils/src/test/java/de/jplag/simple/TestTokenConstants.java
```diff
@@ -0,0 +1,5 @@
+package de.jplag.simple;
+
+public interface TestTokenConstants {
+    int STRING = 2;
+}
```
SuyDesignz, line 5: I was a bit confused at the beginning because it contains only one parameter but in itself it is a good idea. :+1:

jplag.frontend.csharp-6/src/test/java/de/jplag/csharp/MinimalCSharpFrontendTest.java
```diff
@@ -47,11 +45,11 @@ void testParsingTestClass() {
...
         TokenList result = frontend.parse(baseDirectory, input);
-        logger.info(TokenPrinter.printTokens(result, baseDirectory, Arrays.asList(input)));
+        logger.info(TokenPrinter.printTokens(result, baseDirectory));
 
         // Compare parsed tokens:
         assertEquals(expectedToken.size(), result.size());
-        List<Integer> actualToken = StreamSupport.stream(result.allTokens().spliterator(), false).map(Token::getType).collect(toList());
+        List<Integer> actualToken = result.allTokens().stream().map(Token::getType).collect(toList());
         assertEquals(expectedToken, actualToken);
     }
```
SuyDesignz, line 55: good 👍

jplag.frontend.scala/src/test/java/de/jplag/scala/ScalaFrontendTest.java
```diff
@@ -162,7 +161,7 @@ private List<Integer> getCodeLines(List<String> lines) {
      * @param fileName The file name of the complete code example
      */
     private void testTokenCoverage(TokenList tokens, String fileName) {
-        var foundTokens = StreamSupport.stream(tokens.allTokens().spliterator(), true).mapToInt(Token::getType).distinct().boxed().toList();
+        var foundTokens = tokens.allTokens().stream().parallel().mapToInt(Token::getType).distinct().boxed().toList();
         var allTokens = IntStream.range(0, ScalaTokenConstants.numberOfTokens()).boxed().toList();
         allTokens = new ArrayList<>(allTokens);
```
SuyDesignz, line 167: good 👍

jplag.frontend.text/src/test/java/jplag/text/TextFrontendTest.java
```diff
@@ -41,11 +39,11 @@ void testParsingJavaDoc() {
...
-        logger.info(TokenPrinter.printTokens(result, baseDirectory, Arrays.asList(input)));
+        logger.info(TokenPrinter.printTokens(result, baseDirectory));
 
         // Compare parsed tokens:
         Map<Integer, Token> tokenTypes = new HashMap<>();
-        StreamSupport.stream(result.allTokens().spliterator(), false).forEach(it -> tokenTypes.put(it.getType(), it));
+        result.allTokens().forEach(it -> tokenTypes.put(it.getType(), it));
 
         assertEquals(293, result.size());
         assertEquals(156, tokenTypes.values().size());
```
SuyDesignz, line 49: good 👍

jplag/src/test/java/de/jplag/TestBase.java
```diff
@@ -20,11 +21,12 @@ protected String getBasePath() {
...
     protected String getBasePath(String... subs) {
-        String path = BASE_PATH;
+        StringJoiner joiner = new StringJoiner(File.separator);
+        joiner.add(BASE_PATH);
         for (String sub : subs) {
-            path += File.separator + sub;
+            joiner.add(sub);
         }
-        return path;
+        return joiner.toString();
```
SuyDesignz, line 29: nicer solution you have implemented here 👍

jplag/src/test/resources/de/jplag/samples/FilesAsSubmissions/Submission1.java
```diff
@@ -0,0 +1,11 @@
...
+
+    public static void main(String[] args) {
+        System.out.println("Hello World!");
+
+        for(int i = 0; i < 10; i++) {
+            System.out.println("Number is " + i);
+        }
+    }
+
+}
```
SuyDesignz, line 11: Test data looks good  👍

SuyDesignz (approved):
I could not find any real problems or inconsistencies in the code. I have left a few comments but in itself it all fits very well 👍

jplag.frontend-utils/src/main/java/de/jplag/TokenPrinter.java
robinmaisch: That must have been the Spotless plugin. You are right, that looks weird.

robinmaisch: I didn't set them public on purpose, so it's good that this came up.

tsaglam:
@robinmaisch, either let me know if this is ready to merge to resolve all comments that were addressed so I know that I can merge.

robinmaisch:
@tsaglam Yes, it is, please go ahead. I'll merge with Rust right after and update the demo files.

Issue #556, opened 2022-08-03
Text frontend unnecessarily overcomplicated


tsaglam:
The [text frontend](https://github.com/jplag/JPlag/tree/master/jplag.frontend.text) is unnecessarily overcomplicated, as the outdated ANTLR grammar can be replaced with functionality provided by CoreNLP. This reduces the size and the complexity.

The grammar essentially splits a text file into words, punctuations, spaces, newlines, and special characters:
https://github.com/jplag/JPlag/blob/24daec9232849befe98499b53c5aa414a21a8c01/jplag.frontend.text/src/main/antlr/text.g#L75-L104

However, only the words are transferred into tokens:
    
> "Don't give me 2 of those! :("

will lead to the following tokens: ```[Don, t, give, me, 2, of, those]```

<br>

**Tasks to do:** 

- [ ] Remove the grammar and change the `ParserAdapter` as follows to replace the [ANTLR grammar](https://github.com/jplag/JPlag/blob/master/jplag.frontend.text/src/main/antlr/text.g)
    - It reads in a file directly
    - Splits the containing text into segments with a call to the CoreNLP tokenizer (see [this example](https://github.com/Gram21/TokenizerExample))
    - Creates the JPlag tokens directly without the grammar from the words created with CoreNLP
    -  The class InputState will become obsolete
- [ ] Make sure to be mindful of all [edge cases](https://github.com/jplag/JPlag/blob/f3b0d1f371cf5b2d03bff7432a4212180165c5d5/jplag.frontend.text/src/main/antlr/text.g#L73-L102).
- [ ] Verify that the old and the new text frontend behave similarly

dfuchss:
We could also think about enhancing this frontend by using standard language techniques (stemming etc.) to find even more similarities.

tsaglam:
> We could also think about enhancing this frontend by using standard language techniques (stemming etc.) to find even more similarities.

I had the same idea and was considering this as a future PISE topic. Maybe we can use CoreNLP here.

Issue #586, opened 2022-08-11
Jar almost 50MB large as it bundles Scala language


JanWittler:
The JPlag Jar is currently 46.8 MB large which appears unnecessary high to me. If unpacked, you can see that this is mainly due to bundling the entire Scala language which unpacked is 77 MB.
This is not a critical issue but we should think about how we can reduce the memory footprint of the Jar.
Options I currently see are to use a different Scala language module parser or to somehow dynamically poll language modules instead of always bundling all.

dfuchss:
The positive thing is that as soon as we clean up the dependencies.
Only the packaged CLI will has this large size. (And of course, everyone who uses scala)

tsaglam:
In the future, we may release two versions of JPlag (and thus two Jars):

- Slim: Only the main programming languages
- Full: All programming languages, modeling languages and text

alberth:
Currently there is a need for `@MetaInfServices(de.jplag.Language.class)` in a language (unless that is old information).

Can't that be used to make an empty JPlag framework jar, and a bunch of jars (one for each language) so the user can download the parts they want  and combine them locally?

tsaglam:
> Currently there is a need for `@MetaInfServices(de.jplag.Language.class)` in a language (unless that is old information).

Yes, the languages are currently located by the service loader.

> Can't that be used to make an empty JPlag framework jar, and a bunch of jars (one for each language) so the user can download the parts they want and combine them locally?

Yes, that would be possible. Each language could be an additional jar; running them locally would require them to be in the same directory, afaik. We currently, however, favor a solution where we have two versions of JPlag to download. A base version with the most commonly used languages that has a small size. And then an extended version including the larger language modules and more exotic languages like the EMF modeling language module. However, this decision is not final so there is still room for discussions.

PR #622, opened 2022-08-26
Replace ANTLR with CoreNLP in text frontend

SirYwell:
Targets #556.

With this PR, text is now tokenized by CoreNLP.

This has some more or less obvious effects:
- CoreNLP does not provide per-token line numbers. Therefore, we need to go through the file contents ourselves to keep track of line numbers and line break indexes to calculate columns. The behavior of supporting CR, LF, and CRLF stays.
- Without further adjustments, CoreNLP recognizes almost all non-whitespace symbols as tokens. Also, the splitting behavior differs (e.g. `1.5` was tokenized as `|1|5` before, now it's just `|1.5`). As a very simple way to get a somewhat similar behavior, I added the `isWord` method that checks if the token contains any alphanumeric symbols.
- CoreNLP also seems to recognize XML. In the javadoc test, it therefore creates one token for each tag.

There are likely more differences on other inputs.

A diff for the current output of the javadoc test can be found here: <https://gist.github.com/SirYwell/93c22c41af2adfdd549e56564c3260f2/revisions> (I simply pasted the original and edited with the current state)

I didn't change the existing test yet as I first would like to get some feedback on the current implementation.

(cc @tsaglam, I started working on this as discussed)


tsaglam:
@SirYwell just FYI I am out of the office until 5.8, I will have a look at your work after that.

README.md
```diff
@@ -29,7 +29,7 @@ In the following, a list of all supported languages with their supported languag
 | [Scheme](http://www.scheme-reports.org)                          |       ? | scheme                | unknown | JavaCC |
 | [EMF Metamodel](https://www.eclipse.org/modeling/emf/)           |  2.25.0 | emf-metamodel         | alpha | EMF |
 | [EMF Metamodel](https://www.eclipse.org/modeling/emf/) (dynamic) |  2.25.0 | emf-metamodel-dynamic | alpha | EMF |
-| Text (naive)                                                     |       - | text                  | legacy | ANTLR |
+| Text (naive)                                                     |       - | text                  | legacy | CoreNLP |
```
JanWittler, line 32: I'd say we can remove the naive part here with this improvement

jplag.frontend.text/src/main/java/de/jplag/text/TokenPosition.java
```diff
@@ -0,0 +1,8 @@
+package de.jplag.text;
+
+public record TokenPosition(
+    int line,
+    int column,
+    int length
+) {
+}
```
JanWittler, line 8: Please provide some minimal documentation here

jplag.frontend.text/src/main/java/de/jplag/text/ParserAdapter.java
```diff
@@ -1,23 +1,36 @@
...
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
 
 public class ParserAdapter extends AbstractParser {
 
     private final Map<String, Integer> tokenTypes = new HashMap<>();
+    private final StanfordCoreNLP pipeline;
     private int tokenTypeIndex = 1; // 0 is FILE_END token, SEPARATOR is not used as there are no methods.
```
JanWittler, line 22: Seems there is an old bug in the text frontend. Please change this value to 2. Even though `SEPARATOR` is not used, as the constants don't have any identity, the value `1` will still be interpreted as it. Thus, we must not use it for custom token constants.

jplag.frontend.text/src/main/java/de/jplag/text/ParserAdapter.java
```diff
@@ -1,23 +1,36 @@
...
 
     private TokenList tokens;
     private String currentFile;
+    private int currentLine;
+    private int currentLineIndex;
```
JanWittler, line 27: This is rather the rather the `globalCharIndex`, right? At least from the name it is not clear to me what the purpose of the attribute is.

jplag.frontend.text/src/main/java/de/jplag/text/TokenPosition.java
JanWittler: Or actually, you could also just remove this record, adjust the constructor of `TextToken` to take these 3 values and pass them directly. As far as I see, the record has no significance.

JanWittler (commented):
Thanks for the improvement. The code looks overall good to me, I added some minor remarks.

As a remark for @jplag/maintainer: This PR increases the binary to ~70MB (see #586). We should keep an eye on this.

jplag.frontend.text/src/main/java/de/jplag/text/ParserAdapter.java
SirYwell: It's the global index of the current line break. I'm not sure about a better name, but I'll add a comment explaining it. If you have a better name, please let me know.

README.md
SirYwell: I don't have a strong opinion on this, but as we don't use any other CoreNLP feature than tokenzing, it's still somewhat naive.

jplag.frontend.text/src/main/java/de/jplag/text/ParserAdapter.java
```diff
@@ -32,42 +49,75 @@ public TokenList parse(File directory, String[] files) {
...
-            inputState = new InputState(inputStream);
-            TextLexer lexer = new TextLexer(inputState);
-            lexer.setFilename(file);
-            lexer.setTokenObjectClass("de.jplag.text.AntlrParserToken");
+    private boolean isWord(CoreLabel token) {
+        // consider a token as a word if it contains any alphanumeric character
+        String text = token.originalText();
+        for (int i = 0; i < text.length(); i++) {
+            char c = text.charAt(i);
+            if (Character.isAlphabetic(c) || Character.isDigit(c)) {
```
tsaglam, line 99: You can use a foreach loop with `String.toCharArray()`. Makes it a bit more cleaner.

tsaglam: Or you can use a more functional one liner:
`return text.chars().anyMatch(it -> Character.isAlphabetic(it) || Character.isDigit(it))`

tsaglam (changes requested):
Just a few minor comments.

jplag.frontend.text/src/main/java/de/jplag/text/ParserAdapter.java
JanWittler: Then I would name it `currentLineBreakIndex` to make things clear.

README.md
dfuchss: I would suggest to keep "naive" until some mechanism like lemma or stemming is added

tsaglam (approved):
Looks good!

README.md
tsaglam: I agree with @dfuchss

SirYwell:
If you're happy with the resulting tokens, I'd update the old test with the new numbers accordingly.

tsaglam:
> If you're happy with the resulting tokens, I'd update the old test with the new numbers accordingly.

@SirYwell Yes, I think the resulting tokens look goo.! You can move ahead!

PR #624, opened 2022-08-28
Enable parallel comparison strategy by default

### LINKED ISSUE #521: Parallel comparison strategy can lead to deadlock for large data sets
When using JPlag with large data sets and the parallel comparison strategy, the comparison sometimes locks and never terminates.

https://github.com/jplag/JPlag/blob/e38f500e381ddedf46e4dcae2b9df11d2dc536e8/jplag/src/main/java/de/jplag/strategy/ParallelComparisonStrategy.java#L25-L131

JanWittler:
As the latest changes to the core algorithm enable it to run in parallel without any locking problems, this PR removes all locking mechanism from the parallel comparison strategy and enables parallel by default. Thus, the sequential comparison strategy and the CLI and JPlag options for the comparison strategy are removed.

Closes #521


JanWittler:
Tested multi-threading support by comparing similarities computed for 393 real submissions using the parallel strategy with the baseline of the similarities when using the normal strategy. For each result, the `List<Match>` was checked for equality with the baseline. I ran 100 iterations of the parallel strategy (each comparing all 393 submissions), for each shuffling the submission tuples a-new to catch as many race conditions as possible. Test succeeded.

Also it appears that the performance for parallel has marginally increased compared to the state before all the core algorithm refactorings. I had on average 5.3s with the new solution, and 5.6s with the old.

tsaglam:
> Also it appears that the performance for parallel has marginally increased compared to the state before all the core algorithm refactorings. I had on average 5.3s with the new solution, and 5.6s with the old.

I think this is low enough so it doesn't matter. Also, most people probably never used the parallel strategy so this PR will feel like an increase in performance.

JanWittler:
> I think this is low enough so it doesn't matter. Also, most people probably never used the parallel strategy so this PR will feel like an increase in performance.

Sorry, maybe I formulated it miss-understandable. The performance is now **better** than before 😄 .

tsaglam:
> > I think this is low enough so it doesn't matter. Also, most people probably never used the parallel strategy so this PR will feel like an increase in performance.
> 
> Sorry, maybe I formulated it miss-understandable. The performance is now **better** than before 😄 .

Ah, I did not expect this, so this is why I misread it! 😄

core/src/test/java/de/jplag/ParallelComparisonTest.java
```diff
@@ -1,28 +1,25 @@
...
+import de.jplag.exceptions.BasecodeException;
 import de.jplag.exceptions.ExitException;
-import de.jplag.strategy.ParallelComparisonStrategy;
 
-/**
- * Currently just a copy of {@link NormalComparisonTest} but for the {@link ParallelComparisonStrategy}. // TODO TS:
- * de-duplicate this
- * @author Timur Saglam
- */
 public class ParallelComparisonTest extends TestBase {
```
tsaglam, line 15: Now that default is parallel, we can maybe just name this class `DefaultComparisonTest`. In the future, we should split it by test case semantics, e.g. a single class for all multi-root tests, as we did for basecode tests.

tsaglam (approved):
Looks great!

core/src/test/java/de/jplag/ParallelComparisonTest.java
tsaglam: I think I will just add a note to the release board and merge this PR!

Issue #634, opened 2022-09-01
Rust frontend tests not finalized


JanWittler:
https://github.com/jplag/JPlag/blob/8ec60b6a307b74bc55db8e8ea6a34ae045148033/jplag.frontend.rust/src/test/java/de/jplag/rust/RustFrontendTest.java#L90-L91

The Rust frontend tests contain incomplete code. Currently, the test condition fails but due to the unfinished code, the test run succeeds and only logs an error. This should be fixed.

Also as a side note: We should avoid the `fail` assertion as this cancels the entire Maven build (even if options like fail-at-end are specified). Instead, use an assertion like `assertTrue` / `assertFalse` or similar.

@robinmaisch maybe you have the time for that? If not, that is no problem (and won't affect the PISE22 outcome).

PR #636, opened 2022-09-06
Finalize Rust frontend test

robinmaisch:
Concerning #634:
The test is now finalized. The acceptable baseline coverage had to be reduced from 90% (which other frontends achieve) to 75% due to the nature of the language and the content and formatting of the test files.

The test should now only fail if the token extraction process is interfered with, e.g. during maintenance or evolution.


tsaglam (approved):

jplag.frontend.rust/src/test/java/de/jplag/rust/RustFrontendTest.java
```diff
@@ -86,15 +87,15 @@ private void testSourceCoverage(String fileName, TokenList tokens) {
             } else {
                 logger.info("Coverage: %.1f%%.".formatted(coverage * 100));
                 logger.info("Missing lines {}", codeLines);
-                if (coverage - 0.9 <= EPSILON) {
-                    // TODO use fail() instead when frontend is ready
+                if (coverage - BASELINE_COVERAGE <= EPSILON) {
                     logger.error("Source coverage is unsatisfactory");
+                    assertTrue(false);
```
JanWittler, line 92: Closing this was maybe a bit too fast...
This should be `assertEquals("source coverage is only <format coverage> but should be <format expected coverage>", BASELINE_COVERAGE, coverage, EPSILON)`

robinmaisch: The method that you refer to checks that the absolute difference is smaller than a delta, right? `coverage` is supposed to be greater than or equal to `BASELINE_COVERAGE`.

JanWittler: Good point. But the if-condition should still be inlined to the assert statement.
`assertTrue(<description>, coverage - BASELINE_COVERAGE > EPSILON)`

robinmaisch: Ah, much better, that's right. I pushed this version now. I need to open another PR, right?

JanWittler: yes

Issue #632, opened 2022-09-01
Documentation End2End testing


JanWittler:
Referenced files:
- https://github.com/SuyDesignz/JPlag/wiki/Applied-concepts-for-end-to-end-testing
- https://github.com/SuyDesignz/JPlag/tree/feature/endToEndDokumentation/jplag.endtoend-testing

This issue is used to collect and discuss feedback regarding the documentation of the end2end testing strategy.

Relates to #317 and #357.
@SuyDesignz @jplag/maintainer

JanWittler:
Thanks for the good and extensive documentation. 
Maybe check the texts again with grammarly.com or similar. I found and listed some typos, but maybe have overseen some.

## Wiki

- I like the part regarding the **Plagiarism and their backgrounds** but don't think it is necessary in that verbosity for the end2end tests. We should think about migrating this part into a distinct section of the wiki such that the page for the end2end tests remains concise.
- In section **Creation of the test data for the End-To-End tests**
  - please use actual markdown lists (`1) 2) ...`) instead of the current bullet point + number combination.
  - Write code as `code` (e.g. in 6.1 or 7.1)
  - decleration -> declaration (in other places incorrect too)
- Check the code indentation of the examples
- You can enable Java syntax highlighting by starting a code block with ```Java
- (3) missing comma in first paragraph (the normalization has changed the function separation)
- (4) `arrayLenght` typo in code
- Indicate gaps in code with `// ...` as this will have nicer syntax highlighting

> If a result differs only in the options, it is very likely that the change is in the configuration of the minimum_token_match. This means that if Option1 does not change in the result of the detection but the result in Option2 does, this is the basis of the minimum_token_match.

I don't get this sentence. The minimum token match is defined by the JSON, so how can it change?

I like that you provide code examples for all plagiarism cases but I would remove the JSONs there. This can also be expressed in code and is then easier to grasp for someone not used to the JSON structure.

Also, you motivate things using a minimum token match of 1. This value makes no sense at all for code, as this will trigger a match whenever two times the same token is extracted and does not require any structural similarity. Also, why 15? The default match of Java is 9 and thus this should be the first reference to use. As a comparison, you could e.g. take 3,4,5 or similar. They should still show the point that a lower MTM increases the similarity.

## Readme

- Remove the [...] in the JSON examples. That red looks bad 😄 
- I think you should keep the numbering of the plagiarism attacks in the readme (as in the wiki)
- Again code in `code blocks`
- api -> API
[...]

PR #645, opened 2022-09-07
End-To-End Dokumentation and cleanup

SuyDesignz:
@jplag/maintainer  @jplag/studdev 
Changes made to README.md documentation. Hereby spelling mistakes, as well as structures, were revised which should provide better comprehensibility and readability. Adapt and normalize expressions to uniformity in the terms. Also change of options for the test runs, because the old options do not provide enough coverage for default settings from (1, 15) to (9, 3).


jplag.endtoend-testing/src/test/java/de/jplag/endtoend/EndToEndSuiteTest.java
```diff
@@ -60,8 +60,8 @@ public EndToEndSuiteTest() throws IOException {
      */
     private void setRunOptions() {
         options = new ArrayList<>();
-        options.add(new Options(1));
-        options.add(new Options(15));
+        options.add(new Options(9));
+        options.add(new Options(3));
     }
```
JanWittler, line 65: I still don't understand why this has to be specified in code in addition to the JSON?

SuyDesignz: The JSON file is only stored for the results of the tests and not for the creation of test cases. It is not intended to load information from the JSON data to generate test cases. The test cases are generated in the code and the results are stored in the JSON files. This is necessary to be able to recognize the results of the last change in the code.

tsaglam:
@SuyDesignz this is currently still a draft PR. Is that correct?

dfuchss:
@SuyDesignz please resolve merge conflicts. You may have to regenerate the results.

tsaglam (changes requested):
This PR is a bit confusing. @SuyDesignz can you elaborate on why you changed the min token matches?

JanWittler:
> This PR is a bit confusing. @SuyDesignz can you elaborate on why you changed the min token matches?

Probably due to my comments in #632:
> Also, you motivate things using a minimum token match of 1. This value makes no sense at all for code, as this will trigger a match whenever two times the same token is extracted and does not require any structural similarity. Also, why 15? The default match of Java is 9 and thus this should be the first reference to use. As a comparison, you could e.g. take 3,4,5 or similar. They should still show the point that a lower MTM increases the similarity.

dfuchss (approved):

tsaglam (approved):

PR #649, opened 2022-09-07
Fix language file suffix bug

tsaglam:
- Extend CLI language test case to include default file suffixes
- Add CLI test case for custom file suffixes
- Fix bug where language suffixes were not used due to an empty array being interpreted as custom suffixes


dfuchss (approved):

tsaglam:
> If not in this PR, we should create an issue to implement regression tests for this issue. I.e. tests that validate whether a submission with multiple files of different types is currently filtered.

There are already test cases that would crash if that issue would persist outside of CLI usage (e.g. the tests that use [this sample](https://github.com/jplag/JPlag/tree/master/jplag/src/test/resources/de/jplag/samples/emptysubmission)). If file filtering breaks even for non-CLI usage the `empty` folder will cause an error as it is not a java file. This did not happen as this issue was CLI-specific.

dfuchss (approved):

PR #661, opened 2022-09-13
Avoid star imports for IntelliJ users

tsaglam:
- Enforces explicit import declarations by adding some IntelliJ-specific configuration.
- TODO make all imports explicit via auto-formatting

@jplag/maintainer what do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,8 @@ result
 hs_err_pid*
 
 # intelliJ
-.idea/
+.idea/*
+!.idea/codeStyles
 *.iml
 .flattened-pom.xml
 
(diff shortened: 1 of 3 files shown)


JanWittler:
Is it also possible to fix that in the spotless configuration?

tsaglam:
> Is it also possible to fix that in the spotless configuration?

Sadly, that is not possible. We could add a rule to the code style section in the wiki and use Sonarlint checks to fail the build when one pushes to a branch/PR. But Spotless cannot do that...

tsaglam:
Closed, as we currently decided against IDE-specific files in the repository.

PR #671, opened 2022-09-16
Normalized similarity range

### LINKED ISSUE #567: Return normalized comparison similarity value (interval [0, 1])
Currently, `JPlagComparison` returns a value between 0 and 100 for the similarity. This is not really intuitive and probably makes computation on these values more complex.

Instead, I would propose to return a value between 0 and 1. This will require adjustments in the results visualization and maybe in the results serialisation.

JanWittler:
Normalizes the similarity value to be in the interval [0,1].
This has the following implications:
- The CLI and Java API got adapted such that the `similarityThreshold` for filtering similarities is in the interval [0,1] too. I considered leaving it in percent, but decided against it as both interfaces will be used by developers which accept normalized values (thus the previous behavior was rather an exception). However, this might introduce problems for developers migrating from a previous JPlag version. As the logger gives a warning when providing a value > 1, I don't consider it severe.
- The report JSON serialization field `match_percentage` becomes a wrong name. I changed it to `similarity`. Thus, any report generated with a previous version of JPlag will not be compatible with the report viewer once the changes are merged. As the JSON structure already got changed significantly to the road to v4.0.0, I don't consider this a problem.

I tested the report viewer in detail. However, as I rarely use it, someone with more experience should recheck that all report viewer views are adjusted correctly.
In addition to these changes, I adapted the match comparison detail view to round the similarity value to `.2f`, as `99.1238123` didn't seem so helpful to me (but `99.12` suffices).

Closes #567


README.md
```diff
@@ -77,7 +77,7 @@ named arguments:
   -x X             All files named in this file will be ignored in the comparison (line-separated list)
   -t T             Tunes the comparison sensitivity by adjusting the minimum token  required  to be counted as a matching section. A smaller
                         <n> increases the sensitivity but might lead to more false-positives
-  -m M             Comparison similarity threshold [0-100]: All comparisons above this threshold will be saved (default: 0.0)
+  -m M             Comparison similarity threshold [0-1]: All comparisons above this threshold will be saved (default: 0.0)
```
tsaglam, line 80:
```suggestion
  -m M             Comparison similarity threshold [0.0-1.0]: All comparisons above this threshold will be saved (default: 0.0)
```
To make it clear that this is a double.

core/src/main/resources/de/jplag/messages.properties
```diff
@@ -9,7 +9,7 @@ CommandLineArgument.RootDirectory=Root-directory with submissions to check for p
 CommandLineArgument.PlagiarismDirectory=Root-directory with submissions to check for plagiarism
 CommandLineArgument.PriorDirectory=Root-directory with prior submissions to compare against
 CommandLineArgument.ShownComparisons=The maximum number of comparisons that will be shown in the generated report, if set to -1 all comparisons will be shown
-CommandLineArgument.SimilarityThreshold=Comparison similarity threshold [0-100]: All comparisons above this threshold will be saved
+CommandLineArgument.SimilarityThreshold=Comparison similarity threshold [0-1]: All comparisons above this threshold will be saved
```
tsaglam, line 12:
```suggestion
CommandLineArgument.SimilarityThreshold=Comparison similarity threshold [0.0-1.0]: All comparisons above this threshold will be saved
```
To make it clear that this is a double.

tsaglam (approved):
Looks good!

PR #673, opened 2022-09-19
Fix bug that prevented the basecode fallback when the subdirectory option is used

### LINKED ISSUE #660: Legacy bascode fallback does not work when using subdir argument
When providing a submission name as basecode in a single-root scenario, the legacy fallback allows using it even though the basecode is not specified as a global path. However, when using the `-s` argument for subdirectories, this does no longer work. A workaround is to specify the relative path to the basecode: `./rootFolder/submissionName`.

tsaglam:
Resolves #660:
- Adds test case that reproduces the bug
- Fix the bug
- Enforce use of the delta constant for double comparison in test cases
- Fix name of basecode option (usually it is a path not a name)

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,8 +89,8 @@ named arguments:
 The new API makes it easy to integrate JPlag's plagiarism detection into external Java projects:
 
 ```java
-JPlagOptions options = new JPlagOptions(List.of("/path/to/rootDir"), List.of(), LanguageOption.JAVA);
-options.setBaseCodeSubmissionName("template");
+Language language = LanguageLoader.getLanguage(Language.IDENTIFIER).orElseThrow();
+JPlagOptions options = new JPlagOptions(language, List.of("/path/to/rootDir"), List.of()).withBaseCodeSubmissionPath("template");
 
 JPlag jplag = new JPlag(options);
 JPlagResult result = jplag.run();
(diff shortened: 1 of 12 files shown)


JanWittler (changes requested):
Two comments without having thoroughly checked the code.
1) Why do you make `foundSubmissions` a property? From my knowledge, local variables should always be favored over properties.

2) Is it intentional that you don't adapt the `loadBaseCodeAsPath` method? I.e. should the subdirectory only apply when the legacy base code specification is used?

(👍🏼 for the addition of regression tests)

tsaglam:
>     1. Why do you make `foundSubmissions` a property? From my knowledge, local variables should always be favored over properties.

Usually, yes. But in this case, the found submissions are already treated as state, thus I made it explicit. It is modified in several different methods, but it was really hard to find out where because it is passed several levels deep through multiple method calls. By turning it into a property, it becomes clear that it is shared state amongst the methods. It is also easier to see where it is actually used! 

>     2. Is it intentional that you don't adapt the `loadBaseCodeAsPath` method? I.e. should the subdirectory only apply when the legacy base code specification is used?

No, it always applies. `loadBaseCodeAsPath` should work as intended, one of the test cases I added is for that case. I will add some more assertions there, but it seems that subdirectory also works for global basecode.

JanWittler:
Problem also fixed in #674.
(Surprisingly) I am in favor of merging #674 instead of this one and thus will defer my review here until the other PR is closed or merged.

PR #674, opened 2022-09-20
Make directory arguments in JPlagOptions type-safe

### LINKED ISSUE #625: Re-evaluate `JPlagOptions#baseCodeSubmissionName` backwards compatibility
The current implementation supports backwards compatibility of the `JPlagOptions#baseCodeSubmissionName` as documented here:
```Java
/**
 * Path name of the directory containing the base code.
 * <p>
 * For backwards compatibility it may also be a directory name inside the root directory. Condition for the latter is
 * <ul>
 * <li>Specified path does not exist.</li>
 * <li>Name has not have a separator character after trimming them from both ends (leaving at least a one-character
 * name).</li>
 * <li>A submission with the specified name exists in the root directory.</li>
 * </ul>
 * It's an error if a string has been provided but it is neither an existing path nor does it fulfill all the conditions
 * of the compatibility fallback listed above.
 * </p>
 */
```

However, as we are moving towards a new major release, we should re-evaluate whether this functionality should be preserved. If we want to preserve it, we should at least refactor it to resolving the base path before initialising the options, such that the options are always initialised with a path (and then rename the attribute to `baseCodeSubmissionPath`.

### LINKED ISSUE #660: Legacy bascode fallback does not work when using subdir argument

JanWittler:
Currently, the `submissionDirectories`, `oldSubmissionDirectories` and `baseCodeSubmissionName` are weakly typed as `List<String` resp. `String`. This results in several checks in the `SubmissionSetBuilder` that the Strings are actual File paths and that there are no duplicates in the lists. Thus, this PR refactors the API such that the directory arguments are of type `Set<File>`. Additionally, the baseCodeSubmissionName is replaced with `File baseCodeSubmissionDirectory` which requires the user to provide the correct path to the base code submission directory instead of the previously available support for subdirectory names. To preserve the previous behavior, a legacy initializer and a legacy factory method (marked with `@Deprecated(since = "4.0.0", forRemoval = true)`) were added to the options record.

This refactoring greatly simplifies the submission set builder, makes it clear that the string-based base code API is deprecated and has a high separation of concerns, such that the deprecated API can be removed easily in v5.0.0.

Additionally adapts the `-s` CLI argument to be lowercase as all other arguments.

Includes the regression tests from #673 and thus can replace it.
Closes #625.
Closes #660.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ named arguments:
   -bc BC           Path of the directory containing the base code (common framework used in all submissions)
   -v               {quiet,long} Verbosity of the logging (default: quiet)
   -d               Debug parser. Non-parsable files will be stored (default: false)
-  -S S             Look in directories <root-dir>/*/<dir> for programs
+  -s S             Look in directories <root-dir>/*/<dir> for programs
   -p P             comma-separated list of all filename suffixes that are included
   -x X             All files named in this file will be ignored in the comparison (line-separated list)
   -t T             Tunes the comparison sensitivity by adjusting the minimum token  required  to be counted as a matching section. A smaller
--- a/cli/src/main/java/de/jplag/CLI.java
+++ b/cli/src/main/java/de/jplag/CLI.java
@@ -48,6 +48,7 @@ import org.slf4j.LoggerFactory;
 
 import de.jplag.clustering.ClusteringOptions;
 import de.jplag.clustering.Preprocessing;
+import de.jplag.exceptions.BasecodeException;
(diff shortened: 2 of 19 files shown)


tsaglam (approved):
Looks good to me!

Issue #454, opened 2022-06-10
Issues within the frontend architecture


tsaglam:
This issue is meant to collect problems with the design of the frontends. 
Known problems are:

- [ ] The name _frontend_ is not very good, as they are technically not frontends but rather language modules for parsing and tokenization.
- [ ] The token constants should not be constants in an interface with manually defined IDs, they should be enum literals in an enum, where each language should have its own enum and the ordinal is the ID.
- [ ] The common tokens `FILE_END` and `SEPARATOR_TOKEN` should not be identified via the index but rather via methods in the token enums (prescribed by a common interface which should be part of the front-end utils).
- [ ] The language interface should offer default methods for method signatures that are often implemented the same way (e.g. `usesIndex()` with false, `isPreformatted()` with true)

Alberth289346:
``` java
// submission.java line 258
        tokenList = language.parse(submissionRootFile, relativeFilePaths);
        if (!language.hasErrors()) {
```
Just passing by, saw this issue, remembered the above interface weirdness.

It's parsing a submission, and you get the token list. Then separately you must query the error status.
I think a better solution is to pass both back together from the `parse` call by wrapping them into an class or so.

tsaglam:
Some of these issues are currently being addressed by @JanWittler, his PR will follow soon.

tsaglam:
Suceeded by #983.

PR #688, opened 2022-09-22
Language API cleanup

JanWittler:
- Replaces `parseFiles(File directory, String[] fileNames)` with `parseFiles(Set<File> files) throws ParsingException`
  - the exception replaces the `hasErrors` field (and solves #454, bullet point 8)
- Refactors `Token::file` to type `File` instead of `String` to avoid file-path magic in the language frontends
  - to preserve consistency with the previous state, in the `ComparisonReportWriter` the files are relativized to the submission root directory

### DIFF
--- a/languages/java/src/main/java/de/jplag/java/JavacAdapter.java
+++ b/languages/java/src/main/java/de/jplag/java/JavacAdapter.java
@@ -40,7 +40,6 @@ public class JavacAdapter {
             final Trees trees = Trees.instance(task);
             final SourcePositions positions = trees.getSourcePositions();
             for (final CompilationUnitTree ast : executeCompilationTask(task, parser.logger)) {
-                final String filename = ast.getSourceFile().getName();
                 File file = new File(ast.getSourceFile().toUri());
                 final LineMap map = ast.getLineMap();
                 ast.accept(new TokenGeneratingTreeScanner(file, parser, map, positions, ast), null);
--- a/languages/java/src/main/java/de/jplag/java/JavacAdapter.java
+++ b/languages/java/src/main/java/de/jplag/java/JavacAdapter.java
@@ -3,9 +3,9 @@ package de.jplag.java;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 
 import javax.tools.Diagnostic;
 import javax.tools.DiagnosticCollector;
@@ -28,11 +28,11 @@ public class JavacAdapter {
 
     private static final JavaCompiler javac = ToolProvider.getSystemJavaCompiler();
 
-    public int parseFiles(File directory, Iterable<File> pathedFiles, final Parser parser) {
+    public int parseFiles(Set<File> files, final Parser parser) {
         var listener = new DiagnosticCollector<>();
 
         try (final StandardJavaFileManager fileManager = javac.getStandardFileManager(listener, null, StandardCharsets.UTF_8)) {
-            var javaFiles = fileManager.getJavaFileObjectsFromFiles(pathedFiles);
+            var javaFiles = fileManager.getJavaFileObjectsFromFiles(files);
 
             // We need to disable annotation processing, see
             // https://stackoverflow.com/questions/72737445/system-java-compiler-behaves-different-depending-on-dependencies-defined-in-mave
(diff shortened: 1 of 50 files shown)


JanWittler:
The checks fail currently because there is too much code duplication (which is not new, but now triggered as parts of it changed). To solve this, I would need to refactor all language frontend parsers to a shared file-based parser. However, this would make the diff of the PR pretty ugly and extend its scope. Thus, if possible, I would argue for ignoring the code duplication issue (for now).

languages/java/src/main/java/de/jplag/java/JavacAdapter.java
```diff
@@ -61,23 +68,18 @@ private Iterable<? extends CompilationUnitTree> executeCompilationTask(final Com
...
         for (Diagnostic<?> diagnosticItem : listener.getDiagnostics()) {
             if (diagnosticItem.getKind() == javax.tools.Diagnostic.Kind.ERROR) {
+                File file = null;
+                if (diagnosticItem.getSource() instanceof JavaFileObject) {
+                    JavaFileObject fileObject = (JavaFileObject) diagnosticItem.getSource();
+                    file = new File(fileObject.toUri());
+                }
                 logger.error("{}", diagnosticItem);
-                errors++;
+                throw new ParsingException(file, diagnosticItem.getMessage(Locale.getDefault()));
```
tsaglam, line 80: This is a bit of changed behavior, as you now only report the first error found by the compiler. Previously it would tell you, for example, all compiler errors it found. Maybe we add a constructor to the parsing exception that allows wrapping multiple issues (either multiple file/message pairs or just wrapping other parsing exceptions in a composite way).

languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java
```diff
@@ -391,7 +401,7 @@ public Object visitExports(ExportsTree node, Object p) {
 
     @Override
     public Object visitErroneous(ErroneousTree node, Object p) {
-        parser.increaseErrors();
+        parsingException = new ParsingException(file, "error while visiting %s".formatted(node));
```
tsaglam, line 404: Can this be called multiple times? If yes, we may accidentally override this exception.

languages/python-3/src/main/java/de/jplag/python3/Parser.java
```diff
@@ -32,25 +34,22 @@ public Parser() {
...
     }
 
-    private boolean parseFile(File directory, String file) {
+    private void parseFile(File file) throws ParsingException {
         BufferedInputStream inputStream;
 
         CharStream input;
         try {
-            inputStream = new BufferedInputStream(new FileInputStream(new File(directory, file)));
+            inputStream = new BufferedInputStream(new FileInputStream(file));
```
tsaglam, line 52: Not your fault, but this resource is never closed, could be a try-with-resource. Is a legacy issue though.

tsaglam (commented):
Just a few issue. One question regarding the parsing: If I parse a submission with multiple files, and one of the files leads to a ParsinException, will the parsing of the submission be canceled, or will the remaining files still be parsed? If it is the former, this might change the behavior of JPlag since some frontends were previously able to recover from minor issues in some files. Other files of the submission were still parsed.

JanWittler:
> Just a few issue. One question regarding the parsing: If I parse a submission with multiple files, and one of the files leads to a ParsinException, will the parsing of the submission be canceled, or will the remaining files still be parsed? If it is the former, this might change the behavior of JPlag since some frontends were previously able to recover from minor issues in some files. Other files of the submission were still parsed.

Right now, parsing a submission is cancelled as soon as any error occurs. Can you give an example where an error recovery was possible? The specific language frontend can be adapted to again support the error recovery without changing the interface signature.

languages/java/src/main/java/de/jplag/java/JavacAdapter.java
JanWittler: So you mean something like `ParsingException(Iterable<ParsingException> wrappedExceptions)`? Should be possible to do so.

languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java
JanWittler: I did not check the documentation in detail but I guess it was possible. As previously we only shared the count of errors, I didn't care to overwrite error metadata but I can adapt it to support logging multiple exceptions.

tsaglam:
> Right now, parsing a submission is cancelled as soon as any error occurs. Can you give an example where an error recovery was possible? The specific language frontend can be adapted to again support the error recovery without changing the interface signature.

I thought the Java language module could do that, but maybe I am wrong. I will test it.

languages/java/src/main/java/de/jplag/java/JavacAdapter.java
JanWittler: Added in https://github.com/jplag/JPlag/pull/688/commits/b72c4e505e09a45f7b8bb8b2450ef588e0245d2b

JanWittler:
I adapted the Java language frontend to collect all errors first and then throw at the end if any error occurred. However, I did not find any error recovery mechanism.

tsaglam (approved):

Issue #686, opened 2022-09-22
Report Viewer Bug Fixes and Tests for v4.0.0 Release


sebinside:
There are a couple of bugs and tests left to make the report viewer release ready for version 4.0. This issue collects the known problems:

1. Matches are not represented correctly if not all source code files are in the root folder (which is, e.g., the case with packages in java). In the report viewer, files are represented by their name while matches are represented with the full path. This breaks the viewer. This issue is also documented in #357 and #658. Here, we need to use the relative path whenever possible to avoid any collisions in file names (see also problem 2). 
2. The Report Generation does overwrite source code files if they have the same name because all files are "flat mapped", i.e., written in one single folder per submission. It does not throw an error but will put out wrong results that might be noticed later (or even not noticed at all). Here, the solution is probably to preserve the original folder structure also in the resulting zip file in the submissions folder. This can be easily replicated by creating a structure with files with different content in a submission like this:
    ```
    SourceFile.java
    subfolder/SourceFile.java
    ````
3. Problem 2 could also affect the multi-root and subdir features (see the [readme](https://github.com/jplag/JPlag#cli) for details on the usage). In particular, this might break if two root directories have the same name. Here, more testing is required.
4. The report viewer sometimes breaks with clusters activated. This problem requires more insights to be replicated. Also, Problems 1- 3 are more important. For now, cluster serialization is disabled in #691.
5. Add a hard version flag to the `overview.json` that relates to the release version, e.g., `v4.0.0`. If the flag does not fit the report viewer version, a warning should be printed. This is already addressed in a naive way in #691.

JanWittler:
This problem even extends further. We implicitly expect the language frontends to provide tokens with file names relative to the root directory of the associated submission, as we write that file name of the token to the JSON output. Thus, if the assumption is not hold, the file cannot be found, even if we fix things in the report viewer.
I refactored `Token::file` to type `File` in #688 such that the submission-relative file has to be computed explicitly by the report generator instead of having this implicitly distributed over the language frontends.

sebinside:
This issue has been replaced by #710.

PR #691, opened 2022-09-28
Quick fix for matches not display properly in the report viewer.

sebinside:
This fix shall quickly solve the match display in the report viewer. It does not completely solve the underlying issue, see #686. 

In addition, the serialization of clusters is disabled in the report viewer output. 

Last, this PR also adds a very simple version checking in the report viewer.


core/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -42,6 +43,9 @@ public class ReportObjectFactory {
     private static final ToDiskWriter fileWriter = new ToDiskWriter();
     public static final String OVERVIEW_FILE_NAME = "overview.json";
     public static final String SUBMISSIONS_FOLDER = "submissions";
+
+    // This shall be moved to a better visible and upgradable position. Shall be fixed in a future version.
```
JanWittler, line 47:
```suggestion
    // TODO: This shall be moved to a better visible and upgradable position. Shall be fixed in a future version.
```

core/src/main/java/de/jplag/reporting/reportobject/mapper/ClusteringResultMapper.java
```diff
@@ -21,7 +19,11 @@ public ClusteringResultMapper(Function<Submission, String> submissionToIdFunctio
 
     public List<Cluster> map(JPlagResult result) {
         var clusteringResult = result.getClusteringResult();
-        return clusteringResult.stream().map(ClusteringResult::getClusters).flatMap(Collection::stream).map(this::convertCluster).toList();
+
+        // Clustering disabled for the report viewer. Shall be fixed in a future version.
```
JanWittler, line 23:
```suggestion
        // TODO: Clustering disabled for the report viewer. Shall be fixed in a future version.
```

core/src/main/java/de/jplag/reporting/reportobject/model/Version.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * This record represents a version number according to <a href="https://semver.org/">https://semver.org/</a>. This
+ * version number is used to compare generated result.zip files with the version of the report-viewer.
+ * @param major MAJOR version when you make incompatible API changes
+ * @param minor MINOR version when you add functionality in a backwards compatible manner
+ * @param patch PATCH version when you make backwards compatible bug fixes
+ */
+public record Version(@JsonProperty("major") int major, @JsonProperty("minor") int minor, @JsonProperty("patch") int patch) {
+}
```
JanWittler, line 13: Maybe a bit over engineered, don't you think? 😄 I'd say a version String would be sufficient.

report-viewer/src/model/factories/OverviewFactory.ts
```diff
@@ -68,6 +81,14 @@ export class OverviewFactory {
...
   }
 
+  static compareVersions(jsonVersion: Version, reportViewerVersion: Version) {
+    if(jsonVersion.major !== reportViewerVersion.major ||
+       jsonVersion.minor !== reportViewerVersion.minor ||
+       jsonVersion.patch !== reportViewerVersion.patch) {
+        console.warn("The result's version tag does not fit the report viewer's version. Trying to read it anyhow but be careful.")
+       }
+  }
+
```
JanWittler, line 91: If we simplify it to String, we can use `semver` to compare the version here

https://stackoverflow.com/questions/6832596/how-can-i-compare-software-version-number-using-javascript-only-numbers

report-viewer/vue.config.js
```diff
@@ -11,5 +11,6 @@ module.exports = defineConfig({
         chunks: "all",
       },
     },
+    devtool: 'source-map'
```
JanWittler, line 14: What is this for?

core/src/main/java/de/jplag/reporting/reportobject/model/OverviewReport.java
```diff
@@ -8,6 +8,8 @@
 
 public record OverviewReport(
 
+        @JsonProperty("version") Version version,
```
JanWittler, line 11: Maybe we name this property rather `JPlag-version` or similar? I could image in the future to introduce an actual versioning of the JSON report structure, and then we would have two version numbers: version of the json format and the version of JPlag used to generate it.

JanWittler (changes requested):

report-viewer/src/model/factories/ComparisonFactory.ts
```diff
@@ -21,6 +21,13 @@ export class ComparisonFactory {
     );
 
     const matches = json.matches as Array<Record<string, unknown>>;
+    
+    // Quick fix for issue #658, requires deeper insights to be resolved completely
```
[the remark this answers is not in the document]
sebinside: Unfortunately, SonarCloud does not like correct annotations in comments ;)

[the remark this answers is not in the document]
JanWittler: We can still merge it even if SonarCloud is complaining.

report-viewer/src/model/factories/OverviewFactory.ts
sebinside: I don't think that replacing a quite simple solution with yet another dependency is a that good approach :D

report-viewer/vue.config.js
sebinside: Used for debugging the report viewer. I also wondered why it was not already there.

JanWittler (approved):

PR #701, opened 2022-09-29
Suppress the StanfordNLP logger

tsaglam:
Hotfix that suppresses the StanfordNLP logger directly, with the exception of errors.

### DIFF
--- a/languages/text/src/main/java/de/jplag/text/ParserAdapter.java
+++ b/languages/text/src/main/java/de/jplag/text/ParserAdapter.java
@@ -15,6 +15,7 @@ import de.jplag.Token;
 import edu.stanford.nlp.ling.CoreLabel;
 import edu.stanford.nlp.pipeline.CoreDocument;
 import edu.stanford.nlp.pipeline.StanfordCoreNLP;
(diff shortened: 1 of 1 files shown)


dfuchss:
Log levels should not be managed in the client code (from my point of view). This should be done in the logger config of the CLI.

dfuchss:
See https://github.com/ArDoCo/Core/blob/10a0a15317dc9aefe070a50e33ecd6d52b19b4cd/cli/src/main/resources/simplelogger.properties#L13 as an example.

tsaglam:
> See https://github.com/ArDoCo/Core/blob/10a0a15317dc9aefe070a50e33ecd6d52b19b4cd/cli/src/main/resources/simplelogger.properties#L13 as an example.

I tried that before this fix, did not work here. CoreNLP was still logging.

dfuchss:
Ah .. we have to adapt our logger. Nevertheless, we should not change log levels programmatically

tsaglam:
> Ah .. we have to adapt our logger. Nevertheless, we should not change log levels programmatically

I will make an issue for this. for now we use the hotfix due to time constraints.

Issue #576, opened 2022-08-10
Open tasks regarding the EMF metamodel language modules


tsaglam:
The metamodel frontend is now part of JPlag, but there are some tasks left to do:

- [ ] The frontend readmes are missing
- [ ] The CLI arguments defined in the language options are a bit verbose
- [ ] Generally, it is not intuitive that there are two frontends for metamodels, users do not know which to use when
- [ ] The visualization is currently just a very simple tree view, this could be improved (e.g. with [Emfatic](https://www.eclipse.org/emfatic/))
- [ ] The wiki needs documentation for metamodel plagiarism detection
- [ ] The MODELS EduSym paper should be linked

tsaglam:
>  Solve the problem of token order in a match and its visualization

Resolved by #971.

tsaglam:
Some minor tasks are left, but the most important issues for the upcoming release are resolved.

PR #702, opened 2022-09-29
EMF modeling language fixes

tsaglam:
Contributes towards #576:
- Hide the dynamic EMF language, as the other is currently the one to use
- Shorten the EMF language identifiers
- Fixes EMF tree views

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ target/
 
 *.class
 
-result
+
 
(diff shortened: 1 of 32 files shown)


dfuchss:
I would consider just removing the dependency from CLI to emf-dynamic. I'm not sure about hidden languages. Can they be loaded etc. Because the CLI does not know anything about the language option.

sebinside (approved):

PR #731, opened 2022-10-14
Refactor End to End Tests

JanWittler:
This PR refactors the end to end tests. In particular, the following improvements were implemented:
- generation of the result json was extracted into a separate test class `EndToEndGeneratorTest`
- E2E tests were simplified to load all configuration based on the file paths and result json instead of having to keep the file hierarchy and initialization in code in sync
  - this only removes the limitation that was previously present of only supporting the same options for all test suites 
- JPlag is only executed once for each test suite + options pair, instead of for each test case individually
- E2E tests are generated hierarchically as follows:
```
- <language>
  - <test suite>
    - <options>
      - test1
      - test2
      - ...
    - <options>
      - ... 
- ...
```  
- helper files were greatly simplified


tsaglam (approved):
Besides the minor leftover issue from sonar, this looks good!

PR #748, opened 2022-10-18
Filter for deleting tokens in unreachable code

Squamanto:
C and C++ allow placing Code after guaranteed jump statements, like `return` or `break`.
This opens up the possibility for arbitrary code insertions after these statements to escape plagiarism detection.

This PR adds a static filter method based on a state machine.
It can be applied to a list of CPP tokens to filter some of the tokens after a guaranteed jump.
Some cases are ambiguous and are therefore not detectible.

This is a prototypical implementation and can not be enabled using options.
For testing purposes, the filtering can be enabled by adding `UnreachableCodeFilter.applyTo(tokens);` at the end of the `scan` method in `de.jplag.cpp.Scanner`.

This PR is part of the implementation from my Bachelor's Thesis _"Preventing Code Insertion Attacks on Token-Based Software Plagiarism Detectors"_.

### DIFF
--- a/languages/cpp/src/main/java/de/jplag/cpp/experimental/UnreachableCodeFilter.java
+++ b/languages/cpp/src/main/java/de/jplag/cpp/experimental/UnreachableCodeFilter.java
@@ -1,4 +1,4 @@
-package de.jplag.cpp;
+package de.jplag.cpp.experimental;
 
 import static de.jplag.SharedTokenType.FILE_END;
 import static de.jplag.cpp.CPPTokenType.*;
--- a/languages/cpp/src/main/java/de/jplag/cpp/experimental/UnreachableCodeFilter.java
+++ b/languages/cpp/src/main/java/de/jplag/cpp/experimental/UnreachableCodeFilter.java
@@ -6,6 +6,7 @@ import static de.jplag.cpp.CPPTokenType.*;
 import java.util.List;
 import java.util.ListIterator;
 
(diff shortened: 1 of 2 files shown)


languages/cpp/src/main/java/de/jplag/cpp/experimental/UnreachableCodeFilter.java
```diff
@@ -0,0 +1,138 @@
...
+public final class UnreachableCodeFilter {
+
+    private UnreachableCodeFilter() {
+    }
+
+    /**
+     * Applies the filtering on the provided token list.
+     * @param tokenList The list that will be filtered. The contents of this parameter will be modified.
+     */
+    public static void applyTo(List<de.jplag.Token> tokenList) {
```
tsaglam, line 23: Any reason for the fully qualified name instead of an import?

tsaglam (approved):
Just one minor issue...

languages/cpp/src/main/java/de/jplag/cpp/experimental/UnreachableCodeFilter.java
Squamanto: There is a `de.jplag.cpp.Token` and a `de.jplag.Token` class, so I originally put it there for clarity.
But since only one of the two classes is used here, the import suffices and this is indeed redundant.
I will remove the fully qualified names.

PR #749, opened 2022-10-18
Experimental source analysis

Squamanto:
Inserting unused variables into source code can prevent plagiarism detection.

To counteract this, this PR contains a prototypical implementation for using GCC to find unused variables.
The lines containing unused variables are stored and the `de.jplag.cpp.Scanner` can then query if a token should be added to the token list during scanning.

This is a prototypical implementation and can not be enabled using options.
For testing purposes, this can be enabled by adding an `GCCSourceAnalysis` instance to `de.jplag.cpp.Scanner`.
The `findUnusedVariableLines` method has to be called at the beginning of the `scan` method.
The `isTokenIgnored` has to be called in the `add` method to check if the token should be skipped.

This PR is part of the implementation from my Bachelor's Thesis _"Preventing Code Insertion Attacks on Token-Based Software Plagiarism Detectors"_.

### DIFF
--- a/languages/cpp/src/main/java/de/jplag/cpp/experimental/GCCSourceAnalysis.java
+++ b/languages/cpp/src/main/java/de/jplag/cpp/experimental/GCCSourceAnalysis.java
@@ -20,7 +20,7 @@ public class GCCSourceAnalysis implements SourceAnalysis {
 
     private final Logger logger;
 
-    public GCCSourceAnalysis(Logger logger) {
+    public GCCSourceAnalysis() {
         this.logger = LoggerFactory.getLogger(this.getClass());
     }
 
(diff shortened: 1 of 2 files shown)


languages/cpp/src/main/java/de/jplag/cpp/experimental/GCCSourceAnalysis.java
```diff
@@ -0,0 +1,79 @@
...
+        var lineSplit = line.split(":");
+
+        String fileName = new File(lineSplit[0]).getName();
+
+        int lineNumber = Integer.parseInt(lineSplit[1]);
+
+        if (linesToDelete.containsKey(fileName)) {
+            linesToDelete.get(fileName).add(lineNumber);
+        } else {
+            linesToDelete.put(fileName, new ArrayList<>(lineNumber));
```
SirYwell, line 76: This creates a pre-sized ArrayList instead of a List containing the lineNumber as element. I'd recommend to replace the whole if-else with something like `linesToDelete.computeIfAbsent(fileName, fn -> new ArrayList<>()).add(lineNumber);`

Squamanto: This is a huge oversight on my part, thanks for pointing this out!

I was not aware of `computeIfAbsent`, but I really like your solution and added it to the latest commit.

tsaglam (approved):

PR #784, opened 2022-11-11
Feature: Auto Release of JAR

dfuchss:
Added a job that attaches the compiled jar to a new release (part of the release workflow)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,15 +1,14 @@
-name: Deploy to Maven Central
+name: Deploy to Maven Central & GitHub
 on:
   workflow_dispatch:
   release:
     types: [created]
 jobs:
-  publish:
+  publish-maven-central:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


tsaglam (approved):
Looks good, but is it a draft release by default?
Or do we need to specify `draft: true`?

dfuchss:
> Looks good, but is it a draft release by default? Or do we need to specify `draft: true`?

The Trigger for this workflow is the creation of a release. Therefore, I didn't specified draft mode (as you have created the release before ;) )

tsaglam:
> > Looks good, but is it a draft release by default? Or do we need to specify `draft: true`?
> 
> The Trigger for this workflow is the creation of a release. Therefore, I didn't specified draft mode (as you have created the release before ;) )

Ah, my bad, so I create the release manually, and the JAR will be added automatically, right?

dfuchss:
Yes, you create a release -> the workflow will create the release for Maven Central and add the JAR to the release

Issue #658, opened 2022-09-09
Report Viewer does not show matches properly


sebinside:
The current version of the Report Viewer displays matches only if the corresponding files are inside the root folder (e.g., in this example here: [result-working.zip](https://github.com/jplag/JPlag/files/9536410/result-working.zip)). If they are in subfolders (e.g., in this example in submission A: [result-notworking.zip](https://github.com/jplag/JPlag/files/9536409/result-notworking.zip)), they are not resolved properly.

A little debugging showed the underlying problem: Matches are represented with their full path, i.e. `subfolder/SimpleDuplicate.java` in the matches map. File names (that are used in the viewer to query the matches), are represented by their name only, i.e. `SimpleDuplicate.java`.

As I don't know which parts of the viewer depend on which type of file representation, I did not change it but rather hope for the help of @nestabentum 😄 

Screenshot of the bug:
![image](https://user-images.githubusercontent.com/7331248/189379320-1f2b1346-55e6-4e8b-ae10-b12f81d8827a.png)

nestabentum:
Oh, I oversaw this issue, sorry. Currently, I'm pretty swamped, could take a look at this next week, though.

nestabentum:
Do we want  the title of a nested file to indicate the nesting by being named `subfolder/Matches.java` or do we prefer `Matches.java`?

<img width="492" alt="image" src="https://user-images.githubusercontent.com/33717535/192085536-c6e5e252-2909-4137-b32f-dff37d5aeebc.png">

Deep nesting could lead to long titles introducing the need to abbreviate: `../subfolder/Matches.java`. 
Also noteworthy: the submission name generated by JPlag is platform-dependent as it contains the platform-specific folder separator.

tsaglam:
> Do we want the title of a nested file to indicate the nesting by being named `subfolder/Matches.java` or do we prefer `Matches.java`?

`subfolder/Matches.java` is better I think, at least if we abbreviate for long paths.

> Also noteworthy: the submission name generated by JPlag is platform-dependent as it contains the platform-specific folder separator.

Ah, interesting, but makes sense. @sebinside @JanWittler That should still work with the recent/planned changes, right?

nestabentum:
I fear the solution to this problem is not as trivial as I hoped...

The names of the files in a match are determined via `Token#getFile` in `ComparisonReportWriter#convertMatchToReportMatch`. `Token#file` is set by the language model's parser. 
[...]

PR #793, opened 2022-11-15
Fix bug: representation for name of match and file

cyfml:
This is new PR, because last one has conflict problem.
I have changed the submissions's folder structure and json file of comparison(i.e. A-B.json) in zip file. Now match and file name are represented with their full path (include root directory). And the problem in #658 is also solved. And when file name is too long, it only shows part of the full path(i.e. ..subfolder/subsubfolder/algorithm.java).

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ JPlag is released on [Maven Central](https://search.maven.org/search?q=de.jplag)
 5. You will find the generated JARs in the subdirectory `jplag.cli/target`.
 
 ## Usage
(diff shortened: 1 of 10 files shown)


tsaglam:
@cyfml the history is still not clean. Please also do not open up and close multiple PRs.

JanWittler:
The problem is that the branch is based on the `main` branch but the PR targets the `develop` branch. I suggest rebasing your branch to `develop` and dropping all commits that were taken from `main`. You can do this with `git rebase develop` to rebase on develop and `git rebase -i HEAD~n` with `n` the number of last commits you want to consider during the interactive rebasing and then marking the irrelevant commits to drop (`d`).

sebinside:
And, building on Jan's comment, I would suggest to do this in the original PR as my review is there ;)

PR #794, opened 2022-11-15
Fix similarity distribution in JPlagResult

tsaglam:
This PR fixes a bug where the similarity distribution in JPlagResult deviated from the API specification.
The distribution was returned in reverse order (100% to 0% instead of 0% to 100%).

https://github.com/jplag/JPlag/blob/6fc9f77d6bddcde4f1e8ee5c2036597a93692d94/core/src/main/java/de/jplag/JPlagResult.java#L108-L115


JanWittler:
Code overall looks fine for me. 
Why did you choose to reverse the similarity order for the JSON serialization instead of serializing in the "normal" order and simply inverting the displayed order in the report viewer. As a bonus, we could even add fallback for previous versions in the report viewer by inverting the distribution if a version tag of 4.0.0 is detected.

tsaglam:
> did you choose to reverse the similarity order for the JSON serialization instead of serializing in the "normal" order and simply inverting the displayed order in the report viewer?

To save time and not interfere with the report viewer Hiwi, but of course that's a workaround and the long way the report viewer should display it the right way.

> As a bonus, we could even add fallback for previous versions in the report viewer by inverting the distribution if a version tag of 4.0.0 is detected.

I like that, I will add it to the report viewer issue!

PR #809, opened 2022-11-17
Use Maven to load version of JPlag

dfuchss:
Basically, this PR adds a mechanism to load the version based on the current revision of the JPlag project.
(And fixes an issue in the readme: Wrong directory of the build)

### DIFF
--- a/core/src/test/java/de/jplag/reporting/reportobject/ReportObjectFactoryTest.java
+++ b/core/src/test/java/de/jplag/reporting/reportobject/ReportObjectFactoryTest.java
@@ -1,9 +1,9 @@
 package de.jplag.reporting.reportobject;
 
-import de.jplag.reporting.reportobject.model.Version;
-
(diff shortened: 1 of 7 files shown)


core/src/test/java/de/jplag/reporting/reportobject/ReportObjectFactoryTest.java
```diff
@@ -0,0 +1,14 @@
+package de.jplag.reporting.reportobject;
+
+import de.jplag.reporting.reportobject.model.Version;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class ReportObjectFactoryTest {
+    @Test
+    void testVersion() {
```
dfuchss, line 10:
```suggestion
    void testVersionLoading() {
```

tsaglam (approved):
LGTM!

PR #816, opened 2022-11-23
Fix/match display off-by-one error and reenable clustering display

cyfml:
I fixed the bug: The ends of displayed matches are sometimes off by one and re-enable cluster. It is based on that file path problem is  resolved (last PR).
In addition, i added a test case to test cluster.


sebinside:
Just a reminder from the discussion: Please also fix that the starting line is off by one

tsaglam (changes requested):
This PR contains new test resources but no new test case? They are probably from testing and can be removed, right?

cyfml:
> This PR contains new test resources but no new test case? They are probably from testing and can be removed, right?

Yes. After today's classes, i will take a look.

core/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -174,7 +175,7 @@ private void writeOverview(JPlagResult result, String path) {
                 getDate(),// dateOfExecution
                 result.getDuration(), // executionTime
                 getMetrics(result),// metrics
-                List.of()); // clusters (deactivated for now)
+                clusteringResultMapper.map(result)); // clusters (activated for now)
```
sebinside, line 178: Please remove the `(activated for now)` as this information is no longer required

sebinside (commented):
The code looks okay to me, the match display also seems to work correctly now.

However, we noticed a bug in the ReportGenerator (not your fault, this bug probably existed for a long time). Sometimes, the submission IDs in the array of the cluster's members are serialized as `null` which breaks everything. I will try to have a look at it next week. If would be good if you could also try to replicate this behavior. Currently, I don't know why this happens.

sebinside:
We found another bug in the report generation, it was fixed with #842. Please rebase your PR so that we can test it again, afterward, it can be merged.

cyfml:
@sebinside already done.

tsaglam (approved):

sebinside (approved):

PR #819, opened 2022-11-29
Improve EMF metamodel language modules

tsaglam:
Quality of life improvements for the EMF language modules:
- Now also extract end tokens for the handcrafted module, specifically for containment relations 
- Uses Emfatic as a textual model view in the report viewer for both the dynamic and the handcrafted module
- Expands the metamodel test cases

### After this PR: ###
grafik

### Before this PR: ###
grafik


languages/emf-metamodel/src/main/java/de/jplag/emf/util/AbstractModelView.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.IOException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Textual representation of a model for the depiction of matches in submissions.
+ * @author Timur Saglam
+ */
+public class AbstractModelView {
```
JanWittler, line 15: Why is this class specific to models? Looks to me rather like a utility class which is pretty generic for writing any string to a file? Sure there is not already some functionality in the Java API for this?

languages/emf-metamodel/src/main/java/de/jplag/emf/util/EMFUtil.java
```diff
@@ -57,14 +55,13 @@ public static void registerEPackageURIs(Collection<EPackage> ePackages) {
...
+     * @param file is file path to the (meta)model.
+     * @return the resource of the loaded (meta)model or null if it could not be loaded.
      */
-    public static List<EObject> loadModel(File file) {
+    public static Resource loadModelResource(File file) {
         final ResourceSet resourceSet = new ResourceSetImpl();
         try {
-            final Resource resource = resourceSet.getResource(URI.createFileURI(file.getAbsolutePath()), true);
-            return resource.getContents();
+            return resourceSet.getResource(URI.createFileURI(file.getAbsolutePath()), true);
```
JanWittler, line 64:
```suggestion
            return resourceSet.getResource(URI.createFileURI(file.getAbsolutePath()), false);
```

Load on demand should be false here, as the resource is expected to exist

languages/emf-metamodel/src/main/java/de/jplag/emf/util/EmfaticModelView.java
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * Textual view of an EMF metamodel based on Emfatic.
+ * @author Timur Saglam
+ */
+public final class EmfaticModelView extends AbstractModelView {
+    private static final String TYPE_KEYWORD = "(package |class |datatype |enum )";
+    private static final String FEATURE_KEYWORD = "(.*attr .*|op .*|.*ref .*|.*val .*).*";
+    private static final String ANYTHING = ".*";
+    private static final String TYPE_SUFFIX = "(;| extends| \\{)";
```
JanWittler, line 32: add suffix `_REGEX` or similar?

languages/emf-metamodel/src/main/java/de/jplag/emf/util/EmfaticModelView.java
```diff
@@ -0,0 +1,195 @@
...
+
+import de.jplag.Token;
+import de.jplag.emf.MetamodelToken;
+import de.jplag.emf.MetamodelTokenType;
+
+/**
+ * Textual view of an EMF metamodel based on Emfatic.
+ * @author Timur Saglam
+ */
+public final class EmfaticModelView extends AbstractModelView {
```
JanWittler, line 28: This class contains a lot of "magic" to track the objects in the serialization. Maybe the EMFatic parser could be reused to have AST-like information instead of the raw string? Anyways, for the scope of this PR we can leave it as is, but please add more explaining comments about the assumptions you have regarding the syntax.

languages/emf-metamodel/src/main/java/de/jplag/emf/util/EMFUtil.java
tsaglam: Ah, good catch!

tsaglam: It seems the test cases cannot load the resources with this set to false...

JanWittler: ok, seems I misread or misunderstood the documentation 😅

languages/emf-metamodel/src/main/java/de/jplag/emf/util/AbstractModelView.java
JanWittler: I would have expected some method like `FileUtils.write(File, String)` but seems there is none in the stdlib.

languages/emf-metamodel/src/main/java/de/jplag/emf/util/EmfaticModelView.java
```diff
@@ -0,0 +1,194 @@
...
+        return line.matches(TYPE_KEYWORD_REGEX + hash + TYPE_SUFFIX_REGEX + ANYTHING_REGEX);
+    }
+
+    private boolean isEnumLiteral(ENamedElement element, String hash, String line) {
+        return element instanceof EEnumLiteral && line.matches(hash + ANYTHING_REGEX);
+    }
+
+    private boolean isStructuralFeature(ENamedElement element, String hash, String line) {
+        return element instanceof ETypedElement && line.matches(FEATURE_KEYWORD_REGEX + hash + ANYTHING_REGEX);
+    }
```
JanWittler, line 192: Refactor this `+ ANYTHING_REGEX` into a method `boolean prefixMatches(String line, String regex)`

languages/emf-metamodel/src/main/java/de/jplag/emf/util/EmfaticModelView.java
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * Searches for the declaration of an element in the Emfatic code.
+     */
+    private int findLineIndexOf(ENamedElement element) {
+        String hash = Integer.toString(modelCopier.get(element).hashCode());
+        for (int index = 0; index < hashedLines.size(); index++) {
+            String line = hashedLines.get(index);
+            String trimmedLine = line.substring(indentationOf(line));
```
JanWittler, line 167: I would do the indentation removal already when creating the `hashedLines`. As you already get a stream when calling `toLines()`, you can simply add `.map(l -> l.substring(indentationOf(l))` there.

JanWittler (approved):
Just some minor comments to reduce code complexity. Besides that, looks fine.

Issue #710, opened 2022-09-29
Report Viewer Bugfixes and Enhancements


sebinside:
The new report viewer has been introduced with the release [4.0.0](https://github.com/jplag/JPlag/releases/tag/v4.0.0). Although this represents a giant step forward compared to the old string-concatenating HTML-file-generation, there is still some work to do for the next releases of JPlag. This issue shall gather all these TODOs. It replaces #357, #658, and #686.

## Bugs
* [x] Files and matches shall be represented based on their full path (incl. root directory). This shall prevent collisions in the report viewer match display. Please also see the discussions in #686 and #658. The quick fix from #691 has to be therefore reverted. In the viewer, the file paths shall be shown in an abbreviated way if they are too long. File paths shall be platform-independent. (Addressed in #785)
* [x] The report generation does overwrite source code files if they have the same name because all files are "flat mapped", i.e., written in one single folder per submission. It does not throw an error but will put out wrong results that might be noticed later (or even not noticed at all). Here, the solution is probably to preserve the original folder structure also in the resulting zip file in the submissions folder. This could also be based on selecting which files to copy. This problem probably also affects the multi-root and subdir features. (Addressed in #785)
* [x] The ends of displayed matches are sometimes off by one. This is probably a display bug as the information is correctly represented in the generated JSON files. (Addressed in #816)
* [x] Clustering shall be reenabled (disabled in #691) and tested comprehensively (Addressed in #816)
* [x] Match highlighting in the comparison view seems to be broken for single-file submissions (added by @tsaglam)
* [x] Same submission names break the ComparisonView, please see below (Addressed in #829)
* [x] The folder in the results holding the submission files should be renamed from "submissions" to "files" (Addressed in #829)
* [x] Clicking on the colored match highlighting sometimes does not work (or stops working after several clicks), e.g., with [this example](https://github.com/jplag/JPlag/files/10093727/SameSubmissionName.zip), (Addressed in #855)

## Enhancements
* [x] The [main screen](https://jplag.github.io/JPlag/) where the files are dropped should contain a line of text that states that no files get uploaded anywhere (Addressed in #829)
* [x] The code comparison view always shows "submission 1" and "submission 2", but should show the names. But not if the names are hidden (Addressed in #829)
* [x] In the code comparison view, the left sidebar should be visible by default (Addressed in #829)
* [x] In the comparison view, there should be a "back" button (added by @tsaglam), (Addressed in #855)
[...]

PR #829, opened 2022-12-04
Fix/same submission

cyfml:
1. fixed same submission problem which leads Comparision-Viewer broken.
2. fixed issue: The folder in the results holding the submission files should be renamed from "submissions" to "files", and report-viewer also supports old version of Jplag. (e.g. submissions in zip file).
3. Enhancement:  
-  where the files are dropped should contain a line of text that states that no files get uploaded anywhere
-  The code comparison view always shows "submission 1" and "submission 2", but should show the names. But not if the names are hidden.
-  In the code comparison view, the left sidebar should be visible by default


cyfml:
Now, report-viewer supports old version of Jplag. I also find a bug in the previous PR. This bug is: if root name or sub directory is also called "submissions". The Comparison view will not be broken but show not full path, then it influences the highlighting.
Previous one seems a coincidental, now i fixed this problem.

core/src/main/java/de/jplag/reporting/jsonfactory/DirectoryManager.java
```diff
@@ -34,8 +34,7 @@ public static File createDirectory(String path, String name, File file, File sub
         File directory;
         String fileName = file.getPath();
         String submissionRootPath = submissionRoot.getPath();
-        int lastDirectoryIndex = findRootDirIndex(name, submissionRootPath);
-        fileName = fileName.substring(lastDirectoryIndex).replaceFirst(name, "");
+        fileName = fileName.substring(submissionRootPath.length());
```
tsaglam, line 37: If you only want the name, you can fo `file.getName()` instead.

tsaglam (approved):
Just one small thing.

core/src/main/java/de/jplag/reporting/jsonfactory/DirectoryManager.java
cyfml: Actually, it is with a separator, for example, \TerrainType.java or /TerrainType.java.

tsaglam: Then I would suggest `String fileName = File.separator() + file.getName();`. This makes it clearer what is actually required there.

report-viewer/src/components/FilesContainer.vue
```diff
@@ -13,7 +13,9 @@
         :lines="!files.get(file)?.lines ? [] : files.get(file)?.lines"
         :matches="!matches.get(file) ? [] : matches.get(file)"
         :panel-id="containerId"
-        :title="file.length>40?'..'+file.substring(file.length-40,file.length):file"
+        :title="convertSubmissionIdToName(file, submissionId).length > 40?
```
sebinside, line 16: I would suggest moving this to an additional function that calls `convertSubmissionIdToName` with the length check of 40 and so on, to minimize the complexity of this section

sebinside:
Besides this change, I have nothing to complain. I'm not happy with the complexity of the `extractFileNameWithFullPath` function but have no better solution. In my testing, everything worked fine. Can be merged after the small change mentioned above.

cyfml:
@tsaglam @sebinside I suggest, it's better to change the name of PR to "Fix/same submission".  Same root problem was postponed.

sebinside (approved):

sebinside:
Thank you @cyfml, I just merged this PR and updated issue #710. I also added a new strange bug that I discovered while testing (also prior to this PR). Maybe you can also find a solution to this.

PR #830, opened 2022-12-04
Generate Java assign token more often

mbrdl:
In Java, `i++` is syntactic sugar for `i += 1`, which in turn is syntactic sugar for `i = i + 1`. Currently, a `J_ASSIGN` token is only generated in the last case, no tokens are generated in the other two cases. This inconsistency can be resolved by also generating a `J_ASSIGN` token for compound assignments and increment/decrement operations.


tsaglam:
You need to adapt some expected values of test cases for this PR to pass the build. However, do not adapt the E2E tests yet.
For the spotless check, use `mvn spotless:apply` in order to apply formatting.

tsaglam (changes requested):
I have to do some further tests in order to determine the impact of these changes on real datasets. Marked as changes requested to avoid accidental merging.

tsaglam (approved):
Looks good! I tested it thoroughly with real-wold data.

PR #831, opened 2022-12-05
Update spotless formatter configuration file

tsaglam:
Updates the formatter configuration file with a newer eclipse version, thus fixing issues with the formatting of `instanceof` casts.

### DIFF
--- a/cli/src/main/java/de/jplag/cli/CLI.java
+++ b/cli/src/main/java/de/jplag/cli/CLI.java
@@ -36,11 +36,6 @@ import java.util.List;
 import java.util.Random;
 import java.util.stream.Collectors;
 
-import net.sourceforge.argparse4j.ArgumentParsers;
-import net.sourceforge.argparse4j.inf.ArgumentParser;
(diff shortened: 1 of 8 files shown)


JanWittler (approved):
Just for my understanding:
Is the formatter somehow "magically" linked to external rules by the profile's version number? I am confused that some imports are reorganized but I don't see any new rule targeting imports.

tsaglam:
@JanWittler I am not entirely sure why the imports are now ordered differently. However, the first commit is the relevant one. The version of the formatter configuration is the Eclipse version and changes how spotless applies formatting. Only in the newer version the `instanceof` bug was fixed, as only for newer version full java 17 support exists. At least that is my understanding, for more info check `[this issue](https://github.com/diffplug/spotless/issues/1046)`.

The additional rules in the config are added automatically when re-exporting the new config with a recent eclipse version.

tsaglam:
Well, it seems the instanceof problem is not fully fixed. 👎

PR #834, opened 2022-12-06
Overhaul core test cases

tsaglam:
Long overdue clean-up:

- Rename test classes to reflect their purpose
- Split up large test classes
- Move generic test functionality to the test base
- Document test base
- Avoid leftover output files
- Adds useful assertions
- Make assumptions explicit via constants
- Add a single-file submission test
- Re-enable clustering test case
- Adds description tags to test cases


JanWittler:
Did you change any tests or just move them / rename their container? So shall we also check the test method's body or simply whether the overall structure did improve?

core/src/test/java/de/jplag/BasicFunctionalityTest.java
```diff
@@ -87,66 +87,28 @@ void testPartialPlagiarism() throws ExitException {
...
 
-    @Test
-    void testMultiRootDirBasecodeInSubmissionDir() throws ExitException {
-        String basecodePath = getBasePath("basecode", "base");
-        List<String> paths = List.of(getBasePath("basecode"), getBasePath("SimpleDuplicate")); // 2 + 2 submissions.
-        JPlagResult result = runJPlag(paths, it -> it.withBaseCodeSubmissionDirectory(new File(basecodePath)));
-        assertEquals(4, result.getNumberOfSubmissions());
-    }
+        var matches = result.getAllComparisons().get(0).matches();
+        var expectedMatches = runJPlagWithDefaultOptions("SimpleDuplicate").getAllComparisons().get(0).matches();
```
JanWittler, line 103: Hard to see that "SimpleDuplicate" contains the same files as "SimpleSingleFile". Maybe add some explaining comment either to the files themselves or here in code?
Also the method assumes that matches are ordered which is not guaranteed, right? So maybe do some sorting or replace the pairwise iteration by an equalsSize and forEachExists logic.

core/src/test/java/de/jplag/RootFolderTest.java
```diff
@@ -0,0 +1,73 @@
...
+class RootFolderTest extends TestBase {
+
+    private static final String BASECODE_SUBMISSION = "base";
+    private static final String GLOBAL_BASECODE = "basecode-base";
+    private static final String BASE_SAMPLE = "basecode";
+    private static final String SIMPLE_SAMPLE = "SimpleDuplicate";
+
+    @Test
+    void testMultiRootDirNoBasecode() throws ExitException {
+        List<String> paths = List.of(getBasePath(BASE_SAMPLE), getBasePath(SIMPLE_SAMPLE)); // 3 + 2 submissions.
```
JanWittler, line 25: Instead of the comments, why not use an assert /precondition to communicate the number of expected files? This would also help identify incorrect preconditions

core/src/test/java/de/jplag/RootFolderTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Test
+    void testOverlappingNewAndOldDirectoriesOverlap() throws ExitException {
+        List<String> newDirectories = List.of(getBasePath(SIMPLE_SAMPLE)); // 2 submissions
+        List<String> oldDirectories = List.of(getBasePath(SIMPLE_SAMPLE));
+        JPlagResult result = runJPlag(newDirectories, oldDirectories, it -> it);
+        int numberOfExpectedComparison = 1;
+        assertEquals(numberOfExpectedComparison, result.getAllComparisons().size());
+    }
+
```
JanWittler, line 63: Not necessary for this PR, but should overlapping old and new directories not rather throw an error?

core/src/test/java/de/jplag/BasicFunctionalityTest.java
```diff
@@ -3,15 +3,15 @@
...
-import java.util.Optional;
-
 import org.junit.jupiter.api.Test;
 
 import de.jplag.exceptions.ExitException;
 
-class ParallelComparisonTest extends TestBase {
+/**
+ * Test class for the very basic functionality of JPlag.
+ * @author Timur Saglam
```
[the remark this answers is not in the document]
tsaglam: Currently, we have no convention regarding that. In my case, it is just generated by my IDE, but that can be disabled.

core/src/test/java/de/jplag/RootFolderTest.java
tsaglam: These tests were copied from another test class, as tests were split up. I think these comments explain the input scenario (folder/file structure of the sample), but I am not sure.

JanWittler: Would understand them like. While thinking of it, maybe a low overhead refactoring while improving readability a lot would be to extract those numbers into constants (like `BASE_SAMPLE_COUNT`, `SIMPLE_SAMPLE_COUNT`, etc) and computing the expected numbers based on that?

PR #839, opened 2022-12-07
Fix NPE when calling ParsingException constructors with null reason

SirYwell:
A per the Javadocs, `null` values are permitted for the `reason`, however passing `null` or using a constructor that doesn't take a `String` causes a NullPointerException.

I addressed this issue and created a regression test for it.

Note: `String#isBlank` returns true if the string is empty, so this check wasn't needed.

### DIFF
--- a/language-api/src/test/java/de/jplag/ParsingExceptionTest.java
+++ b/language-api/src/test/java/de/jplag/ParsingExceptionTest.java
@@ -6,7 +6,7 @@ import java.io.File;
 
 import org.junit.jupiter.api.Test;
 
-public class ParsingExceptionTest {
+class ParsingExceptionTest {
 
     @Test
     void testParsingExceptionAcceptsNullReason() {
--- a/language-api/src/test/java/de/jplag/ParsingExceptionTest.java
+++ b/language-api/src/test/java/de/jplag/ParsingExceptionTest.java
@@ -1,0 +1,23 @@
+package de.jplag;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
+import java.io.File;
+
+import org.junit.jupiter.api.Test;
+
+public class ParsingExceptionTest {
+
+    @Test
+    void testParsingExceptionAcceptsNullReason() {
+        // placeholder exception to have a non-null argument
+        File file = new File("myFile");
+        // placeholder exception to have a non-null argument
+        Exception exception = new RuntimeException();
+
+        assertDoesNotThrow(() -> new ParsingException(file));
+        assertDoesNotThrow(() -> new ParsingException(file, (String) null));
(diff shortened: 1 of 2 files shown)


tsaglam:
Your branch is based on main, but should be based on develop.

tsaglam (changes requested):
Your branch is based on main but should be based on develop.

SirYwell:
From my understanding of the comment in the pull request template https://github.com/jplag/JPlag/blob/main/.github/pull_request_template.md?plain=1#L3 together with semver (<https://semver.org/#summary>)
> PATCH version when you make backwards compatible bug fixes

I thought main would be the appropriate branch. I can change the branch, but I think the template is confusing then.

language-api/src/test/java/de/jplag/ParsingExceptionTest.java
```diff
@@ -0,0 +1,23 @@
...
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
+import java.io.File;
+
+import org.junit.jupiter.api.Test;
+
+public class ParsingExceptionTest {
+
+    @Test
+    void testParsingExceptionAcceptsNullReason() {
```
JanWittler, line 12: Just a small note but the test method does not match its body because you are testing far more than only a null reason.
`testParsingExceptionConstructor` maybe?

JanWittler (commented):
Just one smaller comment, besides that looks fine

language-api/src/test/java/de/jplag/ParsingExceptionTest.java
SirYwell: I can probably find a better name, but I initially stumbled upon this because I called the `ParsingException(File, Throwable)` constructor. It delegates to the `ParsingException(File, String, Throwable)` constructor with `null` for the string, so the NPE was thrown even when not passing a reason at all. I decided to call all public constructors in the test case to cover internal delegations, but I could also only test the `ParsingException(File, String, Throwable)` constructor instead.

Nullability of the other arguments isn't tested here, so `testParsingExceptionConstructor` might be too generic. Maybe `testParsingExceptionAcceptsNoOrNullReason`?

JanWittler: Ah thanks for the clarification. Actually, I would say in such a case it is sufficient to call the 3 args constructor if for the other constructors internal delegation is performed anyways. But it doesn't harm to have all 4 calls here neither, so then just leave it as is (and also the method name)

JanWittler (approved):

SirYwell:
> I would still merge it into develop, as I am probably not releasing a patch version for this alone (I changed the target initially to develop).

I changed the target branch and rebased the commit onto develop.

tsaglam (approved):

PR #855, opened 2022-12-19
Fix/clicking problem on match highlighting

cyfml:
This PR fixed a new issue: clicking on the colored match highlighting sometimes does not work (or stops working after several clicks). In addition, adding a back button in Comparisons viewer.


sebinside:
Works fine, thank you @cyfml!

PR #856, opened 2022-12-20
Prototype for EMF model instance language module

tsaglam:
This PR adds a language module for parsing EMF models (as in instances of EMF metamodels). In this prototypic implementation, the metamodel (Ecore file) needs to be provided as a submission to parse the metamodel itself. Tokens are extracted dynamically, meaning for every metaclass in the metamodel. This means the token set is very generic and not tailored according to specific kinds of models. As for visualization, Emfatic (which is used for metamodels) does not support models; we thus use a simplified textual tree view.

Additionally, this PR now adds the token normalization mechanism.

Tasks:

- [x] Loading and registering metamodel first
- [x] Parse models and extract tokens dynamically
- [x] Generate simple textual representation (simplified tree view)
- [x] Enrich tokens with tracing information
- [x] Test language module
- [x] Add to CLI
- [ ] Implement a proper way of specifying the metamodels for the model language module
- [ ] Document language module
- [ ] Rename language Interfaces
- [x] Ensure visualization in report viewer works, this recently broke!

Additionally, with #1053 this PR now includes the newly added model normalization.


tsaglam:
For some reason, the coverage calculation is off, even though the test case is run.

language-api/src/main/java/de/jplag/Language.java
```diff
@@ -58,4 +58,21 @@ default boolean useViewFiles() {
...
+    }
+
+    /**
+     * Re-orders the provided submission according the requirements of the language.
+     * @param submissions is the list of submissions.
+     * @return the re-ordered list.
+     */
+    default List<File> customizeSubmissionOrder(List<File> submissions) {
+        return submissions;
+    }
```
JanWittler, line 77: Just an initial comment without having read the entire code:
This is needed such that for the dynamic model instances language the metamodel file is processed first right? I don't think that this is a good solution as the metamodel would not be a submission but rather meta-information necessary to complete the plagiarism check. Thus, mixing it with the submissions introduces technical debt.

I would rather suggest to add some functionality to provide arbitrary metadata necessary for the language module e.g. as key-value input. As an initial sketch this might require some new method like `Language#configureWithMetadata(Map<String,Object>)` to configure the language, maybe some state checking like `Language#isCorrectlyConfigured`, and method(s) to get the CLI description and a streamlined process to provide those metadata as arguments (e.g. something like `-c<key>`, so for the metamodel `-cMetamodel`).

tsaglam: Yes, I agree completely. In the long term, we need language-module-dependent CLI parameters and, thus, an underlying system for configuring the language modules dynamically. The latter is relatively easy to implement. The former is intertwined with many design decisions regarding the CLI. One could make the CLI a two-step dialogue. First, the language is selected, then the resulting language-specific options are shown. These can then be selected to configure the language module, e.g., for metadata or enabling specific defense mechanisms. However, there we need to think about how we want the user to interact with language frontends without custom parameters. Also, we could decide to require a certain CLI argument before the user gets shown the language-specific options. This would then allow using the JPlag CLI in a one-command-to-result-way. If have some more ideas for different ways to solve this, currently I do not have the time for that. Maybe with the new student assistant(s), we could tackle such a feature. For the moment, this is thus a very prototypical and thus, not a very nice solution.

JanWittler: In Java, commonly beta CLI arguments are marked with the `-X` prefix. As such, we might already implement support for configuring languages dynamically and add prototypical support for the metamodel of the dynamic model language frontend using the `-Xmetamodel` parameter. With that, we can already extend the Language frontend with the correct methods instead of these temporary ones while still deferring the choices how users might interact with the CLI.
I would not assume that the model frontend is used that often, thus having incomplete CLI documentation or not the best UX for now seems acceptable to me.

tsaglam:
> @tsaglam coverage should work now. (hopefully)

Ah, TIL that I have to add it there! :D Thanks!

language-api/src/main/java/de/jplag/Language.java
tsaglam: Current plan: One of our upcoming Hiwis will be tasked to design a good solution for this that. I will probably merge this PR before, so we have no branch drift.

tsaglam:
Open issues are documented in #576.

PR #878, opened 2023-01-16
Enhance/version check

cyfml:
This PR enhances the version check and display more than a console warning, also a prompt displaying the warning when the situation is using development versions (0.0.0) or version mismatches . And Report-viewer loads the version from a JSON file.

### DIFF
--- a/report-viewer/src/model/factories/OverviewFactory.ts
+++ b/report-viewer/src/model/factories/OverviewFactory.ts
@@ -8,7 +8,7 @@ import versionJson from "../../version.json";
 
 export class OverviewFactory {
 
-  static reportViewerVersion: Version = JSON.parse(JSON.stringify(versionJson["report-viewer_version"]));
+  static reportViewerVersion: Version = JSON.parse(JSON.stringify(versionJson["report_viewer_version"]));
 
   static getOverview(json: Record<string, unknown>): Overview {
     const versionField = json.jplag_version as Record<string, number>;
--- a/report-viewer/src/model/factories/OverviewFactory.ts
+++ b/report-viewer/src/model/factories/OverviewFactory.ts
@@ -82,10 +82,15 @@ export class OverviewFactory {
   }
 
   static compareVersions(jsonVersion: Version, reportViewerVersion: Version) {
-    if(jsonVersion.major !== reportViewerVersion.major ||
+    if(jsonVersion.major === 0 && jsonVersion.minor === 0 && jsonVersion.patch === 0){
+      window.alert("You are using develop version(0.0.0).");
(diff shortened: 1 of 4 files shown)


report-viewer/src/model/factories/OverviewFactory.ts
```diff
@@ -4,11 +4,11 @@ import { ComparisonListElement } from "../ComparisonListElement";
 import { Cluster } from "@/model/Cluster";
 import store from "@/store/store";
 import { Version } from "../Version";
+import versionJson from "../../version.json";
 
 export class OverviewFactory {
 
-  // TODO: Find a better more visible position to store this
-  static reportViewerVersion: Version = {major: 4, minor: 0, patch: 0};
+  static reportViewerVersion: Version = versionJson["report_viewer_version"] !== undefined ? JSON.parse(JSON.stringify(versionJson["report_viewer_version"])) : {major: -1, minor: -1, patch: -1};
```
sebinside, line 11: Why do you stringify and parse here? Can't you access the properties of the `versionJson["report_viewer_version"]` directly?

report-viewer/src/typings.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.json" {
```
sebinside, line 1: Is this hack really required? A quick google search showed different approaches (like [here](https://stackoverflow.com/questions/46991237/how-to-import-json-file-into-a-typescript-file)) that seam at little less hacky

report-viewer/src/version.json
```diff
@@ -0,0 +1,7 @@
+{
```
sebinside, line 1: I remember that we discussed that this file should contain the version number `0.0.0` so that it always shows the development information. The deployed version will have this file automatically replaced in the github actions pipeline.

sebinside (changes requested):
Here are some minor review comments.

report-viewer/src/typings.d.ts
cyfml: You are right, i try another way and it works too. Just add  ` "resolveJsonModule": true,` in `tsconfig`. This way is much better.

report-viewer/src/version.json
cyfml: @sebinside Ah, i forgot this. So i just need change this number to 0.0.0, right?

report-viewer/src/model/factories/OverviewFactory.ts
cyfml: @sebinside it can access directly right now.

sebinside (approved):

PR #890, opened 2023-01-24
Change documentation of language usage to constructor invocation

SirYwell:
The documentation of how to use a specific language module seems to be outdated. `LanguageOption` does not exist anymore, and the `JPlagOptions` has a `language` component instead.

I updated all readmes from language modules. `LanguageOption` also seems to be referenced in the readme in `endtoend-testing`, however I don't know what the exact replacement would be there, so I left that out.

I'm not sure if you want this to target `develop` or `main`, so let me know if you want me to change it.


tsaglam (approved):
Thanks a lot for noticing! Great PR!

PR #899, opened 2023-01-27
Enhance/warning if some comparisions missing

cyfml:
This PR is based on last unmerged PR (Version check) and fixed problem: Show a warning if some/all comparison files are missing, in the console log and also in the report viewer when scrolling down, e.g., because there are 1000 comparisons but only 100 are included in the resulting zip file. 

Trigger condition: If number of missing comparisons is 0, the warining will not be shown in the report-viewer. Otherwise, it will.

Addationaly, i keep the layout of `FileUploadView` as before, because PR#918 change the layout.

### DIFF
--- a/core/src/main/java/de/jplag/reporting/jsonfactory/ComparisonReportWriter.java
+++ b/core/src/main/java/de/jplag/reporting/jsonfactory/ComparisonReportWriter.java
@@ -48,8 +48,9 @@ public class ComparisonReportWriter {
     public Map<String, Map<String, String>> writeComparisonReports(JPlagResult jPlagResult, String path) {
         int numberOfMaximumComparisons = jPlagResult.getOptions().maximumNumberOfComparisons();
         int numberOfAllComparisons = jPlagResult.getAllComparisons().size();
-        int missingComparisons = numberOfAllComparisons - numberOfMaximumComparisons;
-        logger.warn("Total Comparisons: {}. Shown Comparisons: {}. Missing Comparisons: {}.", numberOfAllComparisons, numberOfMaximumComparisons,
+        int shownComparisons = numberOfAllComparisons > numberOfMaximumComparisons ? numberOfMaximumComparisons : numberOfAllComparisons;
+        int missingComparisons = numberOfAllComparisons > numberOfMaximumComparisons ? (numberOfAllComparisons - numberOfMaximumComparisons) : 0;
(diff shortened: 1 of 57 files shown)


core/src/main/java/de/jplag/reporting/jsonfactory/ComparisonReportWriter.java
```diff
@@ -42,8 +46,12 @@ public ComparisonReportWriter(Function<Submission, String> submissionToIdFunctio
      * yield the same result.
      */
     public Map<String, Map<String, String>> writeComparisonReports(JPlagResult jPlagResult, String path) {
-        int numberOfComparisons = jPlagResult.getOptions().maximumNumberOfComparisons();
-        List<JPlagComparison> comparisons = jPlagResult.getComparisons(numberOfComparisons);
+        int numberOfMaximumComparisons = jPlagResult.getOptions().maximumNumberOfComparisons();
+        int numberOfAllComparisons = jPlagResult.getAllComparisons().size();
+        int missingComparisons = numberOfAllComparisons - numberOfMaximumComparisons;
+        logger.warn("Total Comparisons: {}. Shown Comparisons: {}. Missing Comparisons: {}.", numberOfAllComparisons, numberOfMaximumComparisons,
```
tsaglam, line 52: Here would just use logger.info, as this is where the user just configured the CLI to drop some matches.

core/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -175,7 +178,10 @@ private void writeOverview(JPlagResult result, String path) {
                 getDate(),// dateOfExecution
                 result.getDuration(), // executionTime
                 getMetrics(result),// metrics
-                clusteringResultMapper.map(result)); // clusters
+                clusteringResultMapper.map(result), // clusters
+                totalComparisons, // totalComparisons
+                totalComparisons > numberOfComparisons ? numberOfComparisons : totalComparisons, // shownComparisons
+                totalComparisons > numberOfComparisons ? (totalComparisons - numberOfComparisons) : 0); // missingComparisons
```
tsaglam, line 184: Extract these as variables if possible.

tsaglam (commented):
Just a few small things.

sebinside (approved):

PR #900, opened 2023-01-27
Enhance/color

cyfml:
This PR changed the color scheme of matches.  Some colors (Red and Green) which have implicit semantics are removed.


sebinside (changes requested):
While your approach does the job, nearly 50% of the code is duplicated (for the two intervals). Please define a method that gives you, e.g., an array of n colors of an interval [x,y] and call it twice, for each hue interval.

sebinside (approved):

PR #910, opened 2023-02-05
Report viewer: Allow loading results by URL

jtotht:
When integrating with other software ([TMS](https://gitlab.com/tms-elte) in my case), it’s useful to be able to send users to a URL that will automatically load the results, without having to recompile the report viewer on the live server every time (to make use of overview.json). This pull request introduces a new `file` query parameter, which, if provided, makes the report viewer load the results from the given URL. Similarly to the drag-and-drop upload, the URL can point to both ZIPs and JSONs; since the URL may not have an appropriate extension, let the browser figure out the type based on the HTTP headers. Browsers can figure out the MIME type of drag-and-drop files as well (usually based on the extension), so migrate also the drag-and-drop to MIME-based type determination, dropping the `getFileExtension` function. Also improve error handling: instead of just logging to the browser console, present the error using `alert` – not a five-star UX, but still much better than nothing.

### DIFF
--- a/report-viewer/src/utils/Utils.ts
+++ b/report-viewer/src/utils/Utils.ts
@@ -7,15 +7,3 @@
 export function generateLineCodeLink(index: number, fileName: string, line: number): string {
     return String(index).concat(fileName).concat(String(line))
 }
-
-/**
- * Get the extension of a given file.
- * @param file The file.
- */
-export function getFileExtension(file: File): string {
-    if (file.name.includes(".")) {
-        const split = file.name.split(".")
-        return split[split.length - 1]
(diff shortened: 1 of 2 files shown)


tsaglam:
Thanks for the ideas, we will look into it. We are currently working on improving the error messages of the report viewer (see #710).
If you want access to the newly implemented features early, you can test the development version of the report viewer at https://jplag.github.io/JPlag-Dev/.

sebinside:
It works fine, thank you. Windows uses a different MIME type, I will update that myself.

Also, just a reminder to configure CORS accordingly.

PR #923, opened 2023-02-10
Enhance/add error page if overview or comparison missing

cyfml:
This PR handels an error : report-viewer is broken when overview or some comparisons are missing. I create an `ErrorView` page, which contains `ErrorMessage` and `ErrorRouter` components. 

- Overview missing: error page has a button, which can navigate to `FileUploadView`.

- Some certain comparisons missing:  overview page works continually and can access other comparisons which are not missing. If access missing comparisons, it will navigate to `OverviewView`. 

Note: In comparison missing cases, If you decompress the zip file and delete some comparisons. Then re-compress them. It leads a problem: when you click a match in `MatchTable`, it will not scroll to corresponding position automatically and matched codes have no color. The reason is when you decompress and recompress files, the variable `originalFileName` in `FileUploadView `changes for files under 'files' directory. e.g. From "files\\A\\GSTiling.java" to "files/A/GSTiling.java". So the name of file in `MatchTable` is not same as file header in `codelPanel`. _**If you delete comparisons directly in zip file, there is no problem.**_  This problem only occurs on windows, beacuse It seems zip app uses default path separator '/'. So this zip file are not allowed be decompressed and recompressed on windows. If you want to delete something, please delete it directly in zip file and don't decompress and recompress.


sebinside:
@cyfml Please update this PR on the current state, then I will have look at the changes and review it

cyfml:
@cyfml Done, please review it.

PR #925, opened 2023-02-12
enchance/solve bad performance when clicking metric button

cyfml:
This PR fixed bad performance when clicking metric button. When there are more than thousand comparisons, it becomes very slow. The reason is that, web page needs to re-render thousand of DOM nodes (`<tr>`). Because all comparisons shows in the `ComparisonsTable `. It wastes too much time and mermory. I found it takes 1 second or even serval seconds to load page in Vue dev tool, when there are over 1K comparisons.  This problem should be considered at designing, but not too late.

Actually, there are two solutions;

1. ###  paging: 

show some parts of comparisons in one page. This solution dosen't fit our design. Because scrolling will be more suitable to see the coherence of similarity. From high to low.

2. ###  virtual scroll:  

Virtual scrolling enables a performant way to simulate all items being rendered by making the height of the container element the same as the height of a total number of elements to be rendered, and then only rendering the items in view. So performance is really good, even there are over 10K records. So I choose this way to improve the performance. But we need install this plugin `(vue-virtual-scroller@next)`.

I keep the layout of this `ComparisonsTable ` as before as possible. Because original table layout must be replaced when using this plugin.

Additionally, I also made the route lazy loaded.

### DIFF
--- a/report-viewer/src/components/ComparisonsTable.vue
+++ b/report-viewer/src/components/ComparisonsTable.vue
@@ -2,7 +2,7 @@
   Table which display all of the comparisons with their participating ids and similarity percentage for the selected metric.
 -->
 <template>
-  <table class="table">
+  <table>
     <tr class="head-row">
       <th>No.</th>
       <th>Submission 1</th>
--- a/report-viewer/src/components/ComparisonsTable.vue
+++ b/report-viewer/src/components/ComparisonsTable.vue
@@ -13,6 +13,7 @@
   </table>
 
   <DynamicScroller
+      v-if="topComparisons.length>0"
       class="scroller"
       :items="topComparisons"
       :min-item-size="48"
--- a/report-viewer/src/components/ComparisonsTable.vue
+++ b/report-viewer/src/components/ComparisonsTable.vue
@@ -2,7 +2,7 @@
   Table which display all of the comparisons with their participating ids and similarity percentage for the selected metric.
 -->
 <template>
-  <table>
+  <table class="table">
     <tr class="head-row">
       <th>No.</th>
       <th>Submission 1</th>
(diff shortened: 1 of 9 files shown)


report-viewer/src/components/ComparisonsTable.vue
```diff
@@ -2,121 +2,233 @@
...
+            item.secondSubmissionId
           )
         "
+        class="td5"
       >
-        {{ formattedMatchPercentage(comparison.similarity) }}
+        {{ formattedMatchPercentage(item.similarity) }}
       </td>
-      <td>
+      <td style="width: 5%">
```
sebinside, line 94: Partially including style in the html is a bad smell, please move to the style part

report-viewer/src/components/ComparisonsTable.vue
```diff
@@ -2,121 +2,233 @@
...
             )
           "
-          :comparison="comparison"
+          :comparison="item"
         />
       </GDialog>
-    </tr>
-  </table>
+      </tr>
+      <tr v-else class="selectableOdd">
```
sebinside, line 127: Having the same code twice only because of the highlighting is code duplication. Please use something like the deleted `:nth-child` instead for highlighting and do not duplicate the code.

report-viewer/src/components/ComparisonsTable.vue
```diff
@@ -247,32 +359,50 @@ export default defineComponent({
...
 }
 
-td {
+.td1 {
+  padding-right: 10%;
   padding-top: 3%;
   padding-bottom: 3%;
 }
 
+.td2 {
```
sebinside, line 382: Are these different `td` classes really required? Also, they break the view if the content gets to long, see my example in the comment after the review

report-viewer/src/vue-virtual-scroll.d.ts
```diff
@@ -0,0 +1,21 @@
+// Type definitions for vue-virtual-scroller
```
sebinside, line 1: I understand that this is required, but please add a comment linking to the [original issue](https://github.com/Akryum/vue-virtual-scroller/issues/199).

sebinside (changes requested):
First of all, I like your solution and it greatly improves the performance, thank you! However, I have some comments regarding the code quality. Especially the view breaks if the content of the table cells get to long, which is unfortunately a realistic use case. Nevertheless, looking forward to include this PR!

sebinside:
@cyfml: Here is an example of a result zip file that breaks the new view. It's just the PartialPlagiarism test case with longer (realistic) submission names.
[result.zip](https://github.com/jplag/JPlag/files/10722702/result.zip)

sebinside:
Ah, and of course please also update to the current state of the `develop` branch.

cyfml:
> First of all, I like your solution and it greatly improves the performance, thank you! However, I have some comments regarding the code quality. Especially the view breaks if the content of the table cells get to long, which is unfortunately a realistic use case. Nevertheless, looking forward to include this PR!

@sebinside Thanks, the difficult of this is it doesn't fit for table, it's more suitable for list. Because in table structure, i add the new component `RecycleScroller`, it breaks the structure of table. So the content is not aligned. And this method must use fixed size of table container.

sebinside:
@cyfml What about [DynamicScroller](https://vue-virtual-scroller-demo.netlify.app/dynamic) then? The benefit is described as `The main use case for this is not knowing the height of the items in advance`

cyfml:
@sebinside I am trying.

report-viewer/src/components/ComparisonsTable.vue
cyfml: td1...td6: i need these classes to control them, because each comparison is a table. It can't be aligned automatically, because new component brokes the table structure.

cyfml:
@sebinside All things are done. Please review it and test them.

report-viewer/src/components/ComparisonsTable.vue
cyfml: `:nth-child` works not for this situation. Each comparison is a table, so each comparison only has one `<tr>` element.

sebinside (approved):

sebinside:
Good work, I tested it with 3000 comparisons and did not encounter any problems. Thank you!

cyfml:
@sebinside Thanks. Could you add an item about this problem in issues list?  When i write documentation, i have something to look.

PR #926, opened 2023-02-13
Improve the end-to-end tests by adding summary tests

tsaglam:
This PR adds summary tests to the e2e test suite, thus allowing one to quickly glance at the test result and understand the impact of the changes on all tests in total. The information is presented like this:

```java
org.opentest4j.AssertionFailedError: positive deviation over all AVG similarity values:
	count=42, average=0.0260, min=0.0008, max=0.0833

org.opentest4j.AssertionFailedError: negative deviation over all AVG similarity values:
	count=284, average=0.0906, min=0.0012, max=0.2681
```

This PR also cleans up some code that is touched and improves JavaDoc comments.

### DIFF
--- a/endtoend-testing/src/main/java/de/jplag/endtoend/helper/DeltaSummaryStatistics.java
+++ b/endtoend-testing/src/main/java/de/jplag/endtoend/helper/DeltaSummaryStatistics.java
@@ -1,0 +1,62 @@
+package de.jplag.endtoend.helper;
+
+import java.util.DoubleSummaryStatistics;
+
+/**
+ * Summary statistics when evaluating the delta between two double values. This class manages two
+ * {@link DoubleSummaryStatistics} and delegates the delta of the accepted values to the statistics depending on the
+ * sign of the delta. In contrast to normal {@link DoubleSummaryStatistics}, negative and positive deltas do not cancel
+ * each other out in the average value, as two separate statistics are managed. A delta of zero is not stored.
+ */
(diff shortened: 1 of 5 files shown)


endtoend-testing/src/main/java/de/jplag/endtoend/helper/DeltaSummaryStatistics.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * Calculates the delta of the two double values and delegates it to the statistics depending on the sign of the delta.
+     * A delta of zero is not stored.
+     * @param first is the first double value.
+     * @param second is the second double value.
+     */
+    public void accept(double first, double second) {
+        double delta = first - second;
```
sebinside, line 30: I wonder whether this could lead to false positives due to the floating point nature

tsaglam: The test case that uses this only fails if the average of all deltas is above an epsilon value:
https://github.com/jplag/JPlag/blob/5a2eeaa633dba4a78e97d1ec184cab052b4042d8/endtoend-testing/src/test/java/de/jplag/endtoend/EndToEndSuiteTest.java#L170-L172
This test case is mainly for feedback. For precision, the individual test cases (which produce the deltas) will fail if they deviate by the expected value (again, with a certain epsilon).

PR #940, opened 2023-02-17
Fix two minor issues with the Java language module token extraction

tsaglam:
This PR fixes ~~three~~ two issues with the token extraction:

1. Try and try-with-resource were extracted as different tokens
2. Try-catch-finally blocks did not correctly employ end tokens
3. ~~Initialization and subsequent declaration of a variable in a single line vs. two line lead to different tokens~~

EDIT: Removed the last fix, as it affected the false positives enough while not affecting the true-positives significantly

This PR does not affect the E2E tests, as try/catch-blocks are not part of the dataset used there.


tsaglam:
@mbrdl @SirYwell just FYI.

PR #951, opened 2023-02-27
Enhance/page cannot be reloaded after refreshing

cyfml:
This PR creates a 404 html, because SPA change pages by JS and it has only one page(index.html). All other urls are not accessed through github-pages, because it doesn't exist. So this 404 page will redirect to public path.

### DIFF
--- a/report-viewer/public/404.html
+++ b/report-viewer/public/404.html
@@ -1,0 +1,6 @@
+<!doctype html>
+<html lang="">
+<head>
+    <meta http-equiv="refresh" content="0;URL=https://jplag.github.io/JPlag/">
+</head>
+</html>
--- a/report-viewer/public/404.html
+++ b/report-viewer/public/404.html
@@ -1,6 +1,7 @@
 <!doctype html>
 <html lang="">
 <head>
+    <!--Redirect to entrance-->
     <meta http-equiv="refresh" content="0;URL=./">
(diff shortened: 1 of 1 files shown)


sebinside (changes requested):
Please use relative redirection as an absolute path will break the dev version of Jplag deployed at https://jplag.github.io/JPlag-Dev/

cyfml:
> Please userelative redirection as an absolute path will break the dev version of Jplag deployed at https://jplag.github.io/JPlag-Dev/

@sebinside I try to redirect to index.html directly, but "index.html" will be in the URL and nothing will be shown.  Let me try to find anthor way.

cyfml:
@sebinside Now, i change it in relative path.

sebinside:
I tested this solution with your deployment at https://cyfml.github.io/JPlag/. While is worked well for reloading on the overview page, it infinitely redirects when reloading a comparison page

cyfml:
> I tested this solution with your deployment at https://cyfml.github.io/JPlag/. While is worked well for reloading on the overview page, it infinitely redirects when reloading a comparison page

I find this problem. Because we have two urls, so i decide to use scripts to generate url dynamiclly. You can test it under https://cyfml.github.io/JPlag/ again.

sebinside (approved):

PR #971, opened 2023-03-13
Correctly show matches when the token order is not line index order

tsaglam:
In some cases, the first and last tokens of a match are not the first or last elements of the match in the source code. This happens e.g. when:

- view files are used, as in the EMF language module
- the code is normalized before parsing for obfuscation detection
- in rare edge cases, even in normal language modules, e.g., regarding if/else/end-if/try/catch, etc.

In these cases, JPlag was not able to correctly display the matches in the report viewers. This PR fixes this by comparing the line indices of all tokens of a match.

### DIFF
--- a/core/src/main/java/de/jplag/Match.java
+++ b/core/src/main/java/de/jplag/Match.java
@@ -2,8 +2,8 @@ package de.jplag;
 
 /**
  * Represents two sections of two submissions that are similar.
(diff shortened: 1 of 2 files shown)


tsaglam:
@smjonas this _might_ help with the issues of showing matches for model instances.

sebinside (approved):

PR #982, opened 2023-03-16
fix/match-highlighting-problem

cyfml:
I fixed this problem. Now match highlighting is correct. The reason is, the file path should also be converted into ID. But i find another problem. The size of row is not adaptive. But for other example, it works. It needs investigation. I am busy right now, so if i have free time, i will take a look at it.

NOT ADAPTIVE: image

ADAPTIVE: image


Medo3G:
> I fixed this problem. Now match highlighting is correct. The reason is, the file path should also be converted into ID. But i find another problem. The size of row is not adaptive. But for other example, it works. It needs investigation. I am busy right now, so if i have free time, i will take a look at it.
> 
> NOT ADAPTIVE: image
> 
> ADAPTIVE: image

This can be fixed by simply adding `overflow-wrap: break-word;` to https://github.com/jplag/JPlag/blob/284206ce766e5175cf4aeeddbded63494a105d0e/report-viewer/src/components/ComparisonsTable.vue#L291

tsaglam (approved):

PR #1005, opened 2023-04-02
Speed up comparison

mbrdl:
In the greedy string tiling comparison algorithm, certain tokens are marked. The marks are currently kept track of through sets of indices. By instead using boolean arrays containing whether each token is marked, JPlag's runtime is roughly cut in half.

### DIFF
--- a/core/src/main/java/de/jplag/GreedyStringTiling.java
+++ b/core/src/main/java/de/jplag/GreedyStringTiling.java
@@ -8,8 +8,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


tsaglam (approved):

mbrdl:
Some more measurements, each row being 100 runs:
Screenshot 2023-04-14 at 18 23 33
color = task, dark = without comparison speedup, light = with comparison speedup

I tested some other variations but they don't really make a difference. What does make a difference is the comparison speedup, the speedup factor of 2 I gave may even be a bit of an underestimate. What's noticeable is that for the measurements without the speedup, the standard deviation is much bigger. This gives us a hint toward the reason for the speedup. I think it has to do with caching, maybe not all of the set entries fit into the data cache and they need to be loaded again individually on every lookup. The array is contiguous in memory and can be loaded much much faster. @tsaglam in case you're interested

PR #1007, opened 2023-04-02
Add token string normalization

mbrdl:
This adds the framework for token string normalization. It also generates semantic information and adds it to the tokens in the Java language module. This cannot be disabled. I couldn't measure a difference in runtime, however.


core/src/main/java/de/jplag/JPlag.java
```diff
@@ -59,6 +59,10 @@ public JPlagResult run() throws ExitException {
             throw new SubmissionException("Not enough valid submissions! (found " + submissionCount + " valid submissions)");
         }
 
+        // better solution long-term: pull this into SubmissionSet constructor, use language-specific option
+        // if (language.tokensHaveSemantics())
+        // submissionSet.normalizeSubmissions();
+
```
dfuchss, line 65: Maybe move that to an issue instead of inline comment :)

core/src/main/java/de/jplag/Submission.java
```diff
@@ -265,4 +268,24 @@ private static File createErrorDirectory(String... subdirectoryNames) {
         }
         return true;
     }
+
+    void normalize() {
+        // System.out.println();
+        // System.out.println("original: " + getOrder(tokenList));
+        tokenList = new NormalizationGraph(tokenList).linearize();
+        // System.out.println("normalized: " + getOrder(tokenList));
```
dfuchss, line 276: Remove comments or better add logger (debug)

core/src/main/java/de/jplag/normalization/NormalizationGraph.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.List;
+import java.util.PriorityQueue;
+import java.util.stream.Collectors;
+
+import org.jgrapht.Graphs;
+import org.jgrapht.graph.SimpleDirectedGraph;
+
+import de.jplag.Token;
+
+public class NormalizationGraph {
```
dfuchss, line 15: Add documentation to new public classes (and other classes if needed)

core/src/main/java/de/jplag/normalization/NormalizationGraph.java
```diff
@@ -0,0 +1,78 @@
...
+    private SimpleDirectedGraph<Statement, Edge> graph;
+
+    /**
+     * Construct a new normalization graph from the tokens.
+     * @param tokens The tokens used to construct the normalization graph.
+     */
+    public NormalizationGraph(List<Token> tokens) {
+        graph = new NormalizationGraphConstructor(tokens).get();
```
dfuchss, line 23: The pattern new(X).getX() looks like you want a static method that does this job instead of creating an object that is dropped afterwards

languages/java/src/main/java/de/jplag/java/Parser.java
```diff
@@ -23,14 +22,27 @@ public Parser() {
     public List<Token> parse(Set<File> files) throws ParsingException {
         tokens = new ArrayList<>();
         new JavacAdapter().parseFiles(files, this);
+        // printSemantics();
```
dfuchss, line 25: same as above -> comments

dfuchss (changes requested):
Some initial comments on the code :)

core/src/main/java/de/jplag/normalization/NormalizationGraph.java
mbrdl: Great suggestion, cleans the code up a lot 👍 I didn't do this intitially because of concerns about object orientation but it really doesn't make much sense now that I think about it

mbrdl: Okay, I actually changed my mind. This is good for the normalization itself because there is no state associated with it. But for the normalization graph construction shown here, there is a ton of state necessary. Making all of it static seems really bad. I can't really think of better method, happy to take suggestions

mbrdl:
I did some more measurements and the difference in runtime came out at about 6%. So negligible I would say

core/src/main/java/de/jplag/normalization/Edge.java
```diff
@@ -0,0 +1,33 @@
...
+    private boolean isVariableFlow;
+    private boolean isVariableReverseFlow;
+
+    Edge() {
+        items = new HashSet<>();
+        isVariableFlow = false;
+    }
+
+    boolean isVariableFlow() {
+        return isVariableFlow;
```
smjonas, line 19: You could consider just storing `EdgeType type` as an instance variable in `addItem` (instead of separate variables for `isVariableFlow` and `isVariableReverseFlow`), then change this to
```suggestion
        return type == EdgeType.VARIABLE_FLOW
```

core/src/main/java/de/jplag/normalization/EdgeType.java
```diff
@@ -0,0 +1,29 @@
+package de.jplag.normalization;
+
+/**
+ * Enum for types of edges in normalization graph. Given two statements A and B, A comes before B, there is such an edge
+ * between A and B if...
+ */
+enum EdgeType {
+    /**
+     * B reads from a variable A writes.
```
smjonas, line 9:
```suggestion
     * B reads from a variable A writes to.
```
(makes the statement clearer imo, same applies to the other comments in the class)

core/src/main/java/de/jplag/normalization/NormalizationGraphConstructor.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    /**
+     * Adds an ingoing edge to the current node.
+     * @param start the start of the edge
+     * @param type the type of the edge
+     * @param cause the variable that caused the edge, may be null
+     */
+    private void addIngoingEdgeToCurrent(Statement start, EdgeType type, Variable cause) {
```
smjonas, line 117:
```suggestion
    private void addIncomingEdgeToCurrent(Statement start, EdgeType type, Variable cause) {
```

core/src/main/java/de/jplag/normalization/TokenStringNormalizer.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Queue;
+import java.util.stream.Collectors;
+
+import org.jgrapht.Graphs;
+import org.jgrapht.graph.SimpleDirectedGraph;
+
+import de.jplag.Token;
+
+/**
+ * The class that performs token string normalization.
```
smjonas, line 15:
```suggestion
 * Performs token string normalization.
```

core/src/main/java/de/jplag/normalization/TokenStringNormalizer.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public class TokenStringNormalizer {
+
+    private TokenStringNormalizer() {
+    }
+
+    /**
+     * Normalizes the token string it receives inplace. Tokens representing dead code have been eliminated and tokens
```
smjonas, line 23:
```suggestion
     * Normalizes the token string it receives in place. Tokens representing dead code have been eliminated and tokens
```

language-api/src/main/java/de/jplag/semantics/CodeSemantics.java
```diff
@@ -0,0 +1,211 @@
+package de.jplag.semantics;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * This class contains semantic information about a code snippet, in our case either a token or a statement.
```
smjonas, line 10:
```suggestion
 * Contains semantic information about a code snippet, in our case either a token or a statement.
```

core/src/main/java/de/jplag/normalization/Edge.java
mbrdl: Edge is actually a multiple edge, so it can have multiple types, so this doesn't really work. Bad naming on my part, I've renamed `Edge` to `MultipleEdge` and `EdgeItem` to `Edge` to better reflect what's going on

core/src/main/java/de/jplag/normalization/EdgeType.java
mbrdl: It is inconsistent, but I've gone the opposite way to fix it, it's now just `B reads a variable A writes`. More concise and consistent with my thesis

mbrdl: Good suggestions all around though, thank you!

core/src/main/java/de/jplag/Submission.java
```diff
@@ -83,11 +87,11 @@ public int compareTo(Submission other) {
     }
 
     @Override
-    public boolean equals(Object obj) {
```
dfuchss, line 86: you should not rename parameter names of overridden methods.

core/src/main/java/de/jplag/Submission.java
```diff
@@ -265,4 +269,29 @@ private static File createErrorDirectory(String... subdirectoryNames) {
...
+        List<Integer> originalOrder = getOrder(tokenList);
+        TokenStringNormalizer.normalize(tokenList);
+        List<Integer> normalizedOrder = getOrder(tokenList);
+
+        logger.debug("original line order: {}", originalOrder);
+        logger.debug("line order after normalization: {}", normalizedOrder);
+        Set<Integer> normalizedSet = new HashSet<>(normalizedOrder);
+        List<Integer> removed = originalOrder.stream().filter(l -> !normalizedSet.contains(l)).toList();
+        logger.debug("removed {} line(s): {}", removed.size(), removed);
+    }
```
dfuchss, line 283: Add a JavaDoc here. It's not clear what the normalization will do :)

core/src/main/java/de/jplag/Submission.java
```diff
@@ -265,4 +269,29 @@ private static File createErrorDirectory(String... subdirectoryNames) {
...
+
+    private List<Integer> getOrder(List<Token> tokenList) {
+        List<Integer> order = new ArrayList<>(tokenList.size());  // a little too big
```
dfuchss, line 286: Does this increase the performance ? Otherwise it's not needed to define the size

core/src/main/java/de/jplag/normalization/MultipleEdge.java
```diff
@@ -0,0 +1,33 @@
...
+
+    boolean isVariableReverseFlow() {
+        return isVariableReverseFlow;
+    }
+
+    void addEdge(EdgeType type, Variable cause) {
+        if (type == EdgeType.VARIABLE_FLOW)
+            isVariableFlow = true;
+        if (type == EdgeType.VARIABLE_REVERSE_FLOW)
+            isVariableReverseFlow = true;
```
dfuchss, line 30: maybe use new switch-case-statement to ensure completeness if enum will be extended

core/src/main/java/de/jplag/normalization/NormalizationGraphConstructor.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.jgrapht.graph.SimpleDirectedGraph;
+
+import de.jplag.Token;
+import de.jplag.semantics.Variable;
+
+class NormalizationGraphConstructor {
```
dfuchss, line 16: You should also document internal classes.

dfuchss (changes requested):
Some further comments. Code in general lgtm

core/src/main/java/de/jplag/Submission.java
mbrdl: I haven't tested it but I would assume so.

tsaglam (approved):
Looks good! Just one conflict to solve, @mbrdl, then I can merge.

core/src/main/java/de/jplag/normalization/MultipleEdge.java
mbrdl: This doesn't make sense in this context, there already are edge types where nothing happens here.

dfuchss:
Why did you revert the switch? f3f061e94d7baad5032f2812bcdf3e12c0cd48d7

It should enforce that you have to define behavior for all enum values. (If you don't use default)

mbrdl:
> Why did you revert the switch? [f3f061e](https://github.com/jplag/JPlag/commit/f3f061e94d7baad5032f2812bcdf3e12c0cd48d7)
> 
> It should enforce that you have to define behavior for all enum values. (If you don't use default)

But the default is doing nothing? I don't understand, maybe you could provide a code snippet to show me what you mean.

dfuchss:
> > Why did you revert the switch? [f3f061e](https://github.com/jplag/JPlag/commit/f3f061e94d7baad5032f2812bcdf3e12c0cd48d7)
> > 
> > It should enforce that you have to define behavior for all enum values. (If you don't use default)
> 
> But the default is doing nothing? I don't understand, maybe you could provide a code snippet to show me what you mean. I reverted the switch because Sonar didn't like it.

I can do that later. The idea would be to use the new switch without default . Then you have to define all cases (therefore, it will cause a compile error if new enum values will be added)

dfuchss:
> > Why did you revert the switch? [f3f061e](https://github.com/jplag/JPlag/commit/f3f061e94d7baad5032f2812bcdf3e12c0cd48d7)
> > 
> > It should enforce that you have to define behavior for all enum values. (If you don't use default)
> 
> But the default is doing nothing? I don't understand, maybe you could provide a code snippet to show me what you mean. I reverted the switch because Sonar didn't like it.

I can do that later. The idea would be to use the new switch without default . Then you have to define all cases (therefore, it will cause a compile error if new enum values will be added)

mbrdl:
> > > Why did you revert the switch? [f3f061e](https://github.com/jplag/JPlag/commit/f3f061e94d7baad5032f2812bcdf3e12c0cd48d7)
> > > It should enforce that you have to define behavior for all enum values. (If you don't use default)
> > 
> > 
> > But the default is doing nothing? I don't understand, maybe you could provide a code snippet to show me what you mean. I reverted the switch because Sonar didn't like it.
> 
> I can do that later. The idea would be to use the new switch without default . Then you have to define all cases (therefore, it will cause a compile error if new enum values will be added)

From my understanding, you only have to define all cases in a switch expression, not in a switch statement. 
```java
        switch (type) {
            case VARIABLE_FLOW -> isVariableFlow = true;
            case VARIABLE_REVERSE_FLOW -> isVariableReverseFlow = true;
        }
```
compiles just fine.

dfuchss:
> > > > Why did you revert the switch? [f3f061e](https://github.com/jplag/JPlag/commit/f3f061e94d7baad5032f2812bcdf3e12c0cd48d7)
> > > > It should enforce that you have to define behavior for all enum values. (If you don't use default)
> > > 
> > > 
> > > But the default is doing nothing? I don't understand, maybe you could provide a code snippet to show me what you mean. I reverted the switch because Sonar didn't like it.
> > 
> > I can do that later. The idea would be to use the new switch without default . Then you have to define all cases (therefore, it will cause a compile error if new enum values will be added)
> 
> From my understanding, you only have to define all cases in a switch expression, not in a switch statement. 
> ```java
>         switch (type) {
>             case VARIABLE_FLOW -> isVariableFlow = true;
>             case VARIABLE_REVERSE_FLOW -> isVariableReverseFlow = true;
>         }
> ```
> compiles just fine, even though more types exist.

I have to check that. There was some kind of switch that enforced that all types are mentioned. As far as I remember.

dfuchss:
E.g., https://www.happycoders.eu/java/switch-expression/

-> Completeness Analysis

dfuchss:
But I've recognized in the case where I've marked the code, this does not make sense. So just ignore my initial comment there :)

dfuchss (approved):

dfuchss:
My problem was that I did not realized that you set two different variables based on the enum value.

mbrdl:
Now that I've turned in my thesis, I've decided to add some tests before the presentation on 5/5. Could be good to have them in this PR so we can see the coverage, otherwise would be nice to have this PR out of the way. What do you think @tsaglam

tsaglam:
> Now that I've turned in my thesis, I've decided to add some tests before the presentation on 5/5. Could be good to have them in this PR so we can see the coverage, on the other hand it would be nice to have this PR out of the way. What do you think

Both is fine with me, whatever you prefer goes. Merging this PR now has the benefit of avoiding conflicts when other PRs are merged.

Issue #267, opened 2022-01-11
Enhance public appearance of JPlag


sebinside:
After the maintenance team changed from [IPD](http://www.ipd.uka.de/) to [KASTEL](https://www.kastel.kit.edu/), we chose to drop the legacy website of JPlag that was not up to date at all. However, JPlag shall have at least a landing page and some documentation to get interested people started. This issue serves as a placeholder to discuss this, the realization will probably start around Q2 2022.

The current approach is to make it lightweight and markdown-friendly, e.g., by using [Docosaurus](https://docusaurus.io/) together with GitHub pages and migrating the current [wiki](https://github.com/jplag/JPlag/wiki).

Possible features (please add!):
* Landing page with a quick overview and download links (direct, GitHub release, maven) and quick help
* Wiki (migrated and enhanced, e.g., drawing a real AST 😮)
* Sample reports and downloadable testing data
* A "used by" section? Do we have enough validated information?
* Credits

An archived version of the old Jplag website can be found [here](https://web.archive.org/web/20130516050657/http://jplag.ipd.kit.edu/).

dfuchss:
@sebinside you might also have a look at jekyll
* https://jekyllrb.com/docs/pages/
* https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/about-github-pages-and-jekyll

sebinside:
@dfuchss Yes, i know about jekyll, but finding really good templates is surprisingly hard :)

dfuchss:
Maybe also have a look at sphinx. This works good as well

dfuchss:
I've created a proposal in #470 :D

tsaglam:
Note that moving away from markdown has a few consequences. First of all, there is still some documentation for the clustering and the report viewer that is currently not part of the wiki. They would need to be converted. Second of all, we currently do diagrams via mermaid, meaning either we manually convert all diagrams from the missing documentation or we find a way to still do mermaid in Sphinx. Also: Moving away from markdown introduces an overhead for future contributors, as alternatives are not as well known.

tsaglam:
Current plan is to deploy Wiki directly from the repo, see:

- https://github.com/kit-sdq/programming-lecture-eclipse-artemis/blob/main/.github/workflows/docs.yml
- https://github.com/kit-sdq/programming-lecture-eclipse-artemis/tree/main/docs
- https://github.com/kit-sdq/programming-lecture-eclipse-artemis/wiki

PR #1017, opened 2023-04-05
Documentation Workflow

dfuchss:
This PR adds the data from Wiki to a `docs` folder that will be deployed upon push to main or if a commit is tagged.

See #267

⚠️ The docs folder will always override all contents of the wiki ⚠️

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,36 @@
+name: Documentation
+
+on:
+  push:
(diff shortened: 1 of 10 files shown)


tsaglam:
@jplag/studdev just FYI, after this PR is merged, you can update the Wiki (if required) in the same PR where you change the codebase itself.

tsaglam (approved):

PR #1026, opened 2023-04-12
Feature/file encoding

### LINKED ISSUE #746: Issues with file encoding
JPlag is currently built for UTF-8 encoding and may reject submissions that have improper encoding (e.g. the Java language module and ANSI files). Maybe we can come up with a more convenient system that is more tolerant.

- First priority is to find and test a solution for the java language.
- After that, we could think of language-independent solutions.

We could use heuristics like https://github.com/albfernandez/juniversalchardet, for an example, see:

https://github.com/Feuermagier/autograder/blob/1a47ee7ea91d3d2f7563053528124fd18c2c037f/autograder-core/src/main/java/de/firemage/autograder/core/file/UploadedFile.java#L41-L56

TwoOfTwelve:
fixes #746

Uses https://unicode-org.github.io/icu/userguide/conversion/detection.html to detect the file encoding.


Kr0nox (approved):
Looks good to me

language-api/src/main/java/de/jplag/util/FileUtils.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.stream.Collectors;
+
+import com.ibm.icu.text.CharsetDetector;
+import com.ibm.icu.text.CharsetMatch;
+
+/**
+ * Encapsulates various interactions with files to prevent issues with file encodings.
+ */
+public class FileUtils {
+    private static final Charset defaultOutputCharset = StandardCharsets.UTF_8;
```
dfuchss, line 17: Constants should be named like this ..
```suggestion
    private static final Charset DEFAULT_OUTPUT_CHARSET = StandardCharsets.UTF_8;
```

language-api/src/main/java/de/jplag/util/FileUtils.java
```diff
@@ -0,0 +1,141 @@
...
+            for (CharsetMatch match : matches) {
+                charsetValues.putIfAbsent(match.getName(), new ArrayList<>());
+                charsetValues.get(match.getName()).add(match.getConfidence());
+                remaining.remove(match.getName());
+            }
+            remaining.forEach(it -> {
+                charsetValues.putIfAbsent(it, new ArrayList<>());
+                charsetValues.get(it).add(0);
+            });
+        });
```
dfuchss, line 100: I'd suggest not using lambdas here (esp. lambda blocks). Simply use normal for loops. that could increase readability :)

language-api/src/main/java/de/jplag/util/FileUtils.java
```diff
@@ -0,0 +1,141 @@
...
+
+        AtomicReference<Charset> mostProbable = new AtomicReference<>(StandardCharsets.UTF_8);
+        AtomicReference<Double> mostProbableConfidence = new AtomicReference<>((double) 0);
```
dfuchss, line 103:
```suggestion
        AtomicReference<Double> mostProbableConfidence = new AtomicReference<>(0.0);
```

dfuchss (changes requested):
Code looks good to me. Just some minor remarks

language-api/src/main/java/de/jplag/util/FileUtils.java
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Removes the byte order mark from the beginning of the stream, if it exists and the charset is a UTF* charset. For
+     * details see: <a href="https://en.wikipedia.org/wiki/Byte_order_mark">Wikipedia</a>
+     * @param reader The reader to remove the bom from
+     * @throws IOException If an IO error occurs.
+     */
+    private static void removeBom(BufferedReader reader, Charset charset) throws IOException {
+        if (charset.name().toUpperCase().startsWith("UTF")) {
+            reader.mark(10);
+            if (reader.read() != '\uFEFF') {
```
tsaglam, line 56: Use constants here.

tsaglam (changes requested):
Just one small thing.

tsaglam (approved):

language-api/src/main/java/de/jplag/util/FileUtils.java
```diff
@@ -0,0 +1,154 @@
...
+
+import com.ibm.icu.text.CharsetDetector;
+import com.ibm.icu.text.CharsetMatch;
+
+/**
+ * Encapsulates various interactions with files to prevent issues with file encodings.
+ */
+public class FileUtils {
+    private static final Charset DEFAULT_OUTPUT_CHARSET = StandardCharsets.UTF_8;
+    private static final char BOM = '\uFEFF';
```
dfuchss, line 20:
```suggestion
    private static final char BYTE_ORDER_MARK = '\uFEFF';
```

dfuchss (approved):
Just one minor suggestion

PR #1029, opened 2023-04-14
Move Report Viewer to vite

Kr0nox:
This PR moves the Report Viewer to vite. This includes

- Movint to vite
- Replacing vuex with pinia
- Removing legacy node dependencies
- Adding husky for pre-commit linting
- Adding new testing templates
- Cleaning up the package.json


Kr0nox:
@sebinside Could you test the report viewer for any bugs and issues with a few larger and real samples?

sebinside (changes requested):
First of all, good work, looks promising and is already near to being merged. I had a look at your changes and also tested the report viewer with real-world data consisting of 20,000 comparisons. Here are my findings:

✅ Report Viewer behaves as expected for common operations
✅ Using the zip drag n drop method works in development
✅ Using the zip drag n drop method works after build using `vite preview`
❌ Using the `src/files` method with the extracted content of the zip file works neither in dev nor after build. The XMLHttpRequest gets printed to the console together with the warning `Unable to build comparison file.` This behavior remains when using a different local server like [npm http-server](https://www.npmjs.com/package/http-server).

Please have a look at this remaining bug. Afterward, we can merge this PR. Thank you!

sebinside (approved):

PR #1036, opened 2023-04-20
File total tokens added to report output

jazerix:
This PR adds the count of tokens in each file such that you get an idea of how much of each file is covered by a match. This is especially nice if you are not using the report viewing but ingesting the files in other ways. 

Alternatively, this metric can also be converted to a similarity using the same formula found in `JPlagComparison.java`

I'm aware that it doesn't give you the complete picture, but having a sense of perspective on each file is nice.

Let me know what you think 😃


tsaglam:
That is a good idea, I think we could also display this information in the report viewer.

jazerix:
@tsaglam Certainly! I've also renamed the two new variables to align with the current naming scheme.

core/src/main/java/de/jplag/reporting/jsonfactory/ComparisonReportWriter.java
```diff
@@ -103,9 +103,13 @@ private Match convertMatchToReportMatch(JPlagComparison comparison, de.jplag.Mat
         Token startOfSecond = tokensSecond.stream().min(lineComparator).orElseThrow();
         Token endOfSecond = tokensSecond.stream().max(lineComparator).orElseThrow();
 
+        List<Token> firstTotalTokens = tokensFirst.stream().filter(x -> x.getFile() != null && x.getFile().equals(startOfFirst.getFile())).toList();
```
dfuchss, line 106: You could use Objects.equals to omit the null check

dfuchss (commented):
One minor suggestion

core/src/main/java/de/jplag/reporting/jsonfactory/ComparisonReportWriter.java
jazerix: I think that's a good idea while we wait for Java to implement a null-conditional operator 😄. I've updated my branch.

dfuchss (approved):

tsaglam:
@TwoOfTwelve could you take a look at this PR? Maybe even test if it affects the performance if we run it for a large dataset?

PR #1040, opened 2023-04-25
Remove redundant passing of language

mbrdl:
Language is contained in the options, passing it separately is entirely pointless

### DIFF
--- a/core/src/main/java/de/jplag/SubmissionSetBuilder.java
+++ b/core/src/main/java/de/jplag/SubmissionSetBuilder.java
@@ -31,9 +31,7 @@ public class SubmissionSetBuilder {
 
     private static final Logger logger = LoggerFactory.getLogger(SubmissionSetBuilder.class);
 
-    private final Language language;
     private final JPlagOptions options;
-    private final Set<String> excludedFileNames; // Set of file names to be excluded in comparison.
 
     /**
      * Creates a builder for submission sets.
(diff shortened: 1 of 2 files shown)


core/src/main/java/de/jplag/SubmissionSetBuilder.java
```diff
@@ -31,19 +31,14 @@ public class SubmissionSetBuilder {
...
     /**
      * Creates a builder for submission sets.
-     * @param language is the language of the submissions.
      * @param options are the configured options.
      */
-    public SubmissionSetBuilder(Language language, JPlagOptions options) {
-        this.language = language;
+    public SubmissionSetBuilder(JPlagOptions options) {
         this.options = options;
-        this.excludedFileNames = options.excludedFiles();
```
dfuchss, line 46: JPlag has to ensure backward compatibility. Therefore, you could mark the old constructor as deprecated and just pass the option to a new constructor. But we cannot remove the constructor directly.

dfuchss (changes requested):
In general. the PR looks good to me. One remaining issue needs to be addressed.

core/src/main/java/de/jplag/SubmissionSetBuilder.java
mbrdl: Is the SubmissionSetBuilder part of the interface? I though it was entirely internal, thus making backwards compatability unnecessary

dfuchss: It's a public class in the core module. Therefore, I think we have to consider it as part of the interface

mbrdl:
I'll just make this PR part of #1041

PR #1041, opened 2023-04-25
Add static JPlag::run

mbrdl:
The current idiom to invoke JPlag is `new JPlag(options).run()`. There is really no reason to construct a JPlag instance as it is discarded immediately after `run()`. For this reason, I would argue that `JPlag::run` should be static and the idiom, correspondingly, should be `JPlag.run(options)`. This is much cleaner and better conveys what is happening. This change breaks the programming API, so it would have to be part of a major release.


mbrdl:
Right, that's elegant. Done

core/src/main/java/de/jplag/SubmissionSetBuilder.java
```diff
@@ -31,19 +31,23 @@ public class SubmissionSetBuilder {
...
-    private final Language language;
     private final JPlagOptions options;
-    private final Set<String> excludedFileNames; // Set of file names to be excluded in comparison.
 
     /**
      * Creates a builder for submission sets.
      * @param language is the language of the submissions.
      * @param options are the configured options.
      */
     public SubmissionSetBuilder(Language language, JPlagOptions options) {
```
dfuchss, line 41: Put an @ Deprecated to the old constructors :)

core/src/main/java/de/jplag/JPlag.java
```diff
@@ -30,42 +30,44 @@ private static Version loadVersion() {
...
-        comparisonStrategy = new ParallelComparisonStrategy(options, coreAlgorithm);
     }
 
     /**
      * Main procedure, executes the comparison of source code submissions.
      * @return the results of the comparison, specifically the submissions whose similarity exceeds a set threshold.
-     * @throws ExitException if the JPlag exits preemptively.
+     * @throws ExitException if JPlag exits preemptively.
      */
     public JPlagResult run() throws ExitException {
```
dfuchss, line 46: Same here. Add @ Deprecated .. the users shall use the new methods.

dfuchss (commented):
Just two comments.

mbrdl:
Hang on, need to fix this

mbrdl:
It's failing because of the `info` code smell that the methods should be removed someday 🙃 I don't know why `info` code smells affect the rating but regardless, it's ready to merge

dfuchss (approved):

core/src/main/java/de/jplag/JPlag.java
```diff
@@ -30,42 +30,48 @@ private static Version loadVersion() {
 
     private final JPlagOptions options;
 
-    private final Language language;
-    private final ComparisonStrategy comparisonStrategy;
-
     /**
      * Creates and initializes a JPlag instance, parameterized by a set of options.
+     * @deprecated in favor of static JPlag::run.
```
dfuchss, line 35: Probably you could use @ link in the javadoc :)

core/src/main/java/de/jplag/JPlag.java
```diff
@@ -30,42 +30,48 @@ private static Version loadVersion() {
...
     public JPlag(JPlagOptions options) {
         this.options = options;
-        language = this.options.language();
-        GreedyStringTiling coreAlgorithm = new GreedyStringTiling(options);
-        comparisonStrategy = new ParallelComparisonStrategy(options, coreAlgorithm);
     }
 
     /**
      * Main procedure, executes the comparison of source code submissions.
+     * @deprecated in favor of static JPlag::run.
```
dfuchss, line 45: See above

dfuchss (approved):

tsaglam (approved):

Issue #920, opened 2023-02-09
Language-specific CLI arguments


tsaglam:
We need a system that allows for language-module-specific CLI arguments.

- e.g. for the modeling languages
- e.g. for defense mechanisms against obfuscation attacks.

This system shall work via key-value pairs and a CLI argument for these pairs, e.g. `-o key=value -o key2=value2`.
The help text should show the available options for the selected language.

tsaglam:
Closed by #1064.

Issue #591, opened 2022-08-15
CLI enhancements


tsaglam:
Currently, the CLI is a victim of feature creep. The more features are introduced, the more complex the CLI becomes.
This makes the CLI confusing and hard to use. Moreover, we also have CLI arguments that arguably will not be used anyways.
With the release of v4.0.0 we have the chance to remove anything we do not need. These options can still be available via the Java API.

**Ideas** (feel free to add or comment below):

- [x] Identify and remove unnecessary CLI arguments (e.g. do we both need `-n` and `-m`?)
- [x] Reduce the clustering options to a minimum
- [x] Introduce more verbose forms of argument flags in addition to the short ones (e.g. add `--min-match-length` for `-t`)

**Relates to**:

- #550 
- #572

tsaglam:
I played around with the clustering a bit and found out which CLI arguments definitely need to be kept:

- `--cluster-skip `
- `--cluster-alg`
- `--cluster-metric`

tsaglam:
Currently we are happy with the CLI, thus closed.

Issue #550, opened 2022-08-01
ArgParse4J might be unmaintained


tsaglam:
[ArgParse4J](https://github.com/argparse4j/argparse4j) is currently unmaintained (see this [PR](https://github.com/argparse4j/argparse4j/pull/139)). We should keep an eye on that and might consider using an alternative in the future. For now, this issue serves to document this problem, and no action is required.

sebinside:
Probably related to the incoming CLI rework (do we already have an issue for that?) and also to #572.

dfuchss:
I think the rework of the CLI includes a discussion about the actual parameters that are needed as well.

We could use something like:
[https://github.com/Kotlin/kotlinx-cli](kotlinx-cli) is incubator but makes it very easy to create CLIs (and maintained by jetbrains).

Example also [here](https://github.com/LiSSA-Approach/LiSSA-Core/blob/main/cli/src/main/kotlin/edu/kit/kastel/lissa/cli/CommandLineInterface.kt)

tsaglam:
> do we already have an issue for that?

Now we have one: #591

dfuchss:
@tsaglam may be closed

tsaglam:
Closed by #1064.

Issue #179, opened 2021-08-17
Add CLI test cases


tsaglam:
We should add test cases that ensure the functionality of the command-line interface and the CLI arguments.
Currently, we can only test that manually, which is tedious as it requires building a jar after every change.

These CLI arguments should be tested first:
- [x] root directory (no flag)
- [x] language `-l`
- [x] base code `-bc`
- [x] min token match `-t` (sensitivity)
- [x] similarity threshold `-m`
- [x] stored matches `-n`
- [x] comparison mode `-c`

Later we should also test:
- [x] result folder `-r`
- [x] verbosity `-v` (replaced by --log-level)
- [x] debug `-d`
- [x] subdirectory `-S`
- [x] suffixes `-p`
- [x] exclude file `-x`

tsaglam:
Closed by #1933.

PR #1064, opened 2023-05-03
Implemented new cli, based on a different framework

TwoOfTwelve:
Implementation for language specific cli arguments. Also contains a rewrite for the cli module using a different, still maintained library.

Adresses #920, #591 and #550.

The new cli should support all options, that the previous version had. Due to the cli being implemented using a new library, it might behave a little different.

The language can now be defined as a subcommand ('jplag java' instead of jplag -l java'). This allows language specific parameters to be set.

Every option now has a long name.

Short options with more than one character (-new or -bc) are still supported, but should be discouraged as this is not intuitive.


cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -83,153 +56,156 @@ public final class CLI {
...
+        }).toList();
     }
 
-    /**
-     * Parses an array of argument strings.
-     * @param arguments is the array to parse.
-     * @return the parsed arguments in a {@link Namespace} format.
-     */
-    public Namespace parseArguments(String[] arguments) {
+    public ParseResult parseOptions(String... args) {
```
Kr0nox, line 107: JavaDoc of public methods

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -83,153 +56,156 @@ public final class CLI {
...
             System.exit(1);
         }
     }
 
-    private static void finalizeLogger() {
-        ILoggerFactory factory = LoggerFactory.getILoggerFactory();
-        if (!(factory instanceof CollectedLoggerFactory collectedLoggerFactory))
-            return;
-        collectedLoggerFactory.finalizeInstances();
+    public CLI() {
```
Kr0nox, line 81: JavaDoc of public methods

cli/src/main/java/de/jplag/cli/LanguageCandidates.java
```diff
@@ -0,0 +1,12 @@
+package de.jplag.cli;
+
+import java.util.ArrayList;
+
+/**
+ * Helper class for picocli to find all available languages.
+ */
+public class LanguageCandidates extends ArrayList<String> {
+    public LanguageCandidates() {
```
Kr0nox, line 9: JavaDoc of public methods

Kr0nox (changes requested):
Just three missing Documentations for public methods.

In addition tho: You removed the constants (enum in this case) for the for the CLI parameters. I would put to discussing keeping them in some kind of way, so that they can be reused in the tests (where now they are also just strings) and so in case they may be required somewhere in the future they are uniform across JPlag

Kr0nox:
I agree with your point about refering to the variables instead of the cli argument.

> Also the tests for the cli module are rather sparse currently, so maybe we should create a separate issue to make them more useful.

#179

language-api/src/main/java/de/jplag/options/LanguageOptionImpl.java
```diff
@@ -0,0 +1,58 @@
+package de.jplag.options;
+
+/**
+ * Default implementation for {@link LanguageOption}
+ * @param <T> The type of the option
+ */
+public class LanguageOptionImpl<T> implements LanguageOption<T> {
```
tsaglam, line 7: Why not DefaultLanguageOption or something else? I dont like the impl abbreviation, as it is not a very concrete name with that suffic?

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -83,153 +56,164 @@ public final class CLI {
...
-        } catch (ArgumentParserException exception) {
-            parser.handleError(exception);
+            ParseResult result = commandLine.parseArgs(args);
+            if (result.isUsageHelpRequested() || (result.subcommand() != null && result.subcommand().isUsageHelpRequested())) {
+                commandLine.getExecutionStrategy().execute(result);
+            }
+            return result;
+        } catch (CommandLine.PicocliException e) {
+            logger.error("Error during parsing", e);
             System.exit(1);
```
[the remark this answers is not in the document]
tsaglam: Ah, well, always interesting to see what we previously missed in PR reviews! It would be great if you could improve this/check if ExitExceptions are viable.

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -83,153 +56,164 @@ public final class CLI {
...
         }
+
         return null;
     }
 
+    private static void finalizeLogger() {
+        ILoggerFactory factory = LoggerFactory.getILoggerFactory();
+        if (!(factory instanceof CollectedLoggerFactory collectedLoggerFactory))
```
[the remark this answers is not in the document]
tsaglam: It is our fault that we did not enforce it until now. Maybe you can keep an eye out for those if you change classes. I will see what I can do to enforce it.

cli/src/main/java/de/jplag/cli/CommandLineArgument.java
```diff
@@ -1,268 +0,0 @@
...
-import net.sourceforge.argparse4j.inf.ArgumentContainer;
-import net.sourceforge.argparse4j.inf.ArgumentParser;
-import net.sourceforge.argparse4j.inf.FeatureControl;
-import net.sourceforge.argparse4j.inf.Namespace;
-
-/**
- * Command line arguments for the JPlag CLI. Each argument is defined through an enumeral.
- * @author Timur Saglam
- */
-public enum CommandLineArgument {
```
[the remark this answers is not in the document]
tsaglam: I understand the arguments you make. I would say let us leave the CLI options as is, and I will see how it behaves maintenance-wise.

For the tests, just introduce constants in the test classes to duplicate among the test cases themselves. I do not want the test to use constants of the code under test, I want the tests to share constants.

cli/src/main/java/de/jplag/cli/CliOptions.java
```diff
@@ -0,0 +1,142 @@
...
+import de.jplag.clustering.algorithm.InterClusterSimilarity;
+import de.jplag.options.JPlagOptions;
+import de.jplag.options.SimilarityMetric;
+
+import picocli.CommandLine;
+import picocli.CommandLine.ArgGroup;
+import picocli.CommandLine.Option;
+import picocli.CommandLine.Parameters;
+
+@CommandLine.Command(name = "jplag", description = "", usageHelpAutoWidth = true, abbreviateSynopsis = true)
```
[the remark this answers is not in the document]
tsaglam: I meant examples like --bc and -bc, where the keyword is identical, but the only difference is the dashes. If we do not go the enum route, leave it as is for now. With the enum route, this can be cleanly deduplicated. However, these are fine details I would say.

tsaglam (approved):

Kr0nox (approved):

tsaglam:
@TwoOfTwelve, we probably need to document the changed/additional CLI parameters in the readme + docs, right?

tsaglam:
> I changed the description and the README. I don't know that much about GitHub. Can I update the wiki for a specific branch or should I just update the general Wiki?

We have a specific action that builds the wiki from the docs folder (since #1017 ). So just modify the docs in the branch of this PR.

README.md
```diff
@@ -64,42 +64,98 @@ JPlag can either be used via the CLI or directly via its Java API. For more info
...
-  --cluster-alg {AGGLOMERATIVE,SPECTRAL}
-                         Which clustering algorithm to use. Agglomerative  merges similar submissions bottom up.
-                         Spectral clustering is  combined  with  Bayesian  Optimization  to  execute the k-Means
-                         clustering  algorithm  multiple   times,   hopefully   finding   a   "good"  clustering
+In the new version the language can either be set with the -l parameter or as a subcommand.
+As a subcommand language specific arguments can be set. As of now, there are no such arguments. They will be added to this description.
+
+Usage: jplag [OPTIONS] [root-dirs[,root-dirs...]...] [COMMAND]
+
+      [root-dirs[,root-dirs...]...]
```
Kr0nox, line 72: This text is indented more then the options below. In my opinio this breaks the style and looks wierd.

README.md
```diff
@@ -64,42 +64,98 @@ JPlag can either be used via the CLI or directly via its Java API. For more info
...
+                       Root-directory with submissions to check for plagiarism
+
+      -bc, --bc, --base-code=<baseCode>
+                       Path of  the  directory  containing  the  base  code
+                         (common  framework  used  in  all submissions)
+
+  -h, --help           display this help and exit
+  -l, --language=<language>
```
Kr0nox, line 80: should we make it explicit that languages have the same names as the commands?

Kr0nox, line 80: does this option work if we give a language as an subcommand? If not this should be documented somewhere

docs/1.-How-to-Use-JPlag.md
```diff
@@ -7,42 +7,98 @@ Example: `java -jar jplag.jar path/to/the/submissions`
...
-                         be saved (default: 0.0)
-
-Clustering:
-  --cluster-skip         Skips the clustering (default: false)
-  --cluster-alg {AGGLOMERATIVE,SPECTRAL}
-                         Which clustering algorithm to use. Agglomerative  merges similar submissions bottom up.
-                         Spectral clustering is  combined  with  Bayesian  Optimization  to  execute the k-Means
-                         clustering  algorithm  multiple   times,   hopefully   finding   a   "good"  clustering
+In the new version the language can either be set with the -l parameter or as a subcommand.
+As a subcommand language specific arguments can be set. As of now, there are no such arguments. They will be added to this description.
```
Kr0nox, line 11: Same comments as above apply

Kr0nox (changes requested):
A few inconsistencies and possible missunderstandings I found

README.md
```diff
@@ -63,43 +63,101 @@ JPlag can either be used via the CLI or directly via its Java API. For more info
 ### CLI
 *Note that the [legacy CLI](https://github.com/jplag/jplag/blob/legacy/README.md) is varying slightly.*
 
+In the new version the language can either be set with the -l parameter or as a subcommand. If both a subcommand and the -l option are specified, the subcommand will take priority.
```
tsaglam, line 66: This sentence starts oddly. There is only one version when this PR is merged into dev/main.
Just start the sentence with "The language...".

README.md
```diff
@@ -63,43 +63,101 @@ JPlag can either be used via the CLI or directly via its Java API. For more info
...
+As a subcommand language specific arguments can be set. As of now, there are no such arguments. They will be added to this description.
```
tsaglam, line 67: When using the subcommand...
Also, remove the second sentence.

docs/1.-How-to-Use-JPlag.md
```diff
@@ -5,44 +5,102 @@ JPlag can be used via the Command Line Interface by executing the JAR file.
 
 Example: `java -jar jplag.jar path/to/the/submissions`
 
+In the new version the language can either be set with the -l parameter or as a subcommand. If both a subcommand and the -l option are specified, the subcommand will take priority.
+As a subcommand language specific arguments can be set. As of now, there are no such arguments. They will be added to this description.
```
tsaglam, line 9: Same as above.

tsaglam (commented):
Just a few minor things.

docs/4.-Adding-New-Languages.md
```diff
@@ -385,6 +385,45 @@ if (<condition>) {
...
+import de.jplag.options.LanguageOptions;  
+import de.jplag.options.OptionType;  
+  
+public class ExampleLanguageOptions extends LanguageOptions {  
+    final LanguageOption<String\> testOption = createOption(OptionType.string(), "testOption");  
+    final LanguageOption<Integer\> sensitivity = createDefaultOption(OptionType.integer(), "sensitivity", 3);  
+}
+```
+
+Then you need to publisch it in your language:
```
Kr0nox, line 403: small spelling mistake. Its publish

docs/4.-Adding-New-Languages.md
```diff
@@ -385,6 +385,45 @@ if (<condition>) {
...
+
+```java
+public class Language implements de.jplag.Language {  
+    [...]
+    private ExampleLanguageOptions options = new ExampleLanguageOptions();  
+
+    [...]
+
+    @Override  
+    public ExampleLanguageOptions getOptions() {  // Make sure to set your subclass as the return type. Otherwise it will not be useful for api access.
```
Kr0nox, line 413: Should this be `ExampleLanguageOptions` or `LanguageOptions`?

Kr0nox (changes requested):

Kr0nox:
And in line 19 of the README it  says:
> line using the `-l <cli argument name>`
Should this be changed or at least mention the subcommand system?

Kr0nox (approved):

language-api/src/main/java/de/jplag/options/EmptyOptions.java
```diff
@@ -0,0 +1,15 @@
+package de.jplag.options;
+
+/**
+ * Dummy implementation for {@link LanguageOptions}
+ */
+public class EmptyOptions extends LanguageOptions {
+    /**
+     * Static instance, since multiple instances don't make sense.
+     */
+    public static final EmptyOptions instance = new EmptyOptions();
```
dfuchss, line 10: I'd put the instance to LanguageOptions

cli/src/main/java/de/jplag/cli/CliOptions.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @ArgGroup(validate = false, heading = "Clustering%n")
+    public Clustering clustering = new Clustering();
+
+    /**
+     * Empty run method, so picocli prints help automatically
+     */
+    @Override
+    public void run() {
+    }
```
tsaglam, line 66: If you move the comment into the method as a line comment, sonar will stop complaining.

mbrdl:
@TwoOfTwelve I don't really understand how to use this, could you explain? How do I recover a language-specific option from a `JPlagOptions` object? How do I pass a language-specific option in the first place, is it any different from an option that is not language-specific? Can I set flags instead of using key-value pairs?

mbrdl:
So the language options are completely separate from the JPlagOptions? Why?

```java
Language lang = new Language(); //referring to the java language here
lang.getOptions().maxTokens.setValue(4);
//now call jplag with lang as your language
```
Doesn't the typing prevent this? `lang.getOptions()` has type `LanguageOptions`, which doesn't have a maxTokens field.

Is there a way to have shared language options? I want to implement a normalization option which many languages could have in the future.

mbrdl:
For my use case, I've added a `getOptionsAsMap` method:
```java
/**
  * @return all options, mapped by their name
  */
 public Map<String, LanguageOption<?>> getOptionsAsMap() {
     return options.stream().collect(Collectors.toUnmodifiableMap(LanguageOption::getName, o -> o));
 }
```
which I then use like this:
```java
if (Boolean.TRUE.equals(options.language().getOptions().getOptionsAsMap().get("normalize").getValue()))
    submissions.forEach(Submission::normalize);
```
but this isn't really a great solution because I have to get the option by name and the typing is completely lost. Anyways, thanks for your help.

Also FYI `-h` doesn't show all possible flags with the language subcommands, I'm not sure this is intentional:
Screenshot 2023-06-21 at 19 11 51

mbrdl:
But I can still use all of the other flags, no? To me this is confusing but I guess it is intended behavior then, in which case nevermind

PR #1067, opened 2023-05-06
Print the number of tokens when it’s too little

jtotht:
When trying to prepare a sample submission for [TMS](https://gitlab.com/tms-elte), I was hit by the minimum token count: the submission was rejected due to not containing the required number of tokens; however, the output stated neither the number of tokens found nor the number of tokens required. So I added them to the output, and I share the code in the hope that it will be useful for others as well.


tsaglam (approved):

jtotht:
Thanks for the quick merge!

PR #1070, opened 2023-05-09
Changed the ClusteringFactory to only do something, if there are any …

TwoOfTwelve:
There was an error, if all comparisons were beneath the threshold (-m in cli). Fixed by not executing clustering, if there are no comparisons left.

### DIFF
--- a/core/src/main/java/de/jplag/clustering/ClusteringFactory.java
+++ b/core/src/main/java/de/jplag/clustering/ClusteringFactory.java
@@ -31,35 +31,39 @@ public class ClusteringFactory {
     private static final Logger logger = LoggerFactory.getLogger(ClusteringFactory.class);
 
     public static List<ClusteringResult<Submission>> getClusterings(Collection<JPlagComparison> comparisons, ClusteringOptions options) {
-        if (!options.enabled()) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/de/jplag/clustering/ClusteringFactory.java
```diff
@@ -31,35 +31,39 @@ public class ClusteringFactory {
     private static final Logger logger = LoggerFactory.getLogger(ClusteringFactory.class);
 
     public static List<ClusteringResult<Submission>> getClusterings(Collection<JPlagComparison> comparisons, ClusteringOptions options) {
-        if (!options.enabled()) {
-            logger.warn(CLUSTERING_DISABLED);
-            return Collections.emptyList();
-        } else {
-            logger.info(CLUSTERING_PARAMETERS, options.algorithm(), options.preprocessor());
-        }
+        if (comparisons.isEmpty()) {
```
tsaglam, line 34: Is that correct? Shouldn't that be `!empty`?

tsaglam (commented):
Just one issue. Also, we could add a small unit test for the empty case.

tsaglam (approved):

PR #1081, opened 2023-05-14
Fix supported C# version in README

jtotht:
The C# Language class states it supports C# 6, the README previously stated C# 8 is supported. Expression-bodied not read-only properties ([supported as of C# 7.0](https://github.com/dotnet/docs/blob/c802a0c56bccdc87908365c0788b3c5ecacea2b3/docs/csharp/programming-guide/statements-expressions-operators/expression-bodied-members.md)) like this don’t parse:

```cs
public int foo { get => 4; }
```

which means the Language class is right and C# 8 isn’t fully supported. Since the latest version is 11, and the first unsupported version (C# 7) was released over six years ago, also reclassify C# support as “legacy”.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ In the following, a list of all supported languages with their supported languag
 |--------------------------------------------------------|--------:|-------------------|:----------------------------------------------------------------:|:---------:|
 | [Java](https://www.java.com)                           |      17 | java              |                              mature                              |   JavaC   |
 | [C++](https://isocpp.org)                              |      11 | cpp               |                              legacy                              |  JavaCC   |
-| [C#](https://docs.microsoft.com/en-us/dotnet/csharp/)  |       8 | csharp            |                               beta                               |  ANTLR 4  |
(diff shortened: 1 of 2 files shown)


tsaglam (changes requested):
You are right; the ANTLR grammar only [supports C# 6](https://github.com/antlr/grammars-v4/tree/master/csharp).
However, the legacy tag is intended for language modules from the [legacy](https://github.com/jplag/JPlag/tree/legacy) era of JPlag, which is pre-v3.0.0. We do not know how well the tokenization is implemented for these language modules and thus do not know how well it performs for plagiarism detection. So I would say leave it as beta. Let me know if you know any newer ANTLR grammar for C#.

jtotht:
I’ve updated the PR. However, according to the [criteria in the wiki](../wiki/2.-Supported-Languages), it counts as `legacy`, as “This module is old and may only support outdated language versions. It needs an update.” Please update the criteria as well. (And no, unfortunately I don’t know newer grammars for C#.)

tsaglam (approved):

tsaglam:
> Thanks! I didn’t know the canonical version of the wiki is in the repo, nice!

We have not had that for long, but it is really nice. Our CI pushes the docs subdirectory to the internal wiki repo.

PR #1100, opened 2023-05-22
Feature/new test structure for language modules

TwoOfTwelve:
Expanded the language testutils to simplify test code. Test classes should extend LanguageModuleTest. For an example look at the Kotlin language module tests.

The LanguageModuleTest class can be configured to execute the following tests:

- All source lines covered by at least one token
- All tokens from the language found in source
- A set of tokens contained in the source
- Match the source file against a specific sequence of tokens

An addition to that the following tests are automatically executed for all tests:

- The lines reported by each token are monotonous ascending
- The last token reported in the source is the file end token


Kr0nox:
I would suggest adding JavaDoc to all Methods from LanguageModuleTest, so it is easier to see what each method does.

In addiditon I would suggest updating the wiki, so the section on adding a new Language also contains information on how to implement these tests

language-testutils/src/test/java/de/jplag/testutils/LanguageModuleTest.java
```diff
@@ -0,0 +1,273 @@
...
+     */
+    private <T, C extends Collection<T>> C addNullGuard(C data) {
+        if (data.size() == 0) {
+            data.add(null);
+
+            logger.info("No tests configured. Creating a default null test, to fulfill JUnits requirements.");
+        }
+
+        return data;
+    }
```
tsaglam, line 252: From the comments, I cannot understand why we use this guard. Why do we need it?

language-testutils/src/test/java/de/jplag/testutils/datacollector/TestDataCollector.java
```diff
@@ -0,0 +1,168 @@
...
+     */
+    public List<TokenListTest> getTokenSequenceTest() {
+        return tokenSequenceTest;
+    }
+
+    /**
+     * @return The list of all test data
+     */
+    public Set<TestData> getAllTestData() {
+        return allTestData;
```
tsaglam, line 92: Here as well.

language-testutils/src/test/java/de/jplag/testutils/datacollector/TestData.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+import de.jplag.Language;
+import de.jplag.ParsingException;
+import de.jplag.Token;
+
+/**
+ * Any kind of test source provider
```
tsaglam, line 11: As this is a top-level interface, this could be more descriptive. A reader could ask: What kind of data is it? A set of submissions? A submission? A file? Code? XML?

tsaglam (changes requested):
Looks good. Just a few things to discuss.

language-testutils/src/test/java/de/jplag/testutils/LanguageModuleTest.java
tsaglam: In which cases do we even need 0 executions? Don't we always have at least one?

tsaglam: I see. But this then means there is always one test shown as executed per test type, even though there actually is none, right?

tsaglam: e.g. if we add the line
```java
testContainers.add(DynamicContainer.dynamicContainer("TEST", List.of()));
```
after line 88 in `de.jplag.endtoend.EndToEndSuiteTest` we get the following:

image

PR #1133, opened 2023-06-12
Improve log readability and parse error reporting

tsaglam:
Logging improvements:
- Removes the thread id from the log message (causes a lot of noise during the parallel comparison)
- When submissions cannot be parsed, warn the user at the end of the JPlag run (as previous errors tend to be missed due to the large log)
- When the debug mode is used, also state where the erroneous submissions are copied to
- Improve formatting of logged durations
- Do not expose internal comparison strategy via log (as it is just more visual noise)
- Remove superfluous "start comparing" message

### DIFF
--- a/core/src/main/java/de/jplag/JPlag.java
+++ b/core/src/main/java/de/jplag/JPlag.java
@@ -1,5 +1,7 @@
 package de.jplag;
 
+import java.io.File;
+import java.util.List;
 import java.util.ResourceBundle;
 
 import org.slf4j.Logger;
@@ -72,6 +74,19 @@ public class JPlag {
         if (logger.isInfoEnabled())
             logger.info("Total time for comparing submissions: {}", TimeUtil.formatDuration(result.getDuration()));
         result.setClusteringResult(ClusteringFactory.getClusterings(result.getAllComparisons(), options.clusteringOptions()));
+
+        logSkippedSubmissions(submissionSet, options);
+
         return result;
+    }
+
+    private static void logSkippedSubmissions(SubmissionSet submissionSet, JPlagOptions options) {
+        if (submissionSet.getInvalidSubmissions().size() > 0) {
+            List<Submission> skipSet = submissionSet.getInvalidSubmissions();
+            logger.warn("{} submissions were skipped (see errors above): {}", skipSet.size(), skipSet.stream().map(Submission::getName).toList());
+            if (options.debugParser()) {
+                logger.warn("Erroneous submissions were copied to {}", new File("errors").getAbsolutePath());
+            }
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


core/src/main/java/de/jplag/JPlag.java
```diff
@@ -72,6 +74,19 @@ public static JPlagResult run(JPlagOptions options) throws ExitException {
...
             logger.info("Total time for comparing submissions: {}", TimeUtil.formatDuration(result.getDuration()));
         result.setClusteringResult(ClusteringFactory.getClusterings(result.getAllComparisons(), options.clusteringOptions()));
+
+        logSkippedSubmissions(submissionSet, options);
+
         return result;
     }
+
+    private static void logSkippedSubmissions(SubmissionSet submissionSet, JPlagOptions options) {
+        List<Submission> skipSet = submissionSet.getInvalidSubmissions();
```
Kr0nox, line 84: In my opinion the name `skipSet` is not ideal, since it is a List and not a Set

core/src/main/java/de/jplag/TimeUtil.java
```diff
@@ -12,10 +14,8 @@ private TimeUtil() {
      * @return Readable representation of the time interval.
      */
     public static String formatDuration(long durationInMilliseconds) {
-        int timeInSeconds = (int) (durationInMilliseconds / 1000);
-        String hours = (timeInSeconds / 3600 > 0) ? (timeInSeconds / 3600) + " h " : "";
-        String minutes = (timeInSeconds / 60 > 0) ? ((timeInSeconds / 60) % 60) + " min " : "";
-        String seconds = (timeInSeconds % 60) + " sec";
-        return hours + minutes + seconds;
+        Duration duration = Duration.ofMillis(durationInMilliseconds);
+        return String.format("%dh %02dmin %02ds %03dms", duration.toHours(), duration.toMinutesPart(), duration.toSecondsPart(),
```
Kr0nox, line 18: This will always output `0h ...` if  the execution took less than an hour. The same with other time units. The old code did not do that. Is this the desired behavior?

Kr0nox (changes requested):
Looks good. Just two small things I found

core/src/main/java/de/jplag/TimeUtil.java
tsaglam: Yes, the new representation is chosen to be more consistent.

core/src/main/java/de/jplag/JPlag.java
```diff
@@ -72,6 +74,19 @@ public static JPlagResult run(JPlagOptions options) throws ExitException {
...
+        if (!skipSet.isEmpty()) {
+            logger.warn("{} submissions were skipped (see errors above): {}", skipSet.size(), skipSet.stream().map(Submission::getName).toList());
+            if (options.debugParser()) {
+                logger.warn("Erroneous submissions were copied to {}", new File("errors").getAbsolutePath());
```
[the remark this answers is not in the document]
tsaglam: Good point. The constants could also be moved to the options.

Kr0nox (approved):

Issue #983, opened 2023-03-17
Language module architecture


tsaglam:
The Language module architecture still has several problems. This issue is meant to document all known problems. Successor of #454. 
Solving these problems may introduce API-breaking changes; thus, the solutions require a major release.

- [ ] Language modules should provide readme files that explain the design rationales of the token extraction
- [ ] An abstract parser class should offer shared functionality among frontends
- [ ] The parser should be instantiated per submissions, not globally per language
- [ ] The language facades should not be named the same (API-breaking change!!!)
- [ ] Language modules currently all read in files on their own; see #746 
- [ ] #576 
- [ ] #920

tsaglam:
@jplag/maintainer & @jplag/studdev: Feel free to comment if you have some points to add to this list.

dfuchss:
> We could also add a module with common code for testing, as most languages implement the same stuff for their tests.

What about the language testutils module?

tsaglam:
> I totally missed that. I still think, it would be worth to expand it, to standardize the behavior of language module tests.

Everything that we can standardize or generalize, we should. This is why the testutils were created. If you have any ideas for an extended testing framework for language modules we can discuss that. Would be great to make things easy for new modules.

libre-man:
Right now all languages parse entire files, this makes parsing of files that contain multiple languages (such as `.vue` files and `.ipynb`) files very difficult, as we need to create temporary files.

PR #1159, opened 2023-06-27
Feature/antlr language utils

TwoOfTwelve:
Adds a new maven module, containing a common parser implementation for all antlr languages.
Part of #983


tsaglam:
@TwoOfTwelve, you need to adapt the coverage pom in order for us to see the coverage correctly:
https://github.com/jplag/JPlag/blob/3986f58f2f355f00cac3e87a6e16f82183ad530b/coverage-report/pom.xml

tsaglam (approved):
Looks good. Maybe we can see with the typescript module how intuitive the system is (@Kr0nox just FYI) and then tweak it if required.

Kr0nox:
@TwoOfTwelve Is there a way two tokens for the same RuleContext get added in the desired order?
```
this.createRangeMapping(FunctionDeclarationContext.class, ASSIGNMENT);
this.createStartStopMapping(FunctionDeclarationContext.class, METHOD_BEGIN, METHOD_END);
```
In this case `ASSIGNMENT` always gets added after `Method_BEGIN`. This is also the case if these two calls are swapped

Kr0nox:
Yes that solves the issue 👍

PR #1167, opened 2023-07-04
Feature/refactor cpp2 language module

TwoOfTwelve:
This PR refactors the cpp2 tests and language module.
Also contains minor improvements for the abstract antlr utils and test utils.

The refactoring of the cpp2 module fixes a couple of issues with the module. As far as I can tell all tokens are still extracted by the same rules, although it would be nice if someone could verify that.

The fixed errors are:
-Nested if-else statements lead to the wrong line being assigned to the second else token
-The Antlr grammar was missing DivAssign in the list of operators for operator overloading

The things changed in the testutils:
-The tokenprinter is now called for the tests (should we maybe only do this if the test fails?)
-The configured tests are now passed as a list instead of a set internally, so the order of tests does not change between runs.


mbrdl:
With this change, how could I add semantic information to the tokens as is done in the [TokenGeneratingTreeScanner](https://github.com/jplag/JPlag/blob/develop/languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java)? How could I track variables, for instance?

tsaglam:
@TwoOfTwelve could you extend the PR description a bit? List the fixes to the cpp2 language explicitly (e.g., which tokens/syntax elements were affected).

Also, for testing the token extraction quickly, you can always use the token printer to print the token before and after your change with a large cpp file and then to a text diff.

Finally, the expansion for semantic token information would be of highest priority right now.

tsaglam:
@TwoOfTwelve could you extend the PR description a bit? List the fixes to the cpp2 language explicitly (e.g., which tokens/syntax elements were affected).

Also, for testing the token extraction quickly, you can always use the token printer to print the token before and after your change with a large cpp file and then to a text diff.

Finally, the expansion for semantic token information would be of the highest priority right now.

tsaglam:
@TwoOfTwelve could you extend the PR description a bit? List the fixes to the cpp2 language explicitly (e.g., which tokens/syntax elements were affected).

Also, for testing the token extraction quickly, you can always use the token printer to print the token before and after your change with a large cpp file and then to a text diff.

Finally, the expansion for semantic token information would be of the highest priority right now.

languages/cpp2/src/main/java/de/jplag/cpp2/CPPListener.java
```diff
@@ -0,0 +1,102 @@
...
+import static de.jplag.cpp2.CPPTokenType.*;
+
+import java.io.File;
+
+import de.jplag.antlr.AbstractAntlrListener;
+import de.jplag.antlr.TokenCollector;
+import de.jplag.cpp2.grammar.CPP14Parser;
+import de.jplag.cpp2.grammar.CPP14Parser.*;
+
+public class CPPListener extends AbstractAntlrListener {
```
tsaglam, line 12: Javadoc

languages/cpp2/src/main/java/de/jplag/cpp2/CPPListener.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * New instance
+     * @param collector The token collector
```
tsaglam, line 15: This would be a case of JavaDoc trivial! :D

languages/cpp2/src/test/java/de/jplag/cpp2/CppLanguageTest.java
```diff
@@ -0,0 +1,98 @@
+package de.jplag.cpp2;
+
+import java.util.Arrays;
+
+import de.jplag.testutils.LanguageModuleTest;
+import de.jplag.testutils.datacollector.TestDataCollector;
+import de.jplag.testutils.datacollector.TestSourceIgnoredLinesCollector;
+
+public class CppLanguageTest extends LanguageModuleTest {
```
tsaglam, line 9: Javadoc, what is tested? At which granularity? few sentences can provide outsiders clarity here!

languages/cpp2/src/main/antlr4/de/jplag/cpp2/grammar/CPP14Parser.g4
```diff
@@ -790,6 +790,7 @@ theOperator:
 	| GreaterEqual
 	| PlusAssign
 	| MinusAssign
+	| DivAssign
```
tsaglam, line 793: Maybe add a comment here that this was changed from the original grammar. You could also make a PR to the antlr repo.

tsaglam (changes requested):
We may need to clarify in the language readme that this grammar has been changed compared to the original one. Ideally, add a note on what has been changed.

language-antlr-utils/src/main/java/de/jplag/antlr/AbstractAntlrListener.java
```diff
@@ -188,6 +191,72 @@ protected void createTerminalMapping(int terminalType, TokenType jplagType, Pred
...
+     * @param <T> the type of the element to search for.
+     * @return an ancestor of the specified type, or null if not found.
+     */
+    @SafeVarargs
+    protected final <T extends ParserRuleContext> T getAncestor(ParserRuleContext context, Class<T> ancestor,
+            Class<? extends ParserRuleContext>... stops) {
+        ParserRuleContext currentcontext = context;
+        Set<Class<? extends ParserRuleContext>> forbidden = Set.of(stops);
+        do {
+            ParserRuleContext next = currentcontext.getParent();
```
mbrdl, line 208: `currentContext = currentContext.getParent()` would be simpler

mbrdl (changes requested):
some minor things

languages/cpp2/src/main/java/de/jplag/cpp2/CPPListener.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param currentFile The currently processed file
+     */
+    public CPPListener(TokenCollector collector, File currentFile) {
+        super(collector, currentFile);
+
+        createStartStopMapping(ClassSpecifierContext.class, UNION_BEGIN, UNION_END, rule -> rule.classHead().Union() != null);
```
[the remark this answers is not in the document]
mbrdl: Right, makes sense. Do you have time to make this PR? Otherwise I can do it as well. But I should really get started on the semantic token info 😅

PR #1176, opened 2023-07-05
Internal Report Viewer: host and use the the report viewer locally

dfuchss:
This PR re-integrates a ReportViewer into JPlag. It allows users to launch the web-based report viewer locally by self-hosting it automatically. This feature is currently not mandatory, meaning users can choose to use it via the CLI parameter `--mode`. This parameter supports only running, running + viewing, as well as only viewing. In the future, this local model will become the default.

This PR thus also contributes to #1190.


dfuchss:
JFYI @jplag/maintainer

cli/pom.xml
```diff
@@ -154,6 +166,42 @@
...
+                            <executable>npm</executable>
+                            <workingDirectory>../report-viewer</workingDirectory>
+                            <arguments>
+                                <argument>run</argument>
+                                <argument>build</argument>
+                            </arguments>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
```
dfuchss, line 204: I would not include the build of the report viewer by default. Instead I would add that to a profile. Thus you can implement JPlag without npm :)

cli/pom.xml
```diff
@@ -123,8 +124,20 @@
             <artifactId>picocli</artifactId>
             <version>4.7.5</version>
         </dependency>
+
+        <dependency>
```
dfuchss, line 128: What's jline?

dfuchss (commented):
Just some comments . Probably, Docs are missing

Kr0nox:
@TwoOfTwelve it is merged

Kr0nox:
@TwoOfTwelve The build is currently failing. My guess would be that this is because the action does not have Nodejs installed. Do you see if different issue? If not, do you think it is better to install Nodejs or run the profile that does not need it?

Kr0nox:
Thats interesting. On the merge commit the build failed, but it did not on my commit after that

Kr0nox:
@TwoOfTwelve From my point of view, this is ready for ready. If you dont have anything to add, you can make this a PR

cli/src/main/java/de/jplag/cli/server/ReportViewer.java
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Starts the server
+     * @return The port the server runs at
+     * @throws IOException If the server cannot be started
+     */
+    public int start() throws IOException {
+        if (server != null) {
+            throw new IllegalStateException("Server already started");
+        }
+        server = HttpServer.create(new InetSocketAddress(InetAddress.getLoopbackAddress(), 0), 0);
```
Kr0nox, line 50: If I understood it correctly, at the moment we let the system pick a port. This will make sharing links with other people (was a requested feature) nearly impossible. We should either pick a default port or let the user set a port via a flag

cli/src/main/java/de/jplag/cli/server/ResponseData.java
```diff
@@ -0,0 +1,34 @@
...
+        if (url.endsWith("/")) {
+            return null;
+        }
+
+        InputStream inputStream = ResponseData.class.getResourceAsStream(url);
+
+        if (inputStream != null) {
+            return new ResponseData(inputStream, ContentType.fromPath(url));
+        } else {
+            return null;
```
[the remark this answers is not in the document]
dfuchss: is `inputStream == null` normal behavior here? Then I'd add a small comment here.

cli/src/main/java/de/jplag/cli/server/ReportViewer.java
dfuchss: I'd agree here. You could simply make it configurable. Add a default port; e.g., 1996 (Year of JPlag :D)

tsaglam: +1 for 1996 :D

Kr0nox: What I meant here, was sharing as in sending somebody the zip and the link to the comparison. This was shortly discussed in the workshop last summer.
For this a default/setable port is needed, so that the link works without changing the port

dfuchss:
@jplag/hiwi it seems that we have a merge conflict here :)

dfuchss:
@jplag/hiwi I'd add a cmd line parameter (with default 1996) .

Kr0nox:
The README und wiki CLI help text need to be updated with the mode flag

dfuchss:
> We should automatically use the next available port if the default port is blocked.

I'd say if the port is not available, the CLI should provide an error and exit.

tsaglam:
> I'd say if the port is not available, the CLI should provide an error and exit.

From a usability standpoint, we prefer not throwing an error, as leads to less frustration for users.

dfuchss:
I don't like programs that are randomly blocking my ports tbh :D

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -114,6 +117,25 @@ public CLI() {
...
+        return reportObjectFactory.createAndSaveReport(result, getResultFolder());
+    }
+
+    public void runViewer(File zipFile) throws IOException {
+        ReportViewer reportViewer = new ReportViewer(zipFile, this.options.advanced.port);
+        int port = reportViewer.start();
+        logger.info("ReportViewer started on port http://localhost:{}", port);
+        Desktop.getDesktop().browse(URI.create("http://localhost:" + port + "/"));
+
+        System.out.println("Press Enter kex to exit...");
```
tsaglam, line 134: typo: "key"
Also, why `System.out`?

cli/src/main/java/de/jplag/cli/CliOptions.java
```diff
@@ -53,6 +53,9 @@ public class CliOptions implements Runnable {
             "--result-directory"}, description = "Name of the directory in which the comparison results will be stored (default: ${DEFAULT-VALUE})%n")
     public String resultFolder = "results";
 
+    @Option(names = {"--mode"}, description = "The mode to run jplag in")
+    public JPlagMode mode = JPlagMode.RUN_AND_VIEW;
```
tsaglam, line 57: This means it is currently the default? Should be run then.

cli/src/main/java/de/jplag/cli/JPlagMode.java
```diff
@@ -0,0 +1,19 @@
+package de.jplag.cli;
+
+/**
+ * The mode Jplag runs in. This influences which steps JPlag will execute
```
tsaglam, line 4: Typo: JPlag

cli/src/main/java/de/jplag/cli/JPlagMode.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public enum JPlagMode {
+    /**
+     * Only run jplag and create a results.zip
```
tsaglam, line 8: Type: JPlag (we have three different spellings of JPlag in this file :D)

cli/src/main/java/de/jplag/cli/server/ContentType.java
```diff
@@ -0,0 +1,39 @@
+package de.jplag.cli.server;
+
+/**
+ * The type of data
```
tsaglam, line 4: This comment is a bit trivial...

cli/src/main/java/de/jplag/cli/server/ContentType.java
```diff
@@ -0,0 +1,39 @@
...
+        return value;
+    }
+
+    /**
+     * Guesses the type from the given path using the suffix after the last '.'.
+     * @param path The path to guess from
+     * @return The guessed type
+     */
+    public static ContentType fromPath(String path) {
+        return switch (path.substring(path.lastIndexOf('.'))) {
```
tsaglam, line 30: No switch is required if you make the suffix a field of the numerals.
Then you can just iterate over all and check which one fits. Thus easier to maintain and to extend.

cli/src/main/java/de/jplag/cli/server/HttpMethod.java
```diff
@@ -0,0 +1,29 @@
+package de.jplag.cli.server;
+
+/**
+ * Available http methods
```
tsaglam, line 4: Also here trivial.

cli/src/main/java/de/jplag/cli/server/HttpMethod.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public enum HttpMethod {
+    GET("GET"),
+    POST("POST");
+
+    private final String name;
+
+    HttpMethod(String name) {
```
tsaglam, line 12: JDoc

cli/src/main/java/de/jplag/cli/server/ReportViewer.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * @param zipFile The zip file to use for the report viewer
+     * @param port The port to use for the server. You can use 0 to use any free port.
+     * @throws IOException If the zip file cannot be read
+     */
+    public ReportViewer(File zipFile, int port) throws IOException {
+        this.routingTree = new RoutingTree();
+
+        this.routingTree.insertRouting("", new RoutingResources("report-viewer").or(new RoutingAlias("index.html")));
+        this.routingTree.insertRouting("results.zip", new RoutingStaticFile(zipFile, ContentType.ZIP));
```
tsaglam, line 40: What if I use -r foo? Does this still work?

tsaglam: Also, all these strings should probably constants.

cli/src/main/java/de/jplag/cli/server/ReportViewer.java
```diff
@@ -0,0 +1,110 @@
...
+        this.port = port;
+    }
+
+    /**
+     * Starts the server
```
tsaglam, line 45: Which server? Is a bit trivial...

cli/src/main/java/de/jplag/cli/server/RoutingResources.java
```diff
@@ -0,0 +1,31 @@
...
+public class RoutingResources implements Routing {
+    private String prefix;
+
+    /**
+     * @param prefix The prefix to use within the resources
+     */
+    public RoutingResources(String prefix) {
+        this.prefix = prefix;
+
+        if (!this.prefix.startsWith("/")) {
```
tsaglam, line 17: Let us deduplicate all these magic strings via a constant! :D

cli/src/test/java/de/jplag/cli/ReportViewerTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+
+import de.jplag.cli.server.ReportViewer;
+
+@Timeout(value = 5, unit = TimeUnit.MINUTES)
+class ReportViewerTest {
+    @Test
+    @Disabled
```
tsaglam, line 20: Disabled tests should have a comment `@Disabled("...")`

tsaglam (changes requested):
Preliminary review with some initial pointers on what we still have to work on.

cli/src/main/java/de/jplag/cli/server/RoutingResources.java
tsaglam: Also, this is system-dependent.

dfuchss: Shouldn't be for urls :) or do I miss anything?

tsaglam: Ah, right, did not think about what it was being used for here. My bad!

.github/workflows/maven.yml
```diff
@@ -43,17 +43,21 @@ jobs:
       with:
         java-version: 21
         distribution: 'temurin'
-            
+
+    - uses: actions/setup-node@v4
+      with:
+        node-version: "18"
```
sebinside, line 49: Should we move to the current LTS (v20) here?

README.md
```diff
@@ -59,7 +59,8 @@ JPlag is released on [Maven Central](https://search.maven.org/search?q=de.jplag)
 1. Download or clone the code from this repository.
 2. Run `mvn clean package` from the root of the repository to compile and build all submodules.
    Run `mvn clean package assembly:single` instead if you need the full jar which includes all dependencies.
-5. You will find the generated JARs in the subdirectory `cli/target`.
+   Run `mvn -Pwith-report-viewer clean package assembly:single` to build the full jar with the report viewer. In this case, you'll need `npm` installed.
```
sebinside, line 62: Should we link to nodejs and npm instead of just tagging `npm`?

Kr0nox: I added the link and also changed the link content text to be node.js since that is the correct term for what users need to install

.github/workflows/maven.yml
Kr0nox: We decided to go with the older version, to support devs that still use this version

README.md
```diff
@@ -60,7 +60,8 @@ JPlag is released on [Maven Central](https://search.maven.org/search?q=de.jplag)
 1. Download or clone the code from this repository.
 2. Run `mvn clean package` from the root of the repository to compile and build all submodules.
    Run `mvn clean package assembly:single` instead if you need the full jar which includes all dependencies.
-5. You will find the generated JARs in the subdirectory `cli/target`.
+   Run `mvn -Pwith-report-viewer clean package assembly:single` to build the full jar with the report viewer. In this case, you'll need [Node.js](https://nodejs.org/en/download) installed.
```
Kr0nox, line 63: space between P and w

dfuchss: I think you could do it. But you don't need to do it.

cli/src/main/java/de/jplag/cli/JPlagMode.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public enum JPlagMode {
+    /**
+     * Only run JPlag and create a results.zip
+     */
+    RUN,
+    /**
+     * Only start the report viewer
+     */
+    VIEWER,
```
sebinside, line 14: Minor inconsistency: Should we call this `VIEW` to be consistent with `RUN` and `RUN_AND_VIEW`?

cli/src/main/java/de/jplag/cli/server/ResponseData.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        InputStream inputStream = ResponseData.class.getResourceAsStream(url);
+
+        if (inputStream != null) {
+            return new ResponseData(inputStream, ContentType.fromPath(url));
+        } else {
+            logger.info(
+                    "No response data available for resource url {}. This is probably normal behaviour and the data will be pulled from a different source.",
```
sebinside, line 51: I'm not happy with this logging message as it too visible in a normal run and does not really help the end user (also: `this is probably normal behavior` lacks precision. If it ain't normal behavior, log it, otherwise, don't). Also, the important message `Press Enter key to exit...` gets easy to overlook when this is printed multiple times.

sebinside (approved):
Besides my two small comments (please see above), I'm fine with this PR, works great! We should merge this as soon as possible to have more than 0 days of integration testing before the 5.0 release.

cli/src/main/java/de/jplag/cli/server/ReportViewer.java
```diff
@@ -0,0 +1,131 @@
...
+
+        int port = this.port;
+        int remainingLookups = MAX_PORT_LOOKUPS;
+        BindException lastException = new BindException("Could not bind server.");
+        while (server == null && remainingLookups-- > 0) {
+            try {
+                server = HttpServer.create(new InetSocketAddress(InetAddress.getLoopbackAddress(), port), 0);
+            } catch (BindException e) {
+                lastException = e;
+                port++;
```
Kr0nox, line 69: Add a message that the port was taken and it tries to find a new one

cli/src/main/java/de/jplag/cli/server/ReportViewer.java
```diff
@@ -0,0 +1,131 @@
...
+     * @throws IOException If the server cannot be started
+     */
+    public int start() throws IOException {
+        if (server != null) {
+            throw new IllegalStateException("Server already started");
+        }
+
+        int port = this.port;
+        int remainingLookups = MAX_PORT_LOOKUPS;
+        BindException lastException = new BindException("Could not bind server.");
```
Kr0nox, line 63: Maybe add that it could not find any open port

cli/src/main/resources/index.html
```diff
@@ -0,0 +1,9 @@
+<html lang="en">
```
Kr0nox, line 1: Do we want/need this file. If so should it have a different content?

Kr0nox:
When the initial and the next port are not available it prints the initial port twice
grafik
In the second message it should say `1997 is not available`

Kr0nox (approved):

cli/src/main/java/de/jplag/cli/CliOptions.java
```diff
@@ -53,6 +53,9 @@ public class CliOptions implements Runnable {
             "--result-directory"}, description = "Name of the directory in which the comparison results will be stored (default: ${DEFAULT-VALUE})%n")
     public String resultFolder = "results";
 
+    @Option(names = {"--mode"}, description = "The mode to run jplag in")
```
tsaglam, line 56: Two issues here:

- the default value is missing
- this is very much a central new feature, but the description is lacking.

```suggestion
    @Option(names = {"--mode"}, description = "The mode of JPlag: either only run analysis, only open the viewer, or do both (default: ${DEFAULT-VALUE})")
```

tsaglam (changes requested):
Some minor issues below. The bigger issue is: When I package the jar and either use VIEW or RUN_AND_VIEW I end up with:

```bash
2024-02-20-12:05:59_748 [INFO] CLI - ReportViewer started on port http://localhost:1996
2024-02-20 12:05:59.899 java[52709:33325826] WARNING: Secure coding is automatically enabled for restorable state! However, not on all supported macOS versions of this application. Opt-in to secure coding explicitly by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState:.
Press Enter key to exit...
```

tsaglam:
> Did you make sure you compiled JPlag with -P with-report-viewer? Otherwise the report-viewer will not be bundled and exactly this error will occur. Also I think sometimes maven does not build the report-viewer for some reason, so you could try to build it manually beforehand.

Thanks, now it works! 👍

tsaglam (approved):
Local mode works when built as a jar. Both view only as well as run and view. What is currently not adapted is the readme, but I will do this separately.

PR #1177, opened 2023-07-06
Expanded the language-antlr-utils to allow the extraction of semantic…

TwoOfTwelve:
Adds the option to extract semantics information to the AbstractAntlrListener. For an example how to use, take a look at the tests in language-antlr-utils.


mbrdl (commented):
I'm not sure how this framework pans out in practice, but I think I'll just make adjustements myself if needed.

mbrdl:
I can't really make a call on the framework until I've tried using it. I think you can merge @tsaglam

language-antlr-utils/src/main/java/de/jplag/antlr/ContextTokenBuilder.java
```diff
@@ -0,0 +1,59 @@
...
+        } else {
+            return antlrContent.getStop().getStopIndex() - antlrContent.getStart().getStartIndex() + 1;
+        }
+    }
+
+    enum Type {
+        START,
+        STOP,
+        RANGE
+    }
```
tsaglam, line 58: Not the biggest fan of nested types, especially as this one is not private. Nested types tend to be improperly documented and harder to maintain.

language-antlr-utils/src/main/java/de/jplag/antlr/InternalListenerException.java
```diff
@@ -0,0 +1,15 @@
+package de.jplag.antlr;
+
+/**
+ * Exception type used internally within the antlr utils. Has to be a {@link RuntimeException}, because it is thrown
+ * within the antlr listener methods. Should not be thrown outside the antlr utils.
+ */
+public class InternalListenerException extends RuntimeException {
```
tsaglam, line 7: The name of this exception does not really help with understand when this exception is thrown or what scenario it covers.

language-antlr-utils/src/main/java/de/jplag/antlr/SemanticsDataHolder.java
```diff
@@ -0,0 +1,10 @@
+package de.jplag.antlr;
+
+import de.jplag.semantics.VariableRegistry;
+
+/**
+ * Holds all required context objects for the generation of code semantics
+ * @param registry The variable registry
+ */
+public record SemanticsDataHolder(VariableRegistry registry) {
```
tsaglam, line 9: If this is a record encapsulating a single object, why do we need the record at all?

language-antlr-utils/src/main/java/de/jplag/antlr/TokenBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+import de.jplag.Token;
+import de.jplag.TokenType;
+import de.jplag.semantics.CodeSemantics;
+
+/**
+ * Handles the extraction of tokens. Contains information on the appropriate antlr types, the conditions under which the
+ * token should be extracted and semantics information.
+ * @param <T> The antlr type being mapped
+ */
+@SuppressWarnings("UnusedReturnValue")
```
tsaglam, line 18: Why do we suppress this? We should avoid doing so!

language-antlr-utils/src/main/java/de/jplag/antlr/TokenBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+public abstract class TokenBuilder<T> {
+    private static final Logger LOG = Logger.getLogger(TokenBuilder.class.getName());
```
tsaglam, line 20: Not in line with the usual naming in JPlag.

tsaglam (changes requested):
Some issues we need to address.

language-antlr-utils/src/main/java/de/jplag/antlr/TokenBuilder.java
tsaglam: Yes, I think that is better.

language-antlr-utils/src/main/java/de/jplag/antlr/TokenBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+
+        int line = antlrToken.getLine();
+        int column = antlrToken.getCharPositionInLine() + 1;
+        int length = antlrToken.getText().length();
+
+        Token token;
+        if (semantics != null) {
+            if (semanticsSupplier == null) {
+                throw new InternalListenerException(MISSING_SEMANTICS);
+            }
```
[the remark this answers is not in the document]
tsaglam: I would say, in that case you can just throw an IllegelStateException or a similar runtime exception and just not catch it. Parser exceptions are for the user, but the scenario you are describing here would affect the developer.

language-antlr-utils/src/main/java/de/jplag/antlr/SemanticsDataHolder.java
tsaglam: Makes sense, but could be considered premature optimization of sorts. I am unsure if we should keep it in or not. @mbrdl do you have a preference?

mbrdl: The variable registry even started off as a semantics data holder, the name only came later.

tsaglam: @TwoOfTwelve then let us remove it for now! 👍

tsaglam (approved):

PR #1180, opened 2023-07-11
Simplified names for map functions in AbstractAntlrListener.java

TwoOfTwelve:
Simplified the names of the mapping functions in AbstractAntlrListener


mbrdl:
What's the difference between `mapRange` and `mapStart`? Is it simply the token's length? If so, I would suggest removing `mapRange` for clarity, as token length isn't used anywhere and should probably be removed altogether

mbrdl:
And what about `mapTerminal`? I don't really get it

Kr0nox:
Looks good. Only thing missing is what we discussed in #1159

tsaglam:
@TwoOfTwelve should we merge this PR now and address the changes discussed above in another PR? Or do we want to include it here?

mbrdl:
@tsaglam please merge this, I'm starting work on the C++ semantic tokens based on it

tsaglam (approved):
@TwoOfTwelve Lets move the other issues discussed here into another PR!

Issue #1134, opened 2023-06-13
Add more similarity metrics


tsaglam:
Currently, we only support two similarity metrics:

_let there be two files $F_a$ and $F_b$ with token lengths $a$ and $b$ as well as matched tokens $m$_

- average similarity: $\frac{\frac{m}{a} + \frac{m}{b}}{2}$ 
- maximum similarity: $max(\frac{m}{a}, \frac{m}{b})$

We could add two different metrics:

- Symmetric similarity:  $\frac{2m}{a + b}$ (better name required)
- Overlap: $m$

To implement this, both the core and the report viewer need to support it. With this issue, we could also look at the similarity metric again, where it is used, and where it is _not_ used.

uuqjz:
As discussed:
A metric regarding the number of tokens in submissions, marking outliers, would be interesting for the report viewer.

SimDing:
> - Symmetric similarity: $\frac{2m}{a+b}$ (better name required)

This is often equivalent to the Jaccard Index, aka. the Jaccard similarity coefficient, or intersection over union.

Also, it would be interesting to see more similarities that take into account the length of the matches (apart from "Longest Match") and maybe even the full sample of matches in the whole set of submissions.

One idea: "Longest Match" and "Overlap" both belong to a family of related similarities:
Let $t$ be the vector of token sizes of the matches where  $||t||_1=m$. Every vector norm applied on $t$ induces a similarity.
The max norm induces "Longest Match" and the sum norm induces "Overlap". p-norms with p > 1, might be particularly interesting because they'd assign higher similarities to submissions with longer matches, compared to submissions with many short matches. This might allow choosing smaller minimal match sizes.

robinmaisch:
> Also, it would be interesting to see more similarities that take into account the length of the matches (apart from "Longest Match") and maybe even the full sample of matches in the whole set of submissions.
> 
> One idea: "Longest Match" and "Overlap" both belong to a family of related similarities: Let t be the vector of token sizes of the matches where | | t | | 1 = m . Every vector norm applied on t induces a similarity. The max norm induces "Longest Match" and the sum norm induces "Overlap". p-norms with p > 1, might be particularly interesting because they'd assign higher similarities to submissions with longer matches, compared to submissions with many short matches. This might allow choosing smaller minimal match sizes.

That's a nice idea! Rather large matches should be brought to the users' attention, even in relatively much larger submissions. 

[...]

PR #1189, opened 2023-07-13
Slight changes to the report json format

Kr0nox:
This PR applies some slight changes to the format of the overview.json file amd comparison files to make it easier to extend with new Metrics (#1134) and make it fit the format used inside the report viewer.

# Changes to the overview.json
The `metric` attribute is removed.
A new attribute `distributions` is added. This is a map of each metric as a key (3 Letter abbreviation in upper case e.g. "AVG") and as a key it as a List of numbers as the distribution. This List is identical to the one previously saved in each metric.
A new attribute `top_comparisons` is added. This is a list of objects. Each object saves the ids of the two submissions. It also saves a map with each metric as a key and the similarity for this metric as a value.

Here is an exampl of how the new overview.json looks compared to the old one. All unchanged attributes values are replaced by `...`:
Old: 
```
{
    "jplag_version": ...,
    "submission_folder_path": ...,
    "base_code_folder_path": ...,
    "language": ...,
    "file_extensions": ...,
    "submission_id_to_display_name": ...,
    "submission_ids_to_comparison_file_name": ...,
    "failed_submission_names": ...,
    "excluded_files": ..,
    "match_sensitivity": ...,
    "date_of_execution": ...,
    "execution_time": ...,
    "metrics": [
        {
            "name": "AVG",
            "distribution": [1, 0, 2, 0, 0, 0, 0, 3, 0, 0],
            "topComparisons": [
                {
                    "first_submission": "A",
                    "second_submission": "C",
                    "similarity": 0.9959390862944163
                },
                {
                    "first_submission": "B",
                    "second_submission": "A",
                    "similarity": 0.23864648263579696
                },
                {
                    "first_submission": "B",
                    "second_submission": "C",
                    "similarity": 0.23864648263579696
                }
            ],
            "description": "Average of both program coverages. This is the default similarity which works in most cases: Matches with a high average similarity indicate that the programs work in a very similar way."
        },
        {
            "name": "MAX",
            "distribution": [5, 1, 0, 0, 0, 0, 0, 0, 0, 0
            ],
            "topComparisons": [
                {
                    "first_submission": "A",
                    "second_submission": "C",
                    "similarity": 0.9959390862944163
                },
                {
                    "first_submission": "B",
                    "second_submission": "A",
                    "similarity": 0.9710144927536232
                },
                {
                    "first_submission": "B",
                    "second_submission": "C",
                    "similarity": 0.9710144927536232
                }
            ],
            "description": "Maximum of both program coverages. This ranking is especially useful if the programs are very different in size. This can happen when dead code was inserted to disguise the origin of the plagiarized program."
        }
    ],
    "clusters": ...,
    "total_comparisons": ...
}
```
New:
```
{
    "jplag_version": ...,
    "submission_folder_path": ...,
    "base_code_folder_path": ...,
    "language": ...,
    "file_extensions": ...,
    "submission_id_to_display_name": ...,
    "submission_ids_to_comparison_file_name": ...,
    "failed_submission_names": ...,
    "excluded_files": ..,
    "match_sensitivity": ...,
    "date_of_execution": ...,
    "execution_time": ...,
    "distributions": {
        "AVG": [1, 0, 2, 0, 0, 0, 0, 3, 0, 0],
        "MAX": [5, 1, 0, 0, 0, 0, 0, 0, 0, 0]
    },
    "top_comparisons": [
        {
          "first_submission": "A",
          "second_submission": "C",
          "similarities": { "AVG": 0.9960435212660732, "MAX": 0.9960435212660732 }
        },
        {
          "first_submission": "B",
          "second_submission": "A",
          "similarities": { "AVG": 0.2378472222222222, "MAX": 0.9716312056737588 }
        },
        {
          "first_submission": "B",
          "second_submission": "C",
          "similarities": { "AVG": 0.2378472222222222, "MAX": 0.9716312056737588 }
        }
      ],
    "clusters": ...,
    "total_comparisons": ...
}
```

# Changes to the comparison files
`similarity` gets replaced by `similarities` and is now a map of similarities.

Old:
```
{
  "id1": ...
  "id2": ...
  "similarity": 0.3291991495393338,
  "matches":  ...
}
```
New:
```
{
  "id1": ...
  "id2": ...
  "similarities": { "AVG": 0.9960435212660732, "MAX": 0.9960435212660732 },
  "matches":  ...
}
```


Kr0nox:
One thing I would also like to put to discussion is modifying the distributions to contain a value for each percentile and not for cummulative 10 percentiles. This would enable us the display more detailed distribution graphs

tsaglam:
> One thing I would also like to put to discussion is modifying the distributions to contain a value for each percentile and not for cummulative 10 percentiles. This would enable us the display more detailed distribution graphs

I would say we could play around with some more fine-grained options for the distribution (1 or 5, for example). In the UI we should always offer an option to toggle back to the old 10-based distribution.

tsaglam:
@Alberth289346 just FYI, this PR will be part of a major release sometimes in the next quarter, where the [overhauled report viewer](https://jplag.github.io/JPlag-Dev/) will be released. This might break things for you.

sebinside:
How much effort would it be to encapsulate the deserialization in the report viewer and also to support legacy reports, i.e., reports generated with the current version prior to this PR? If it is reasonable I would vote for supporting this as the current JSON format was at least somewhat stable throughout the past.

Kr0nox:
> How much effort would it be to encapsulate the deserialization in the report viewer and also to support legacy reports

Not much work. Would be as easy as testing if the json has an attribute called metric and then extracting the values the old way and if it does not exists we would use the new way.
So if that is wanted I an implement that

sebinside:
Also looking at the fact that the current version of the Report Viewer will be the legacy one deployed online (maybe for a long time), I would opt for that. However, only if this does not significantly degrade the code quality and structure of the viewer. Any thoughts on this @tsaglam?

tsaglam:
> Also looking at the fact that the current version of the Report Viewer will be the legacy one deployed online (maybe for a long time), I would opt for that.

Good point, I agree.

For the distribution we could proceed similarly, if it only has ten values, you only see the old histogram.
If you have more values, you see the new one and can toggle between old and new.

Alberth289346:
> @Alberth289346 just FYI, this PR will be part of a major release sometimes in the next quarter, where the [overhauled report viewer](https://jplag.github.io/JPlag-Dev/) will be released. This might break things for you.

Thanks for the head-up.

I had similar thoughts about this issue as sebinside, why is data *moved* ? It breaks compatibility without getting much in return (data isn't becoming more useful by moving it as far as I can see). I pondered about it, and wrote #1190 to move the discussion away from here.

report-viewer/src/components/DistributionDiagram.vue
```diff
@@ -51,7 +51,7 @@ const chartData = ref({
   datasets: [
     {
       ...dataSetStyle.value,
-      data: props.distribution.distribution
+      data: props.distribution.getTenthPercentileFormattedValues()
```
Alberth289346, line 54: You could use `perMille` here (if that's what you mean to express) https://en.wikipedia.org/wiki/Per_mille

Kr0nox: No thats not what I mean, but now that you mention it I can see where the confusion comes from and we should probably search for a better name.
The function does is return the distribution in the "old" way, so summed up to every thenth percentile (so 0%-10%, 10%-20%, ...). 
Do you think it would be clear enough if the funtion was named `getTenthPercentileSumDistribution()`?

Alberth289346: I read "tenth" as `1/10`, `TenPercentile-something` would be better for me.

However, the name seems complicated to grasp and very specific. I might write a somewhat more generic `splitInEqualBuckets(N)` function, and then stick `10` in as an argument.

If you want a dedicated function for it, something like `splitIntoTenBuckets` or so?

core/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -190,7 +189,8 @@ private void writeOverview(JPlagResult result, String path) {
                 result.getOptions().minimumTokenMatch(), // matchSensitivity
                 getDate(),// dateOfExecution
                 result.getDuration(), // executionTime
-                getMetrics(result),// metrics
+                MetricMapper.getDistributions(result), // distribution
+                new MetricMapper(submissionToIdFunction).getTopComparisons(result),// topComparisons
```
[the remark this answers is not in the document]
tsaglam: I think they enforce the linebreak for spotless, but `//` would be sufficient, yes.

tsaglam (commented):
Regarding the Java side, this looks good.

sebinside (approved):

PR #1202, opened 2023-07-20
Match merging against obfuscation

uuqjz:
This PR proposes Match Merging, a new defence mechanism against obfuscations like insertions, alterations and swapping.

How does it work:

The greedy string tiling algorithm at the core of JPlag calculates matches between submissions. 
In order to not raise similarity with very short matches the MinimumTokenMatch (MTM) defines the minimal length of a token-match in order to be considered for the similarity calculation. In Java it is 9 and in CPP 12. 
Tools like Mossad or JPlag-GEN use this for plagiarism obfuscations by applying changes to submissions which keep the semantics identically but break up large matches into multiple smaller ones. 
When done often enough most matches are below MTM and therefore ignored, resulting in a very low similarity score between the original and obfuscated plagiarism.
My approach aims to revert these changes on submissions by merging neighbouring matches and is based on two parameters mergeBuffer and seperatingThreshold.
mergeBuffer defines how lower the length of a match can be than the MTM.
seperatingThreshold defines how many tokens can be between two neighbouring matches.
My approach considers previously ignored matches and regular ones and merges them based on the parameters.
Merged-over tokens are removed from the submission.
Currently, both parameters default to 0 which disables my logic.
The default values will be changed in a later PR.

Which classes are affected and why:

- MatchMerging contains the core logic for my approach
- JPlag now calls MatchMerging in its pipeline
- MergingParameters stores mergeBuffer and seperatingThreshold and follows ClusteringOptions
- JPlagOptions, CLI and CliOptions now include MergingParameters also following ClusteringOptions
- JPlagComparison has a new field ignoredMatches, which stores previously ignored Matches
- GreedyStringTiling now lowers its MinimumMatchLength according to mergeBuffer and stores the previously ignored matches in ignoredMatches
- Submission has a new copy function which is necessary as my approach removes tokens that are merged-over from the submission
- MergingTest contains 7 test cases for my approach based on two samples


uuqjz:
My last commit fixed an oversight while logging similarities.
Initially, the similarities were logged in the ComparisonStrategy, which made them inaccurate when applying match merging to the results.
With these changes, the correct similarities are logged from JPlag instead.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,134 @@
...
+ * This class implements match merging against obfuscation. Based on configurable parameters MergeBuffer and
+ * SeperatingThreshold, it alters prior results and merges all neighboring matches that fit the specified thresholds.
+ * When neighboring matches get merged they become one and the tokens separating them get removed from the submission
+ * clone. MergeBuffer describes how shorter a match can be than the Minimum Token Match. SeperatingThreshold describes
+ * how many tokens can be between two neighboring matches. Both are set in {@link JPlagOptions} as
+ * {@link MergingParameters} and default to 0 (which deactivates merging).
+ */
+public class MatchMerging {
+    private int minimumTokenMatch;
+    private Submission leftSubmission;
```
Kr0nox, line 25: I would suggest going with JPlags naming of `firstSubmission` and `secondSubmission`

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,134 @@
...
+                computeNeighbors();
+                i = 0;
+            } else {
+                i++;
+            }
+        }
+    }
+
+    private void removeToken(int startLeft, int startRight, int lengthUpper, int seperatingLeft, int seperatingRight) {
+        List<Token> tokenLeft = new ArrayList<>(leftSubmission.getTokenList());
```
Kr0nox, line 104: Here the same with `left`, `right` and `first`, `second`

core/src/main/java/de/jplag/strategy/AbstractComparisonStrategy.java
```diff
@@ -46,7 +41,6 @@ protected void compareSubmissionsToBaseCode(SubmissionSet submissionSet) {
      */
     protected Optional<JPlagComparison> compareSubmissions(Submission first, Submission second) {
         JPlagComparison comparison = greedyStringTiling.compare(first, second);
-        logger.info("Comparing {}-{}: {}", first.getName(), second.getName(), comparison.similarity());
```
Kr0nox, line 49: Could removing this result in having a long time without a log in large submission sets? If so I think we should still keep it and maybe just remove the similarity or add that it is only a temporary value.
Otherwise outside users might have the feeling that JPlag is not working.

Kr0nox (changes requested):
A few small things I found

tsaglam:
> With these changes, the correct similarities are logged from JPlag instead.

@uuqjz  The problem with your current implementation is that JPlag is logging nothing for quite some time for large datasets and then dumps all of the similarities in the log. While this logging of similarities has been a topic for discussion recently anyways, I would say for now let us keep the old logging even though the similarities might be then off.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,134 @@
...
+            rightSubmission = comparisons.get(i).secondSubmission().copy();
+            globalMatches = new ArrayList<>(comparisons.get(i).matches());
+            globalMatches.addAll(comparisons.get(i).ignoredMatches());
+            computeNeighbors();
+            mergeNeighbors();
+            removeBuffer();
+            comparisons.set(i, new JPlagComparison(leftSubmission, rightSubmission, globalMatches, new ArrayList<>()));
+
+        }
+        return new JPlagResult(comparisons, result.getSubmissions(), result.getDuration(), options);
```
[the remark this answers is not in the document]
uuqjz: Could you explain what you mean or want to be changed?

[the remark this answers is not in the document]
Kr0nox: Currently you instantiate the new result, with the duration from the old result. To my understanding (correct me if I'm wrong) this means this is just the time from the GreedyStringTiling.
If the MatchMerging takes a long time especially for larger submissions sets, I think it would be usefull to add this time to the total duration

uuqjz:
@Kr0nox should I click on "Resolve conversation" for stuff I fixed, or does the requester usually do this?

core/src/main/java/de/jplag/merging/MatchMerging.java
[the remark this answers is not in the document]
uuqjz: I've never considered the duration. Should not be too hard to increase it in match merging.

Kr0nox:
@uuqjz You can close them. I personally find it easier then to see what excactly you did and you also know better what issues you solved

cli/src/main/java/de/jplag/cli/CliOptions.java
```diff
@@ -109,6 +112,15 @@ public static class ClusteringEnabled {
         }
     }
 
+    public static class Merging {
+        @Option(names = {"--merge-buffer"}, description = "Buffer for merging (default: 0)\n")
```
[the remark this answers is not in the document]
uuqjz: `Match Merging
Defense mechanism against obfuscation that merges neighboring matches based on these parameters:
      --merge-buffer=<mergeBuffer>
                       Defines how much lower the length of a match can be than the minimum match length (default: 0)

      --seperating-threshold=<seperatingThreshold>
                       Defines how many token can be between two neighboring matches (default: 0)`
Do you think this is verbose enough?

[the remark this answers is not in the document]
Kr0nox: yes that sounds good 👍

uuqjz:
@tsaglam @Kr0nox regarding logging the similarities:
I agree with your point that the user should receive constant feedback that jplag is working.

But I think that logged similarities should be accurate, in case someone is using the console stdout instead of the report.zip.
If simply reverting my commit, these people would be under the impression that they are still getting the right result when they're not.

I would propose some kind of heartbeat logging in the AbstractComparisonStrategy that does not include actual similarities.
For example:
`logger.info("Working on {}-{}", first.getName(), second.getName());`

In this way, users still get constant feedback while the actual similarities logged are accurate.

What do you think?

tsaglam:
> But I think that logged similarities should be accurate, in case someone is using the console stdout instead of the report.zip. If simply reverting my commit, these people would be under the impression that they are still getting the right result when they're not.

Fair point, however, in the long run we might replace this kind of logging entirely.

> I would propose some kind of heartbeat logging in the AbstractComparisonStrategy that does not include actual similarities. For example: `logger.info("Working on {}-{}", first.getName(), second.getName());`
> In this way, users still get constant feedback while the actual similarities logged are accurate.

This would make the heartbeat logging less useful than it currently already is.
I would say we return to the original logging for now, as this is a problem that should not be solved in this PR.

However, for the match merging we should think about a good and clean way to log information.
E.g. one line if the merging is started, making it clear that it was something the user configured via the CLI.
And then one line after it is done, maybe even with some stats (e.g. diff in avg and med).

uuqjz:
FYI

`if (leftTokens.size() <= minimumMatchLength || rightTokens.size() <= minimumMatchLength) {
            return new JPlagComparison(leftSubmission, rightSubmission, List.of(), List.of());
        }`

in GreedyStringTiling cannot be covered as 

`if (submission.getTokenList() != null && submission.getNumberOfTokens() < options.minimumTokenMatch()) {
                logger.error("Submission {} contains {} token(s), which is less than the minimum match length ({})!", currentSubmissionName,
                        submission.getNumberOfTokens(), options.minimumTokenMatch());
                submission.setTokenList(null);
                tooShort++;
                ok = false;
                submission.markAsErroneous();
            }`

from SubmissionSet does not let submissions pass if they don't have enough tokens.
Therefore these submissions never reach GreadyStringTiling

tsaglam:
@uuqjz can you make an issue for this? I will take a look at it when I find time for it.

Kr0nox (changes requested):

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -220,6 +222,10 @@ private static ClusteringOptions getClusteringOptions(CliOptions options) {
         return clusteringOptions;
     }
 
+    private static MergingParameters getMergingParameters(CliOptions options) {
+        return new MergingParameters().withMergeBuffer(options.merging.mergeBuffer).withSeperatingThreshold(options.merging.seperatingThreshold);
```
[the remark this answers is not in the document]
uuqjz: I followed ClusteringOptions from CLI:

`new ClusteringOptions().withEnabled(!options.clustering.disable)
                .withAlgorithm(options.clustering.enabled.algorithm).withSimilarityMetric(options.clustering.enabled.metric).withSpectralKernelBandwidth(options.clusterSpectralBandwidth).withSpectralGaussianProcessVariance(options.clusterSpectralNoise)
                .withSpectralMinRuns(options.clusterSpectralMinRuns).withSpectralMaxRuns(options.clusterSpectralMaxRuns)
                .withSpectralMaxKMeansIterationPerRun(options.clusterSpectralKMeansIterations)
                .withAgglomerativeThreshold(options.clusterAgglomerativeThreshold)
                .withAgglomerativeInterClusterSimilarity(options.clusterAgglomerativeInterClusterSimilarity);`

I wanted to keep it consistent, but if you want I can give the parameters directly to the constructor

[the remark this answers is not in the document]
uuqjz: Additionally:
If we add more parameters in the future, the cast 
`new MergingParameters(x,y,z,i,j,k)` 
will get quite messy in terms of understanding the parameters

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+     * Constructor for class MatchMerging
+     * @param result is the initially computed result object
+     * @param options encapsulates the adjustable options
+     */
+    public MatchMerging(JPlagResult result, JPlagOptions options) {
+        this.result = result;
+        this.comparisons = new ArrayList<>(result.getAllComparisons());
+        this.options = options;
+        this.minimumTokenMatch = options.minimumTokenMatch();
+        this.seperatingThreshold = options.mergingParameters().seperatingThreshold();
```
tsaglam, line 44: Both of these fields are only used once, but you also keep the options as fields. Remove both fields and access options directly.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+     * Runs the internal match merging pipeline. It computes neighboring matches, merges them based on
+     * {@link MergingParameters} and removes the merge buffer afterwards.
+     * @return JPlagResult containing the merged matches
+     */
+    public JPlagResult run() {
+        long timeBeforeStartInMillis = System.currentTimeMillis();
+        for (int i = 0; i < comparisons.size(); i++) {
+            firstSubmission = comparisons.get(i).firstSubmission().copy();
+            secondSubmission = comparisons.get(i).secondSubmission().copy();
+            globalMatches = new ArrayList<>(comparisons.get(i).matches());
```
tsaglam, line 57: If the global matches are only instantiated once, you should make them final and instantiate them empty in the constructor.

tsaglam: Also, regarding naming, are these just the combined matches (normal + ignored)? Do you make sure to not match across file boundaries?

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * Runs the internal match merging pipeline. It computes neighboring matches, merges them based on
+     * {@link MergingParameters} and removes the merge buffer afterwards.
+     * @return JPlagResult containing the merged matches
+     */
+    public JPlagResult run() {
+        long timeBeforeStartInMillis = System.currentTimeMillis();
+        for (int i = 0; i < comparisons.size(); i++) {
```
tsaglam, line 54: Why do you need the array-style loop here? A foreach loop would probably be cleaner, right?
I would change move the JPlagResult class from a constructor parameter to a parameter of this method. Thus the algorithm is parameterized via the JPlagOptions and then executed for a certain result object.
Then I would use the comparisons directly here and then add the new comparisons in an empty list, thus avoiding unnecessary indexing.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+        this.minimumTokenMatch = options.minimumTokenMatch();
+        this.seperatingThreshold = options.mergingParameters().seperatingThreshold();
+    }
+
+    /**
+     * Runs the internal match merging pipeline. It computes neighboring matches, merges them based on
+     * {@link MergingParameters} and removes the merge buffer afterwards.
+     * @return JPlagResult containing the merged matches
+     */
+    public JPlagResult run() {
```
tsaglam, line 52: We need a more descriptive name here.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+        long durationInMillis = System.currentTimeMillis() - timeBeforeStartInMillis;
+        return new JPlagResult(comparisons, result.getSubmissions(), result.getDuration() + durationInMillis, options);
+    }
+
+    private void computeNeighbors() {
+        neighbors = new ArrayList<>();
+        List<Match> sortedByFirst = new ArrayList<>(globalMatches);
+        Collections.sort(sortedByFirst, (m1, m2) -> m1.startOfFirst() - m2.startOfFirst());
+        List<Match> sortedBySecond = new ArrayList<>(globalMatches);
+        Collections.sort(sortedBySecond, (m1, m2) -> m1.startOfSecond() - m2.startOfSecond());
```
tsaglam, line 74: Regarding naming, I would prefer something like first, second.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+            mergeNeighbors();
+            removeBuffer();
+            comparisons.set(i, new JPlagComparison(firstSubmission, secondSubmission, globalMatches, new ArrayList<>()));
+
+        }
+        long durationInMillis = System.currentTimeMillis() - timeBeforeStartInMillis;
+        return new JPlagResult(comparisons, result.getSubmissions(), result.getDuration() + durationInMillis, options);
+    }
+
+    private void computeNeighbors() {
```
tsaglam, line 69: Even those these methods are private, we could add JDoc comments just for ease of maintenance.
Describe what these substeps of the algorithm do.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+        List<Match> sortedBySecond = new ArrayList<>(globalMatches);
+        Collections.sort(sortedBySecond, (m1, m2) -> m1.startOfSecond() - m2.startOfSecond());
+        for (int i = 0; i < sortedByFirst.size() - 1; i++) {
+            if (sortedBySecond.indexOf(sortedByFirst.get(i)) == (sortedBySecond.indexOf(sortedByFirst.get(i + 1)) - 1)) {
+                neighbors.add(Arrays.asList(sortedByFirst.get(i), sortedByFirst.get(i + 1)));
+            }
+        }
+    }
+
+    private void mergeNeighbors() {
```
tsaglam, line 82: Also here: JDoc.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+                        seperatingSecond);
+                computeNeighbors();
+                i = 0;
+            } else {
+                i++;
+            }
+        }
+    }
+
+    private void removeToken(int startFirst, int startSecond, int lengthUpper, int seperatingFirst, int seperatingSecond) {
```
tsaglam, line 105: Also here JDoc.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+            }
+            if (globalMatches.get(i).startOfSecond() > startSecond) {
+                Match alteredMatch = new Match(globalMatches.get(i).startOfFirst(), globalMatches.get(i).startOfSecond() - seperatingSecond,
+                        globalMatches.get(i).length());
+                globalMatches.set(i, alteredMatch);
+            }
+        }
+    }
+
+    private void removeBuffer() {
```
tsaglam, line 127: Also here.

core/src/test/java/de/jplag/merging/MergingTest.java
```diff
@@ -0,0 +1,137 @@
...
+import de.jplag.Match;
+import de.jplag.SubmissionSet;
+import de.jplag.SubmissionSetBuilder;
+import de.jplag.TestBase;
+import de.jplag.exceptions.ExitException;
+import de.jplag.options.JPlagOptions;
+import de.jplag.strategy.ComparisonStrategy;
+import de.jplag.strategy.ParallelComparisonStrategy;
+
+class MergingTest extends TestBase {
```
tsaglam, line 25: JDoc: What do we test in this class?

core/src/test/java/de/jplag/merging/MergingTest.java
```diff
@@ -0,0 +1,137 @@
...
+        ComparisonStrategy comparisonStrategy = new ParallelComparisonStrategy(options, coreAlgorithm);
+
+        SubmissionSetBuilder builder = new SubmissionSetBuilder(options);
+        SubmissionSet submissionSet = builder.buildSubmissionSet();
+
+        result = comparisonStrategy.compareSubmissions(submissionSet);
+        comparisonsBefore = result.getAllComparisons();
+
+        result = new MatchMerging(result, options).run();
+        comparisonsAfter = result.getAllComparisons();
```
tsaglam, line 45: Test state should not be shared between test cases, use @BeforeEach and a dedicated method.
Also, we could use a constant for the merging parameter record to emphasize them.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+
+    private void mergeNeighbors() {
+        int i = 0;
+        while (i < neighbors.size()) {
+            int lengthUpper = neighbors.get(i).get(0).length();
+            int lengthLower = neighbors.get(i).get(1).length();
+            int seperatingFirst = neighbors.get(i).get(1).startOfFirst() - neighbors.get(i).get(0).endOfFirst() - 1;
+            int seperatingSecond = neighbors.get(i).get(1).startOfSecond() - neighbors.get(i).get(0).endOfSecond() - 1;
```
tsaglam, line 88: Accessing the neighbors gets really hard to read. Can we redesign this so it is less index magic?
The underlying reason for this is probably the `List<List<Match>>`, which is also not straightforward.
Is it for mapping each match to its neighbors? If yes, we should use a `HashMap<Match, List<Match>`, as this avoids index magic in favor of something like this: `neighbors.get(match);`

tsaglam: I have also trouble understanding the magic numbers used here. Why do we only check the first two neighbors? Are there only two because it is the direct ones? Then we should think about appropriate modeling and create a custom datastructure where the semantics are more explicitly modeled.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+            double seperating = (seperatingFirst + seperatingSecond) / 2.0;
```
tsaglam, line 89: What does separating mean here?

tsaglam: Separation? TokensBetweenNeighbors?

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+            // Checking length is not necessary as GST already checked length while computing matches
+            if (seperating <= seperatingThreshold) {
+                globalMatches.removeAll(neighbors.get(i));
+                globalMatches
+                        .add(new Match(neighbors.get(i).get(0).startOfFirst(), neighbors.get(i).get(0).startOfSecond(), lengthUpper + lengthLower));
+                removeToken(neighbors.get(i).get(0).startOfFirst(), neighbors.get(i).get(0).startOfSecond(), lengthUpper, seperatingFirst,
```
tsaglam, line 95: As discussed previously, index magic makes it hard to read!

core/src/main/java/de/jplag/merging/MatchMerging.java
tsaglam: This should also get a more descriptive name: What buffer? What is its purpose?

tsaglam: If this method remove matches that are too short, let's name this accordingly.

core/src/main/java/de/jplag/merging/MatchMerging.java
tsaglam: For these three parameters it is hard to grasp what they are due to their name: int lengthUpper, int seperatingFirst, int seperatingSecond.

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -0,0 +1,136 @@
...
+            if (globalMatches.get(i).startOfFirst() > startFirst) {
+                Match alteredMatch = new Match(globalMatches.get(i).startOfFirst() - seperatingFirst, globalMatches.get(i).startOfSecond(),
+                        globalMatches.get(i).length());
+                globalMatches.set(i, alteredMatch);
+            }
+            if (globalMatches.get(i).startOfSecond() > startSecond) {
+                Match alteredMatch = new Match(globalMatches.get(i).startOfFirst(), globalMatches.get(i).startOfSecond() - seperatingSecond,
+                        globalMatches.get(i).length());
+                globalMatches.set(i, alteredMatch);
+            }
```
tsaglam, line 123: For me personally, it hard to understand this section, as it is a bit difficult due to naming, lack of comments, and index magic. Why not add the changes to a new list of matches here?

core/src/main/java/de/jplag/merging/MatchMerging.java
tsaglam: Ah now I understand, these are not instantiated once, they are instantiated for each comparison. This is due to relying on shared state between methods instead of passing values. I will add a comment regarding that as general review comment.

tsaglam (changes requested):
This is a very exhaustive review, be aware. One underlying issue I see is the anti-pattern of relying too much on fields instead of passing via parameters. All your private methods are void, thus, do not return a result. Instead, they read and write into the shared state, mainly the global matches list. I think we should refactor this so the global matches are passed as a parameter instead, and it is no longer a field. Then, the return values of the private methods should be used. `computeNeighbors` should return the neighbors. `removeToken` should return the list of altered matches (thus solving a problem described in one of the comments below).  Similar for the others.

core/src/main/java/de/jplag/merging/MatchMerging.java
uuqjz: All it does, is move the starting positions of matches that are over the merge-point(is clear what I mean by this?) down by how many tokens were removed from the list.

I don't understand what your suggestion is.

core/src/main/java/de/jplag/merging/MatchMerging.java
uuqjz: seperating is the average of how many tokens separate neighbouring matches in the first submission and how many separate them in the second

core/src/main/java/de/jplag/Submission.java
```diff
@@ -294,4 +294,14 @@ private List<Integer> getOrder(List<Token> tokenList) {
         }
         return order;
     }
+
+    /**
+     * @return Submission containing a deep copy of the tokenList and shallow copies of the remaining fields.
```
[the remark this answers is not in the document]
uuqjz: After reading through docs.oracle.com I think you are right.
Using ArrayList in this way only yields a shallow copy, which suffices for my use case.

Should i just edit the JavaDoc?

core/src/main/java/de/jplag/merging/MatchMerging.java
uuqjz: Would renaming it TokenBetweenNeighbours resolve this for you?

core/src/main/java/de/jplag/merging/MatchMerging.java
tsaglam: - Regarding naming: Adapt separatingFirst and separatingSecond
- Regarding comments: Your explanation here is could be a line comment, but you should explaining merge-point further.
- Regarding adding the changed matches to a list, see the global comment.

core/src/main/java/de/jplag/merging/MatchMerging.java
uuqjz: Neighbors contains pairs of neighbouring matches
If there are matches 1,2,3,4 (in Order), neighbours would contain (1,2),(2,3),(3,4)

core/src/main/java/de/jplag/merging/MatchMerging.java
uuqjz: Ah, thats what you meant

core/src/main/java/de/jplag/merging/MatchMerging.java
tsaglam: How about naming `separating` to `averageTokensBetweenMatches`? I think that helps to understand it better.
the `seperatingFirst` and `seperatingSecond` could be `tokenBetweenFirst` and `tokensBetweenSecond`.
I think one problem is that first/second is ambiguous, because it can both refer to the matches of a single token sequence, but also to the two sequences in the comparison itself. One idea would be to use left/right for the token sequences, and the first, second etc. only regarding the matches in a sequence and their relation.

core/src/test/java/de/jplag/merging/MergingTest.java
uuqjz: Could you elaborate a bit?
I am not that firm with JUnit

core/src/main/java/de/jplag/Submission.java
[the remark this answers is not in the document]
tsaglam: Yes, just adapting the comment should be fine.

core/src/main/java/de/jplag/merging/MatchMerging.java
uuqjz: The ambiguity was the reason I initially named it left, right, upper and lower.
But I changed that due to a request from Alex.

core/src/test/java/de/jplag/merging/MergingTest.java
tsaglam: The confusing thing is, that they changed the names between JUnit 4 and 5, but this guide should help: https://www.baeldung.com/junit-before-beforeclass-beforeeach-beforeall#beforeeach-and-beforeall

core/src/main/java/de/jplag/merging/MatchMerging.java
uuqjz: You are talking about matches m1 and m2?

cli/src/main/java/de/jplag/cli/CliOptions.java
```diff
@@ -109,6 +112,17 @@ public static class ClusteringEnabled {
         }
     }
 
+    public static class Merging {
+        @Option(names = {
+                "--merge-buffer"}, description = "Defines how much lower the length of a match can be than the minimum match length (default: 0)\n")
```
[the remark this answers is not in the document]
uuqjz: 0,0 is default at the moment as it "disables" match merging.
If we add a new bool as discussed for enabling it, these defaults can be changed

core/src/main/java/de/jplag/merging/MatchMerging.java
uuqjz: I don't see a reason why it cant be a for each.
I only use 'i' for accessing the current comparison.

I'll change it

uuqjz: `for (JPlagComparison comparison : comparisons)`
You want something like this, right?

uuqjz: Oh, i do use the 'i' here:
`comparisons.set(i, new JPlagComparison(firstSubmission, secondSubmission, globalMatches, new ArrayList<>()));`

What shall i do about it?

uuqjz:
@tsaglam can you take a look at the code smell?

My idea would be to make removeBuffer a void function but that's contrary to one of your requests changes.

core/src/main/java/de/jplag/merging/MatchMerging.java
uuqjz: What do you have in mind?
`MatchMerging.mergeMatches()` 
would be kinda redundant.

cli/src/main/java/de/jplag/cli/CliOptions.java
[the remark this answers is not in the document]
uuqjz: I've added said toggle

core/src/main/java/de/jplag/merging/MatchMerging.java
tsaglam: Yes, we try to avoid abbreviations.

uuqjz: Does match1 and match2 work for you?
first and second could be ambiguous.

core/src/main/java/de/jplag/merging/MatchMerging.java
tsaglam: You do not need to use .set here if you use a new list and add. Try to consider the review feedback and come up with an improved version.

core/src/main/java/de/jplag/merging/MatchMerging.java
tsaglam: It is not a static method, so if it is redundant or not entirely depends on the user of your code.
Could also be `mergeMatchesOf` as you will change this to a method with a parameter anyways.

core/src/test/java/de/jplag/merging/MergingTest.java
tsaglam: The test subject should never be shared between test cases, as the introduction of a bug to the match merging might change its behavior. At the very least lines 44 and 45 should be created before each test.

tsaglam:
> can you take a look at the code smell?

This is caused by the habit in the code to re-use input collections of a method as output collection. This should be avoided because it can lead to side-effects that were not considered.

uuqjz:
> > can you take a look at the code smell?
> 
> This is caused by the habit in the code to re-use input collections of a method as output collection. This should be avoided because it can lead to side-effects that were not considered. E.g. someone changing the returned list can modify the input list.

I think a simple return new ArrayList<>(originalList) would solve this issue

cli/src/main/java/de/jplag/cli/CLI.java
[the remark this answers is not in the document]
uuqjz: @tsaglam what do you think?

cli/src/main/java/de/jplag/cli/CliOptions.java
```diff
@@ -109,6 +112,17 @@ public static class ClusteringEnabled {
...
+        public int mergeBuffer;
+
+        @Option(names = {
+                "--seperating-threshold"}, description = "Defines how many token there can be between two neighboring matches (default: 0)\n")
```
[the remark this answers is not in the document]
uuqjz: Sure, --gap-size sounds good enough.
What about --merge-buffer, can we keep this?

core/src/main/java/de/jplag/merging/MatchMerging.java
uuqjz: I could use a pair for neighbors intead of an array:
`https://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/tuple/Pair.html`
This would remove one level of index magic, as pairs can be accessed via getLeft() and getRight()

uuqjz:
As discussed with Timur:
`--merge-buffer` and `--seperating-threshold` will be renamed and have their default values changed in another PR.

core/src/main/java/de/jplag/JPlag.java
```diff
@@ -71,6 +72,12 @@ public static JPlagResult run(JPlagOptions options) throws ExitException {
 
         // Compare valid submissions.
         JPlagResult result = comparisonStrategy.compareSubmissions(submissionSet);
+
+        // Use Match Merging against obfuscation
+        if (options.mergingParameters().enable()) {
```
tsaglam, line 77: Tiny detail I just noticed: We could rename this to `enabled` (with a d), then it matches the clustering options.

core/src/main/java/de/jplag/merging/MergingParameters.java
```diff
@@ -0,0 +1,25 @@
+package de.jplag.merging;
+
+/**
+ * Collection of parameters that describe how a match merging should be performed.
+ * @param mergeBuffer describes how shorter a match can be than the Minimum Token Match (Defaults to 0).
+ * @param seperatingThreshold describes how many tokens can be between to neighboring matches (Defaults to 0).
+ */
+public record MergingParameters(boolean enable, int mergeBuffer, int seperatingThreshold) {
+
+    public MergingParameters() {
```
tsaglam, line 10: JDoc to describe default values.

core/src/main/java/de/jplag/merging/MergingParameters.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public MergingParameters withEnable(boolean enable) {
+        return new MergingParameters(enable, mergeBuffer, seperatingThreshold);
+    }
+
+    public MergingParameters withMergeBuffer(int mergeBuffer) {
+        return new MergingParameters(enable, mergeBuffer, seperatingThreshold);
+    }
+
+    public MergingParameters withSeperatingThreshold(int seperatingThreshold) {
```
tsaglam, line 22: JDoc

core/src/main/java/de/jplag/merging/Neighbor.java
```diff
@@ -0,0 +1,9 @@
+package de.jplag.merging;
+
+import de.jplag.Match;
+
+/*
+ * This class realizes a pair of neighboring matches, named upperMatch and lowerMatch
```
tsaglam, line 6: Could you add your definition of neighbors (both sides next to each other, without interrupting matches)? Remember in our meeting when you asked me for four matches what are the neighbors? To avoid ambiguity, just explain it here.

tsaglam (approved):
Looks great, just 4 minor things, and then we can merge! 👍

Kr0nox (approved):

uuqjz:
@tsaglam 
I made the last changes.
Let's merge!

core/src/main/java/de/jplag/merging/MatchMerging.java
```diff
@@ -110,6 +113,46 @@ private List<Match> mergeNeighbors(List<Match> globalMatches, Submission leftSub
...
+
+        for (Token token : tokenRight) {
+            if (token.getType().equals(SharedTokenType.FILE_END)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
```
tsaglam, line 154: Could be also:
```java
return containsFileEndToken(tokenLeft) || containsFileEndToken(tokenRight) 
```

with
```java
private boolean containsFileEndToken(List<Token> token) {
    return token.stream().map(Token::getType).anyMatch(it->it.equals(SharedTokenType.FILE_END);
}
```

(reduces duplication, makes it more compact)

tsaglam (approved):
Sorry, I reviewed it early as I missed that the file end behavior was not in there. Now everything is in, right? One minor thing regarding duplication.

core/src/main/java/de/jplag/merging/MatchMerging.java
uuqjz: Does containsFileEndToken need JDoc?

tsaglam: No, I think it is self-explaining enough.

uuqjz:
I don't have merging rights, so you have to click the button

tsaglam:
> I don't have merging rights, so you have to click the button

Yes, that is intentional. 👍

PR #1204, opened 2023-07-25
Updated the java tests to use the new test framework

TwoOfTwelve:
This PR updates the java tests to use the new test framework and also adds a new tests using the CLI class from jplag, to have a more complex test case.

I also fixed a small bug in the language framework, where the TRY_END token was added to early.


tsaglam (changes requested):
Looks good, but could you try to preserve the documentation? Makes it easier to read. Se comments below.
Test values seem fine. As discussed, adapt the expected values in another commit.

tsaglam (approved):

PR #1222, opened 2023-08-05
Added the option to initialize the parser lazily for AbstractAntlrLan…

TwoOfTwelve:
For language-antlr-utils - AbstractAntlrLanguage:
Added the option to initialize the parser later in a separate method, to allow passing more data to the parser. To do that, you need to implement the initializeParser method and pass nothing to the super constructor.

### DIFF
--- a/language-antlr-utils/src/main/java/de/jplag/antlr/AbstractAntlrLanguage.java
+++ b/language-antlr-utils/src/main/java/de/jplag/antlr/AbstractAntlrLanguage.java
@@ -42,6 +42,10 @@ public abstract class AbstractAntlrLanguage implements Language {
         return this.parser.parse(files);
     }
 
+    /**
+     * Lazily creates the parser. Has to be implemented, if no parser is passed in the constructor.
+     * @return The newly initialized parser
+     */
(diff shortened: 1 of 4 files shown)


language-antlr-utils/src/main/java/de/jplag/antlr/AbstractAntlrLanguage.java
```diff
@@ -22,8 +25,25 @@ protected AbstractAntlrLanguage(AbstractAntlrParserAdapter<?> parser) {
...
     @Override
     public List<Token> parse(Set<File> files) throws ParsingException {
+        if (this.parser == null) {
+            this.parser = this.initializeParser();
+        }
+
         return this.parser.parse(files);
     }
+
+    protected AbstractAntlrParserAdapter<?> initializeParser() {
```
Kr0nox, line 45: Maybe there shoudl also be a doc comment here about it needing to be implemented when using the empty constructor and it not being needed when using the other one

Kr0nox (changes requested):

Kr0nox (approved):

tsaglam (approved):

Issue #1235, opened 2023-08-10
Possible issue with the mapRange function of the AbstractAntlrListener class


NiklasHeneka:
There is a possible issue with the mapRange function of the AbstractAntlrListener class or maybe I just misunderstand something 
@TwoOfTwelve.

After updating the LLVM IR language module to the new structure I saw that the printing of the token was different.

In the old structure you could use following method and for printing tokens with descriptions that are too long it would generate the "|" at the end and not directly after it.
```
transformToken(<TokenType>, ctx.getStart(), ctx.getStop());
```
An example: 

```
17        %5 = alloca %struct.Struct, align 4
               |ALLOC                       |

18        store i32 0, ptr %1, align 4
          |STORE                     |

19        store i32 4, ptr %2, align 4
          |STORE                     |

20        %6 = load i32, ptr %2, align 4
               |LOAD                   |

21        %7 = add nsw i32 5, %6
               |ADD            |

22        store i32 %7, ptr %3, align 4
          |STORE                      |

23        %8 = load i32, ptr %2, align 4
               |LOAD                   |

24        %9 = sub nsw i32 5, %8
               |SUB            |
```
With the new language module structure neither mapRange nor mapEnter generate the second "|" for many token types.
It just looks like this:
```
18        %5 = alloca %struct.Struct, align 4
               |ALLOC

19        store i32 0, ptr %1, align 4
          |STORE

20        store i32 4, ptr %2, align 4
          |STORE

21        %6 = load i32, ptr %2, align 4
               |LOAD

22        %7 = add nsw i32 5, %6
               |ADD

23        store i32 %7, ptr %3, align 4
          |STORE

24        %8 = load i32, ptr %2, align 4
               |LOAD

25        %9 = sub nsw i32 5, %8
               |SUB
```

Although for other tokens it does place a second "|" at the end.
I am not sure if this is a bug or I am just misunderstanding something.

tsaglam:
Closed, as this was addressed by #1236.

PR #1236, opened 2023-08-10
Fixed an error when extracting the token length

TwoOfTwelve:
Addresses #1235

The code, that generated the tokens did not call the right method.


tsaglam (approved):

mbrdl:
I'm confused about this. @TwoOfTwelve can you explain?

Issue #1001, opened 2023-03-30
Add minimal update notifications


sebinside:
To raise awareness for newer JPlag versions, the program should check its version against the current release using the public GitHub API. If a newer version is available, this should be shown in a small notification (that does **not** annoy the user). Alternatively, this could also be shown in a small info text in the report viewer.

tsaglam:
Ideally, this would be shown in the report viewer.

Kr0nox:
@sebinside This should be completed with #1240 if im not mistaken?

tsaglam:
Closed by #1240.

PR #1240, opened 2023-08-15
implement version info

Kr0nox:
Solves #1001 
In the FileUploadView in the bottom right corner the user gets a small info text about the current version (either version number, reference to the newest version or that is a development version (0.0.0)).

With this the Alert about using the develoment version was also removed, since it is now indicated in the FileUploadView directly.

The version type was modified to be a class that has usefull functions (like comparing versions and printing them as strings) to clean up the code


sebinside (approved):
I like the simple solution, thank you. There are only two minor comments. Besides that, you can merge this PR after resolving the conflicts.

PR #1251, opened 2023-08-20
Adjusted behavior, default values and naming of MergingParameters

uuqjz:
This PR does 3 things:

1. Change mergeBuffer from being relative to MTM to an absolute Value
2. Set sensible default values for mergeBuffer and seperatingThreshold
3. Rename mergeBuffer to neighborLength and separatingThreshold to gapSize

### DIFF
--- a/cli/src/main/java/de/jplag/cli/CliOptions.java
+++ b/cli/src/main/java/de/jplag/cli/CliOptions.java
@@ -59,7 +59,7 @@ public class CliOptions implements Runnable {
     @ArgGroup(validate = false, heading = "Clustering%n")
     public Clustering clustering = new Clustering();
 
-    @ArgGroup(validate = false, heading = "Match Merging defense that merges neighboring matches to increase the similarity between plagiarisms:%n")
+    @ArgGroup(validate = false, heading = "Merging of neighboring matches to increase the similarity of concealed plagiarism:%n")
     public Merging merging = new Merging();
 
     /**
@@ -116,7 +116,7 @@ public class CliOptions implements Runnable {
         @Option(names = {"--match-merging"}, description = "Enables match merging (default: false)%n")
         public boolean enabled;
 
-        @Option(names = {"--neighbor-length"}, description = "Defines how low the length of a match can be, to be considered (default: 2)%n")
+        @Option(names = {"--neighbor-length"}, description = "Defines how short a match can be, to be considered (default: 2)%n")
         public int neighborLength;
 
         @Option(names = {"--gap-size"}, description = "Defines how many token there can be between two neighboring matches (default: 6)%n")
--- a/cli/src/main/java/de/jplag/cli/CliOptions.java
+++ b/cli/src/main/java/de/jplag/cli/CliOptions.java
@@ -116,11 +116,11 @@ public class CliOptions implements Runnable {
         @Option(names = {"--match-merging"}, description = "Enables match merging (default: false)%n")
         public boolean enabled;
 
-        @Option(names = {"--merge-buffer"}, description = "Defines how low the length of a match can be, to be considered (default: 0)%n")
+        @Option(names = {"--merge-buffer"}, description = "Defines how low the length of a match can be, to be considered (default: 2)%n")
         public int mergeBuffer;
 
         @Option(names = {
-                "--seperating-threshold"}, description = "Defines how many token there can be between two neighboring matches (default: 0)%n")
+                "--seperating-threshold"}, description = "Defines how many token there can be between two neighboring matches (default: 6)%n")
         public int seperatingThreshold;
 
     }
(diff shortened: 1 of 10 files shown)


cli/src/main/java/de/jplag/cli/CliOptions.java
```diff
@@ -59,7 +59,7 @@ public class CliOptions implements Runnable {
     @ArgGroup(validate = false, heading = "Clustering%n")
     public Clustering clustering = new Clustering();
 
-    @ArgGroup(validate = false, heading = "Match Merging defense mechanism against obfuscation that merges neighboring matches based on these parameters:%n")
+    @ArgGroup(validate = false, heading = "Match Merging defense that merges neighboring matches to increase the similarity between plagiarisms:%n")
```
tsaglam, line 62:
```suggestion
    @ArgGroup(validate = false, heading = "Merging of neighboring matches to increase the similarity of concealed plagiarism:%n")
```

cli/src/main/java/de/jplag/cli/CliOptions.java
```diff
@@ -116,13 +116,11 @@ public static class Merging {
         @Option(names = {"--match-merging"}, description = "Enables match merging (default: false)%n")
         public boolean enabled;
 
-        @Option(names = {
-                "--merge-buffer"}, description = "Defines how much lower the length of a match can be than the minimum match length (default: 0)%n")
-        public int mergeBuffer;
+        @Option(names = {"--neighbor-length"}, description = "Defines how low the length of a match can be, to be considered (default: 2)%n")
```
tsaglam, line 119: How long? Or how short?

core/src/main/java/de/jplag/merging/MergingParameters.java
```diff
@@ -2,17 +2,17 @@
...
-public record MergingParameters(boolean enabled, int mergeBuffer, int seperatingThreshold) {
+public record MergingParameters(boolean enabled, int neighborLength, int gapSize) {
 
     /**
-     * The default values of MergingParameters are false for the enable-switch and 0 for both mergeBuffer and
-     * seperatingThreshold. These completely deactivate MatchMerging.
+     * The default values of MergingParameters are false for the enable-switch, which deactivate MatchMerging, while
+     * neighborLength and gapSize default to (2,6), which in testing yielded the best results.
      */
     public MergingParameters() {
```
tsaglam, line 14: `MergingOptions` would be more in lien with the `ClusteringOptions`.

core/src/main/java/de/jplag/merging/MergingParameters.java
```diff
@@ -2,17 +2,17 @@
 
 /**
  * Collection of parameters that describe how a match merging should be performed.
- * @param mergeBuffer describes how shorter a match can be than the Minimum Token Match (Defaults to 0).
- * @param seperatingThreshold describes how many tokens can be between to neighboring matches (Defaults to 0).
+ * @param neighborLength describes how short a match can be, to be considered (Defaults to 2).
+ * @param gapSize describes how many tokens can be between to neighboring matches (Defaults to 6).
  */
-public record MergingParameters(boolean enabled, int mergeBuffer, int seperatingThreshold) {
+public record MergingParameters(boolean enabled, int neighborLength, int gapSize) {
```
tsaglam, line 8: `minimumNeighborLength` would match the naming scheme of the `minimumTokenMatch`.

tsaglam (changes requested):
Just a few thoughts regarding the naming.

core/src/main/java/de/jplag/merging/MergingParameters.java
uuqjz: What about the cli option?
Wont `--minimum-neighbor-length` be too long

cli/src/main/java/de/jplag/cli/CliOptions.java
uuqjz: short would be better

core/src/main/java/de/jplag/merging/MergingParameters.java
uuqjz: Following the same logic, `gapSize` should be `maximumGapSize`, which would get tedious to write.
Ultimately it is your decision.

core/src/main/java/de/jplag/merging/MergingParameters.java
uuqjz: Do you know an efficient way to rename the file, class and all references in eclipse?

core/src/main/java/de/jplag/merging/MergingParameters.java
uuqjz: So `maximumGapSize`, `--maximum-gap-size` `minimumNeighborLength`and `--minimum-neighbor-length` ?

tsaglam: for the CLI flags, we could leave minimum and maximum away, as we can add it directly to the flag description. They are more often written than read.

tsaglam (approved):

PR #1252, opened 2023-08-20
update the report viewer to new chart library

Kr0nox:
The vue wrapper of chart.js which we currently use ([vue-chart-3](https://github.com/victorgarciaesgi/vue-chart-3)) is no longer maintained and runs on an over two year old version of chart js (version 3.1.0)

To update chartjs to version 4 and also have a maintained library with all new features this PR proposes the change to [vue-chartjs](https://github.com/apertureless/vue-chartjs). This library also provides better type safety on the options and input data.

All components using charts were updated to the new library


sebinside (commented):
Regarding the code, this PR is fine. However, I noticed a bug in the cluster view in this branch: Changing the entry in the drop-down menu of the radar chart does not update the radar chart. Please have a look at this. Afterward, this PR can be merged.

sebinside (approved):

PR #1264, opened 2023-08-29
Updates to report viewer code styling

Kr0nox:
This PR adds sorting of the tailwind classes through prettier.

In addition the pre commit hook now also effects JavaScript, HTML and CSS which it previosly did not.


sebinside (approved):
@Kr0nox I approved all assigned PRs (minus that one with the buggy radar chart); please feel free to merge them in your preferred order.

PR #1275, opened 2023-08-31
Rework Code displaying

Kr0nox:
This PR reworks the components involved in code displaying, to improve code quality and fix some bugs. It also improved the code in some other areas.

The following bugs have been fixed:
- Syntax highlighting in the comparison view seems to be broken for JDoc comments
- Fix the indent of code with different line numbers
- Matches sometimes requiere two clicks to become visible in the comparison view

We no longer highlight every line individually (`LineOfCode.vue`) and rather highlight the entire code in `CodePanle.vue`. Here we now display the code and line numbers in a table to ensure that all lines start at the indetation.
Storing matches and the connection scrolling to the matching Match have been simplified alot, by not storing a component to scroll to, but rather having the components expose a function for scrolling to a certain file/line


sebinside (approved):
Here are some nitty-gritty details regarding the code view:
- Please have a look at the bug report from sonarcloud whether or not it is applicable
- Clicking on an entry from the top token list at the top should probably scroll to the code snippet in both views
- The currently selected view (left or right) gets scrolled using the scroll wheel even if the mouse is over the other side which is probably confusing to the user

Apart from that, this PR can be merged. I also noticed, that at least for me, the current version of the code view of the `develop` branch is broken.

Kr0nox:
> Please have a look at the bug report from sonarcloud whether or not it is applicable

One of them I fixed, for the other I would have said its not applicable since we are using a table as its inteded purpose to structure data (in our case code) and we dont want a header in this place. So the remaining issue is not applicable in my opinion

> Clicking on an entry from the top token list at the top should probably scroll to the code snippet in both views

Chromium based browsers apperantly have a problem with smooth scrolling two divs. Edge especially since it just stops the scrolling in place. Fixed by not having smooth scrolling.

> The currently selected view (left or right) gets scrolled using the scroll wheel even if the mouse is over the other side which is probably confusing to the user

I could only reproduce this bug in Edge. All other browsers do not have this issue. Beacause of this and since it is a problem I could also reproduce on other websites having two scrollable containers (also exclusivly on edge), I would have deamed this an issue with Edge.

Kr0nox:
> Apart from that, this PR can be merged. I also noticed, that at least for me, the current version of the code view of the develop branch is broken.

I now what the error is. I could do a fixing PR. On the other hand these changes in the PR would be overwritten by this one anyways.

sebinside (approved):

PR #1279, opened 2023-09-05
Overhauled the end to end tests and added the test code from progpedia

TwoOfTwelve:
Reworked the End-to-End tests. They now use a data set descriptor to identify the sources to be tested.

Also added the progpedia code to the end to end tests.


dfuchss:
@TwoOfTwelve could you take a look into the failing build :) ?

tsaglam (commented):
With my generated expected test values, the E2E tests still fail, but due to:
`Error:    Run 4646: EndToEndSuiteTest.lambda$generateGoldStandardTest$6:214 The gold standard tests are skipped, because no gold standard is defined`

tsaglam (commented):
Just a summary of the recent discussions:

- GST changes should be separate PR
- Datsets should be zipped due to bloat
- End-to-end test results should use language regarding plagiarisms and non-plagiarisms and include if the value is higher or lower than expected

tsaglam (changes requested):
A few comments.

tsaglam (approved):

PR #1290, opened 2023-09-07
Store selections from overview when switching views

Kr0nox:
This PR adds the storing of the selected metrics and resolution of the x-scale on the distribution diagramm when switching views


report-viewer/src/views/OverviewView.vue
```diff
@@ -194,6 +196,14 @@ function getMetricFromNumber(metric: number) {
   }
 }
 
+function getIndexFromMetric(metric: MetricType) {
+  if (metric == MetricType.AVERAGE) {
+    return 0
+  } else {
+    return 1
+  }
+}
```
dfuchss, line 205: Is it possible in TS to define that more extensible ? Seems like some hard-coding :)

sebinside:
#1292 has been merged so you can update this PR @Kr0nox

sebinside (approved):

PR #1291, opened 2023-09-08
Add parse and compare hooks

olmokramer:
This adds 4 hooks that can be set on the `JPlagOptions` class:

- A pre-parsing stage hook, which is called before all submissions are parsed. It is called with the list with all submission paths eligible for the comparison, i.e. all paths that are not filtered out by `SubmissionSetBuilder.isExcludedEntry`.
- A post-parsing hook per submission, which is called after the submission has been parsed. It is called with the `Submission` object that has been parsed.
- A pre-comparing stage hook, which is called before the comparisons are started. It is called with a list of all comparison tuples.
- A post-comparing hook, which is called after each comparison. It is called with the `SubmissionTuple` that has just been compared.

The default for each hook is to do nothing.

For some background, we are using JPlag for plagiarism checking and want to be able to track the progress of the plagiarism check (e.g. the percentage of submissions that have already been parsed, or the amount of pairs that have already been compared). With these hooks we can implement that, but simultaneously they might be used for other unforeseen use cases.

Looking forward to feedback on the concept and the implementation.

### DIFF
--- a/core/src/test/java/de/jplag/HooksTest.java
+++ b/core/src/test/java/de/jplag/HooksTest.java
@@ -1,0 +1,53 @@
+package de.jplag;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.junit.jupiter.api.Test;
+
+import de.jplag.exceptions.ExitException;
+
+class HooksTest extends TestBase {
+    private static final String ROOT = "NoDuplicate";
+    private static final int SUBMISSION_COUNT = 3;
+    private static final int COMPARISON_COUNT = 3;
+
+    @Test
+    void testPreParseHook() throws ExitException {
+        final AtomicInteger timesCalled = new AtomicInteger(0);
+        runJPlag(ROOT, it -> it.withPreParseHook(submissions -> {
+            timesCalled.incrementAndGet();
+            assertEquals(SUBMISSION_COUNT, submissions.size());
+        }));
+        assertEquals(1, timesCalled.get());
+    }
+
+    @Test
+    void testParseHook() throws ExitException {
+        final AtomicInteger totalSubmissions = new AtomicInteger(0);
+        runJPlag(ROOT, it -> it.withParseHook(submission -> {
+            totalSubmissions.incrementAndGet();
+        }));
+        assertEquals(SUBMISSION_COUNT, totalSubmissions.get());
+    }
+
+    @Test
+    void testPreCompareHook() throws ExitException {
+        final AtomicInteger timesCalled = new AtomicInteger(0);
+        runJPlag(ROOT, it -> it.withPreCompareHook(comparisons -> {
+            timesCalled.incrementAndGet();
+            assertEquals(COMPARISON_COUNT, comparisons.size());
+        }));
+        assertEquals(1, timesCalled.get());
+    }
+
+    @Test
+    void testCompareHook() throws ExitException {
+        final AtomicInteger totalComparisons = new AtomicInteger(0);
+        runJPlag(ROOT, it -> it.withCompareHook(comparison -> {
+            totalComparisons.incrementAndGet();
+        }));
+        assertEquals(COMPARISON_COUNT, totalComparisons.get());
+    }
+}


core/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -225,6 +258,46 @@ private Integer normalizeMinimumTokenMatch(Integer minimumTokenMatch) {
         return (minimumTokenMatch != null && minimumTokenMatch < 1) ? Integer.valueOf(1) : minimumTokenMatch;
     }
 
+    public interface PreParseHook {
```
dfuchss, line 261: You could simply use the Functional Interface Consumer<List<Submission>>

This would reduce the amount of newly created interfaces.

core/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -225,6 +258,46 @@ private Integer normalizeMinimumTokenMatch(Integer minimumTokenMatch) {
...
+        void callback(List<Submission> submissions);
+    }
+
+    public static final PreParseHook DEFAULT_PRE_PARSE_HOOK = new PreParseHook() {
```
dfuchss, line 265: I'd suggest to use (submissionList) -> log.debug("PreParse {} submissions", submissionList.size()) or something similar as default

core/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -225,6 +258,46 @@ private Integer normalizeMinimumTokenMatch(Integer minimumTokenMatch) {
...
+        @Override
+        public void callback(List<Submission> submissions) {
+        }
+    };
+
+    public interface ParseHook {
```
dfuchss, line 271: Similar as above

core/src/main/java/de/jplag/strategy/AbstractComparisonStrategy.java
```diff
@@ -45,8 +45,16 @@ protected void compareSubmissionsToBaseCode(SubmissionSet submissionSet) {
...
+        return compareSubmissions(new SubmissionTuple(first, second));
+    }
+
+    protected Optional<JPlagComparison> compareSubmissions(SubmissionTuple tuple) {
+        Submission first = tuple.left();
+        Submission second = tuple.right();
+
         JPlagComparison comparison = greedyStringTiling.compare(first, second);
         logger.info("Comparing {}-{}: {}", first.getName(), second.getName(), comparison.similarity());
+        this.options.compareHook().callback(tuple);
```
dfuchss, line 57: You could use the [BiConsumer](https://docs.oracle.com/javase/8/docs/api/java/util/function/BiConsumer.html) interface to reduce refactorings :)

dfuchss (commented):
I made a short code review and supplied some suggestions, feel free to comment on them :)

core/src/main/java/de/jplag/options/JPlagOptions.java
olmokramer: Yes makes sense, that fits the goal of these hooks exactly.

core/src/main/java/de/jplag/options/JPlagOptions.java
olmokramer: I haven't done this, because we only have access to the `JPlagOptions` logger and so every line will start with `JPlagOptions` instead of the class where the hook was actually called from, and passing the `logger` to use into these hooks seems a bit clunky.

olmokramer:
@dfuchss thanks. I've now added some test cases for all the hooks as well.

dfuchss (commented):
Thanks for the tests. I've one minor suggestion regarding them.

core/src/test/java/de/jplag/HooksTest.java
```diff
@@ -0,0 +1,49 @@
...
+import de.jplag.exceptions.ExitException;
+
+public class HooksTest extends TestBase {
+    private static final String ROOT = "NoDuplicate";
+    private static final int SUBMISSION_COUNT = 3;
+    private static final int COMPARISON_COUNT = 3;
+
+    @Test
+    void testPreParseHook() throws ExitException {
+        final int[] totalSubmissions = {0};
```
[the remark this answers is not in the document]
olmokramer: Ah yes of course, this is also what we do in our implementation of the progress counters... I have changed the tests to use `AtomicInteger`. Thanks for the suggestion :)

dfuchss (approved):
I think the idea and realization is good.
Maybe you can add some docs to the public methods/fields you've created.

Afterwards, let's wait for the review of @tsaglam :)

dfuchss:
Some tests seem to fail :)

olmokramer:
@dfuchss Thanks again for the reviews. I have added documentation for the new fields on the `JPlagOptions` class.

> Some tests seem to fail :)

Whoops, shouldn't have copy-pasted that test... Fixed now :)

tsaglam:
Sorry for my absence, I was out of office. I will take a look at it as soon as I can find the time!

olmokramer:
@tsaglam Have you already had a chance to take a look at this?

olmokramer:
Thank you for your response and good luck with your work the upcoming month :-)

olmokramer:
@TwoOfTwelve thank you for the review! This is exactly the kind of feedback I was looking for :) I agree with the points you brought up and that the current implementation is not necessarily the best for the JPlag architecture.

I'm also on a quite tight schedule, so I don't know when I will have time to take a look at your suggested implementation. We are running a build of our fork with this PR, so there's no real rush from our side to get this in. I'd rather get it done in a way that is maintainable and sustainable. I'll try to make some time next week, though, to take a look.

Our requirements are not very complex. In particular, we need to be able to

- get the total amount of submissions that will be parsed, before parsing has started. Though it would also be fine to get this information as soon as the first submission has been parsed. (This is what the `preParseHook` in this PR is for).

- update a counter as soon as a submission has been parsed. (This is what the `parseHook` in this PR is used for)

- get the total amount of comparisons that will be made, before comparing has started. Though here too it would be fine to get this information directly after the first comparison. (This is what we use the `preCompareHook` from this PR for)

- update a counter as soon as a comparison between 2 submissions has finished. (We use the `compareHook` from this PR for that)

Thanks again for you consideration.

PR #1292, opened 2023-09-10
Tooltips in report viewer

Kr0nox:
This PR adds an easy way of adding tooltips to the report viewer.
In addition it adds a generalized selector for metrics.

Some components already got tooltips. more can be added in either this or a seperate PR.
Due to the growing amount of components, I moved the components related to the options selector into its own sub folder


sebinside (approved):
We should discuss the fade-in delay and the content of the tooltips (and possibilities for more helpful tooltips!) in a future meeting with @tsaglam, but for now, this can be merged.

PR #1301, opened 2023-09-14
Allow passing multiple options in the "Advanced" group

olmokramer:
By default, `picocli` sets `exclusive = true` on `ArgGroup`s, which means that it is not allowed to pass multiple options within the group [1].

[1] https://picocli.info/#_mutually_exclusive_options


tsaglam (approved):

olmokramer:
Thanks @tsaglam and @TwoOfTwelve !

PR #1311, opened 2023-09-25
new color generation for matches

Kr0nox:
This PR changes the color generation for matches.
Instead of distributing the colors across the hue spectrum, now there is a set of 7 (handpicked) colors from which each match is colored.
The algortihm for choosing colors: First all matches get sorted 3 times. (By line number in first and second submission and by number of tokens).  It then runs through the list of matches and chooses the first color it can find that is not neigboring this match in any of the three lists. To avoid having only the first few colors or even only two alternating colors, we dont start with the first color when having found a fitting color but go on with the next in the list and go through the list cyclical.

### DIFF
--- a/report-viewer/src/model/Match.ts
+++ b/report-viewer/src/model/Match.ts
@@ -7,9 +7,9 @@
  * @property startInSecond - Starting line of the match in the second file.
  * @property endInSecond - Ending line of the match in the second file.
  * @property tokens - Number of tokens in the match.
- * @property color - Color of the match.
+ * @property colorIndex - Index of the color to use for the match.
  */
-export type Match = {
+export interface Match {
   firstFile: string
   secondFile: string
   startInFirst: number
@@ -17,5 +17,5 @@ export type Match = {
   startInSecond: number
   endInSecond: number
   tokens: number
-  color: string
+  colorIndex?: number
 }
--- a/report-viewer/src/model/factories/ComparisonFactory.ts
+++ b/report-viewer/src/model/factories/ComparisonFactory.ts
@@ -1,7 +1,7 @@
 import { Comparison } from '../Comparison'
 import type { Match } from '../Match'
 import { store } from '@/stores/store'
-import { generateColors } from '@/utils/ColorUtils'
+import { getMatchColorCount } from '@/utils/ColorUtils'
 import slash from 'slash'
 import { BaseFactory } from './BaseFactory'
 import { MetricType } from '../MetricType'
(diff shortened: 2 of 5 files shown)


Kr0nox:
We should take a look at the list of colors in the next dev meeting

Kr0nox:
The currently chosen colors
grafik

report-viewer/src/model/Match.ts
```diff
@@ -7,15 +7,15 @@
...
+export interface Match {
   firstFile: string
   secondFile: string
   startInFirst: number
   endInFirst: number
   startInSecond: number
   endInSecond: number
   tokens: number
-  color: string
+  colorIndex?: number
```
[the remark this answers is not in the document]
Kr0nox: This is an optional, since for the color comparison we need a number here that is no possible index. Another option would be putting something like -1 in the initialization here. To not cause an error I made `getMatchColor` return a fully transparent color on an undefined index.

We only save the index here, so we can set a different alpha for different occasions, for example for dark and light mode if it is needed

report-viewer/src/model/factories/ComparisonFactory.ts
```diff
@@ -35,19 +35,15 @@ export class ComparisonFactory extends BaseFactory {
...
+    const unColredMatches = matches.map((match) => this.getMatch(match))
 
     return new Comparison(
       firstSubmissionId,
       secondSubmissionId,
       this.extractSimilarities(json),
       filesOfFirstSubmission,
       filesOfSecondSubmission,
-      coloredMatches
+      this.colorMatches(unColredMatches)
```
sebinside, line 46: typo: `colred matches`?

report-viewer/src/model/factories/ComparisonFactory.ts
```diff
@@ -101,16 +97,55 @@ export class ComparisonFactory extends BaseFactory {
...
+      const startCounter = currentColorIndex
+      while (
+        !isColorAvailable(matchesFirst, firstIndex) ||
+        !isColorAvailable(matchesSecond, secondIndex) ||
+        !isColorAvailable(sortedSize, sortedIndex)
+      ) {
+        currentColorIndex = (currentColorIndex + 1) % maxColorCount
+
+        if (currentColorIndex == startCounter) {
+          throw 'No solution'
```
sebinside, line 143: Would this throw an error in production only called `No solution`? I'm not happy with this as this is unspecific and somewhat noisy. I think we agree, that this should never happen anyway, right? In this case, I would opt for a random index rather than a error message in production.

sebinside:
> The currently chosen colors grafik

If you don't have a strong reason against it, I would opt for using the orange in the upper right corner instead of the red-like color in the second row, as it is still distinct but, well, less red.

sebinside (changes requested):
Works fine, the new colors look great with real data! I only have minor and easy to implement change requests, please see the comments above.

report-viewer/src/model/factories/ComparisonFactory.ts
Kr0nox: It noew returns the index it started on. In addition I added a comment, that this should never happen and is just a safety meassure

sebinside:
Oh, my mistake: I did not mean second *row* but second *column*. That's what happens if you review to many PRs at once, sorry! Please see the attached picture: 
image

PR #1320, opened 2023-09-28
Wrapper for async Views

Kr0nox:
The router has some problems with the suspense block for async components.
Because of this, this PR adds a wrapper for all async components


sebinside (approved):
Great improvement, the bug seems to be fixed and the report viewer feels even more snappier. I like it!

PR #1321, opened 2023-09-28
Bump vue-tsc to 1.8.15

Kr0nox:
Replaces #1317 

Since vue-tsc now has more restrictions in emits, this PR implements these.


sebinside:
OK, but what about? https://github.com/jplag/JPlag/blob/6eb1da8eba7ea55b377a58d4098a512c94bd19cb/report-viewer/src/views/ComparisonView.vue#L54

Kr0nox:
Its is not standard to use camelCase for these html tags.

The [Vue Documentation](https://vuejs.org/guide/components/events.html#emitting-and-listening-to-events) also does it the way we currently do it

PR #1323, opened 2023-10-02
New Cluster Visuaization

Kr0nox:
This PR adds a new way to default way to visualize clusters. The default is a graph having the submissions in the cluster as nodes and the average similarity as edges. Edges can only exists between the comparisons in the top list.

The old radar chart visualization is still accassible through a selector at the top of the container.

Unbenannt

### DIFF
--- a/report-viewer/src/components/ClusterGraph.vue
+++ b/report-viewer/src/components/ClusterGraph.vue
@@ -13,6 +13,7 @@ import ChartDataLabels from 'chartjs-plugin-datalabels'
 import { EdgeLine, GraphController, GraphChart } from 'chartjs-chart-graph'
 import { store } from '@/stores/store'
 import { graphColors } from '@/utils/ColorUtils'
+import { start } from 'repl'
 
 const props = defineProps({
   cluster: {
@@ -85,8 +86,27 @@ const graphData = computed(() => {
   }
 })
 
+const yPadding = 40
+const xPadding = computed(() => {
+  const avgCharacterLength = 8
+
+  const widths = labels.value.map((label) => label.length * avgCharacterLength)
+  const maxWidth = Math.max(...widths)
+  // Space needed for the longest name, but at most 200
+  console.log(maxWidth)
+  return Math.min(200, maxWidth)
+})
+
 const graphOptions = computed(() => {
   return {
+    layout: {
+      padding: {
+        top: yPadding,
+        bottom: yPadding,
+        left: xPadding.value,
+        right: xPadding.value
+      }
+    },
     animation: false as false,
     plugins: {
       legend: { display: false },
@@ -94,12 +114,13 @@ const graphOptions = computed(() => {
         display: true,
         font: {
           weight: 'bold' as 'bold',
-          size: 16
+          size: 12
         },
         formatter: (value: any, ctx: any) => {
           return labels.value[ctx.dataIndex]
         },
-        offset: 12,
+        align: (ctx: any) => (-360 * ctx.dataIndex) / keys.value.length,
+        offset: 8,
         color: graphColors.ticksAndFont.value
       },
       tooltip: {
@@ -124,7 +145,6 @@ function drawGraph() {
     loaded.value = false
     return
   }
-  console.log(graphCanvas.value)
   chart.value = new Chart(ctx, {
     type: 'graph',
     data: graphData.value,
--- a/report-viewer/src/components/ClusterGraph.vue
+++ b/report-viewer/src/components/ClusterGraph.vue
@@ -1,0 +1,172 @@
+<template>
+  <div>
+    <canvas ref="graphCanvas"></canvas>
+    <div v-show="!loaded">Could not display graph</div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import type { ClusterListElement } from '@/model/ClusterListElement'
+import { Chart, registerables } from 'chart.js'
+import { ref, type PropType, type Ref, onMounted, computed, watch } from 'vue'
+import ChartDataLabels from 'chartjs-plugin-datalabels'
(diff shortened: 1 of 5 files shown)


tsaglam (commented):
I tested the feature with a few real-world datasets. Looks already really good! I have the following comments solely based on using the feature:

- When hiding submissions, the names are hidden in the list of cluster members but not in the graph view itself
- It would be cool if the cluster member list also supports filtering/searching
- It would be cool to have either a mouseover for edges showing the similarity or if a mouseover for the edges highlights the corresponding comparison in the cluster member list to the right
- If comparisons are not included in the report, then there are no lines in the graph view. Currently, this is not explained to the user. We could either introduce thin, gray, dashed lines for these missing edges or add a legend that states why these are missing (or both)
- the edges of the graph are in the foreground, it would probably look a bit better if the nodes are in the foreground
- In some clusters, it is hard to differentiate by the line opacity which edges are the important ones. Maybe we could scale the opacity based on the lowest and highest edges in the graph? (I have more ideas here; we can discuss them in the next dev meeting)
- unrelated: The radar view breaks when either there are <= 3 submissions in the cluster or when comparisons are not part of the report. We could disable the radar view in these cases.

Not all of these points need to be addressed in this PR, maybe we can discuss which ones we want to include directly and which ones we want in a separate PR.

sebinside:
In big clusters, I also wondered about the missing lines, see the attached image. I don't know what you discussed with @tsaglam, but this can of course also be a separate PR together with the other ideas by timur. If time allows it, these other improvements would be great for v5, nevertheless.

image

report-viewer/src/components/ClusterGraph.vue
```diff
@@ -0,0 +1,177 @@
...
+      datalabels: {
+        display: true,
+        font: {
+          weight: 'bold' as 'bold',
+          size: 12
+        },
+        formatter: (value: any, ctx: any) => {
+          return labels.value[ctx.dataIndex]
+        },
+        align: (ctx: any) => (-360 * ctx.dataIndex) / keys.value.length,
```
sebinside, line 122: Again magic math

report-viewer/src/components/ClusterGraph.vue
```diff
@@ -0,0 +1,177 @@
...
+        top: yPadding,
+        bottom: yPadding,
+        left: xPadding.value,
+        right: xPadding.value
+      }
+    },
+    animation: false as false,
+    plugins: {
+      legend: { display: false },
+      datalabels: {
```
sebinside, line 113: Is it possible to make the background color of the labels transparent? Sometimes, there is a little overlap of the label and the node of the graph which cuts of parts of the node's displayed circle

sebinside (changes requested):
Really great PR, could be easliy one of the biggest improvements of the JPlag clustering since its introduction. I only have minor comments, please see above.

report-viewer/src/components/ClusterGraph.vue
sebinside: Yes, this is what i meant. If there is no easy fix, let's ignore it for now

report-viewer/src/components/ClusterGraph.vue
```diff
@@ -0,0 +1,177 @@
...
+    datasets: [
+      {
+        pointRadius: 10,
+        pointHoverRadius: 10,
+        pointBackgroundColor: graphColors.contentFill,
+        pointHoverBackgroundColor: graphColors.contentFill,
+        pointBorderColor: graphColors.ticksAndFont.value,
+        pointHoverBorderColor: graphColors.ticksAndFont.value,
+        data: Array.from(keys.value).map((_, index) => ({
+          x: Math.cos((2 * Math.PI * index) / keys.value.length) + 1,
```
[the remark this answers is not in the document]
Kr0nox: Would you prefer this to be solved by extracting it into its own method or would you rather have a comment explaining it?

Kr0nox:
I have implemented the dashed line. (newest version [here](https://kr0nox.github.io/JPlag/)
The issues I see with it are
- Especially with large clusters it gets really overloaded
- The use of a potential mouse over will be really difficult to use, due to all the lines so close to each other

What are your opinions on that @tsaglam @sebinside ?

tsaglam:
> What are your opinions on that?

I think it looks good! On a large screen, the lines are not too crowded. On my smaller one it is a bit more dense but still fine! Ideally, we would not have mouseover/clicking for the dashed lines, only for the included ones.
Distinction between red lines is now also much better!

image

For some additional opinions: @dfuchss @dr6817

Kr0nox:
I have implemented that the row to the hovered edge in the tablle gets highlighted now. (accessile [here](https://kr0nox.github.io/JPlag/) again)  
Is that how you wanted it @tsaglam ?

Do you also want me to try and implement a tooltip in addition to that? If so what text would you want on that tooltip?

tsaglam:
> Is that how you wanted it 

Yes looks great! One minor thing: When an edge is at the bottom of the list, it seems at first glance that there is no mousover. Would it be possible to scroll down if the mouseover edge is not visible in the scroll view?

> Do you also want me to try and implement a tooltip in addition to that? If so what text would you want on that tooltip?

No, I think we do not need that anymore, as the highlighting fulfills all the needs.

One minor issue: When min=100% then the edges look somewhat thin. They should probably strong and red, as it is 100%:
image

dfuchss:
> > What are your opinions on that?
> 
> I think it looks good! On a large screen, the lines are not too crowded. On my smaller one it is a bit more dense but still fine! Ideally, we would not have mouseover/clicking for the dashed lines, only for the included ones. Distinction between red lines is now also much better!
> 
> image
> 
> For some additional opinions: @dfuchss @dr6817

It looks nice :)

dfuchss (approved):

sebinside (approved):
LGTM

PR #1331, opened 2023-10-10
Enhance comparison table

Kr0nox:
This PR adds 3 small enhancements for the report viewers comparison table.

- In the Comparison View there now is an option to sort by Cluster. This method works by grouping all shown comparisons in a cluster toghether in the table. Inside the cluster submissions are sorted by their average similaritys.  All clusters are sorted by  their average similarity.  All submissions that are not part of a cluster are shown below the clusters, also sorted by their average similarity
- It is now possible to hide and show usernames directly in the table. When hovering over a name in table it now shows an eye (or an eye with a slash) and clicking it will hide/show that name.
- It is now possible to insert content into a slot below the comparison table. This is currently used to display warnings, regarding missing comparisons due to the limit set in the CLI


sebinside (commented):
I'm fine with the changes. However, the sorting by cluster did behave somewhat intuitively when tested with real world data. We should this discuss in the next meeting.

sebinside (approved):

Kr0nox:
@sebinside It should behave like expected now.
The issue was that I first sorted by the similarity and then by the index of the cluster, so in the end the table was just sorted by the index. Switching them around should have solved the problem. Can you confirm thtat?

sebinside (approved):

PR #1349, opened 2023-10-24
Minimal Report Version

Kr0nox:
This PR adds a minimal requiered report version to the report viewer.
This version is displayed on the landing page.
Reports generated with a lower version will not be accepted

### DIFF
--- a/.github/workflows/report-viewer.yml
+++ b/.github/workflows/report-viewer.yml
@@ -24,9 +24,9 @@ jobs:
           MAJOR=$(echo $VERSION | cut -d '.' -f 1)
           MINOR=$(echo $VERSION | cut -d '.' -f 2)
           PATCH=$(echo $VERSION | cut -d '.' -f 3)
-          sed -i "/major/s/.*/    \"major\": $MAJOR,/" report-viewer/src/version.json
-          sed -i "/minor/s/.*/    \"minor\": $MINOR,/" report-viewer/src/version.json
-          sed -i "/patch/s/.*/    \"patch\": $PATCH/" report-viewer/src/version.json
+          sed -i "3 s/.*/    \"major\": $MAJOR,/" report-viewer/src/version.json
+          sed -i "4 s/.*/    \"minor\": $MINOR,/" report-viewer/src/version.json
+          sed -i "5 s/.*/    \"patch\": $PATCH/" report-viewer/src/version.json
           echo "Version of Report Viewer:"
           cat report-viewer/src/version.json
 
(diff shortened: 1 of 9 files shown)


report-viewer/src/model/factories/OverviewFactory.ts
```diff
@@ -203,29 +212,40 @@ export class OverviewFactory extends BaseFactory {
...
       sessionStorage.setItem('versionAlert', 'true')
     }
+    if (jsonVersion.compareTo(minimalVersion) < 0) {
+      throw (
+        "The result's version(" +
+        jsonVersion.toString() +
+        ') is older than the minimal support version of the report viewer(' +
+        reportViewerVersion.toString() +
+        '). ' +
+        'Can not read report.'
```
sebinside, line 247: typo: `...read the report`

.github/workflows/report-viewer.yml
```diff
@@ -24,9 +24,9 @@ jobs:
           MAJOR=$(echo $VERSION | cut -d '.' -f 1)
           MINOR=$(echo $VERSION | cut -d '.' -f 2)
           PATCH=$(echo $VERSION | cut -d '.' -f 3)
-          sed -i "/major/s/.*/    \"major\": $MAJOR,/" report-viewer/src/version.json
-          sed -i "/minor/s/.*/    \"minor\": $MINOR,/" report-viewer/src/version.json
-          sed -i "/patch/s/.*/    \"patch\": $PATCH/" report-viewer/src/version.json
+          sed -i "3 s/.*/    \"major\": $MAJOR,/" report-viewer/src/version.json
+          sed -i "4 s/.*/    \"minor\": $MINOR,/" report-viewer/src/version.json
+          sed -i "5 s/.*/    \"patch\": $PATCH/" report-viewer/src/version.json
```
sebinside, line 29: Are these hard coded line numbers of the `version.json`? We should at least add a comment to document the dependency

sebinside (commented):
Again only minor comments

.github/workflows/report-viewer.yml
Kr0nox: Yes. I did not find a better way of only editing the first occurance of each string. Maybe @dfuchss knows a better way?

dfuchss: According to some AI model we all know, something like that should work:

```bash
json=$(cat example.json)

json=$(echo "$json" | jq --arg MAJOR "$MAJOR" --arg MINOR "$MINOR" --arg PATCH "$PATCH" '.report_viewer_version |= { "major": $MAJOR | tonumber, "minor": $MINOR | tonumber, "patch": $PATCH | tonumber }')

echo "$json" > example.json

```

Kr0nox: Tested it on a fork an works just like expected👍  Thank you

sebinside (approved):

PR #1351, opened 2023-10-25
Made the GreedyStringTiling class deterministic in case both submissi…

TwoOfTwelve:
The greedy string tiling class used to result in an arbitrary comparison direction if both submissions had the same amount of tokens. This did result in noticeable discrepancies in the similarity (see https://github.com/jplag/JPlag/pull/1279). This PR fixes that by using the submission name as a tie breaker.

Also made some small changes removing warnings.

### DIFF
--- a/core/src/main/java/de/jplag/GreedyStringTiling.java
+++ b/core/src/main/java/de/jplag/GreedyStringTiling.java
@@ -78,8 +78,10 @@ public class GreedyStringTiling {
     public final JPlagComparison compare(Submission firstSubmission, Submission secondSubmission) {
         Submission smallerSubmission;
         Submission largerSubmission;
-        Comparator<Submission> comp = Comparator.comparing((Submission s) -> s.getTokenList().size()).thenComparing(Submission::getName);
-        if (comp.compare(firstSubmission, secondSubmission) <= 0) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/de/jplag/GreedyStringTiling.java
```diff
@@ -78,12 +78,13 @@ public final JPlagComparison generateBaseCodeMarking(Submission submission, Subm
     public final JPlagComparison compare(Submission firstSubmission, Submission secondSubmission) {
         Submission smallerSubmission;
         Submission largerSubmission;
-        if (firstSubmission.getTokenList().size() > secondSubmission.getTokenList().size()) {
-            smallerSubmission = secondSubmission;
-            largerSubmission = firstSubmission;
-        } else {
+        Comparator<Submission> comp = Comparator.comparing((Submission s) -> s.getTokenList().size()).thenComparing(Submission::getName);
```
tsaglam, line 81: `sizeComparator` or something similar, to avoid abbreviations.

tsaglam: Also probably `it` or `submission` instead of `s`?

tsaglam (changes requested):
Just one small thing...

tsaglam (approved):

PR #1360, opened 2023-10-28
Improve comparison view

Kr0nox:
This PR adds an collapse all button to both file containers.

The code is also cleaned up a bit, by moving all the files related to the displaying of code into its own folder.


Kr0nox:
The bug acknowledged by sonarcloud is the same one already discussed in #1275

sebinside (approved):
Everything is fine. For a future version, we could think about giving the collapse button an additional collapse icon before the text.

Kr0nox:
> The bug acknowledged by sonarcloud is the same one already discussed in https://github.com/jplag/JPlag/pull/1275

Kr0nox:
> The bug acknowledged by sonarcloud is the same one already discussed in https://github.com/jplag/JPlag/pull/1275

@tsaglam Is there a way of "removing" this issue from sonarcloud? I think you did something similar with a Java Code Smell once

Kr0nox:
> giving the collapse button an additional collapse icon before the text.

I didnt really find an icon that I found 100% fitting.
What do you thing about [chevron-up](https://fontawesome.com/icons/chevron-up?f=classic&s=solid) from font-awesome?

sebinside:
I would rather go with one of the minimize icons: https://fontawesome.com/search?q=minimize&o=r
Maybe two arrows are better as four arrows could look to much, but you have to try

PR #1371, opened 2023-11-07
Add normalization info to cpp tokens

mbrdl:
Enables using token string normalization for CPP code (didn't test thoroughly for effectiveness, but stability is given).
I also changed some other, minor things which should be merged:
- Added logging for normalization and fixed mistaken usage of java.util.Logger
- Normalization is no longer performed in-place and instead returns a new array
- Added extra checks in grammar handling to avoid crashes in special cases
- Split cpp TRY token into TRY_BEGIN and TRY_END


tsaglam:
> I also changed some other, minor things which should be merged.

@mbrdl Can you briefly elaborate on these changes in the PR description?

mbrdl:
@tsaglam Done

core/src/main/java/de/jplag/Submission.java
```diff
@@ -247,6 +247,11 @@ private static File createErrorDirectory(String... subdirectoryNames) {
 
         try {
             tokenList = language.parse(new HashSet<>(files));
+            for (Token token : tokenList) {
+                if (logger.isDebugEnabled()) {
+                    logger.debug(String.join(" | ", token.getType().toString(), Integer.toString(token.getLine()), token.getSemantics().toString()));
+                }
+            }
```
tsaglam, line 254: We could probably move the loop into the if condition in order to avoid iterating whenever debug is not enabled.

tsaglam (approved):
Looks good, just one minor comment.

core/src/main/java/de/jplag/Submission.java
mbrdl: Good catch 👍

tsaglam (approved):

Issue #1082, opened 2023-05-15
CSV mode for scientific applications


tsaglam:
For scientific scenarios, a CSV mode would be great. When this mode is run, a CSV file is generated for each run that contains all matches in the following format: `submissionName1, SubmissionName2, avgSimilarity, maxSimilarity`. This would be in addition to the ZIP report.
Ideally, we would have a second version of this mode, where instead of submission names, the CSV only contains integer IDs, and an additional CSV is generated where each ID is mapped to the submission name. This second variant would allow giving data to other people without leaking sensitive data like names or the submission's code.

tsaglam:
Closed by #1385.

PR #1385, opened 2023-11-18
Implemented csv export.

TwoOfTwelve:
Implemented csv export. Includes:

1. An abstract csv printer that can be reused
2. Code to output comparisons in a csv and optionally anonymize the data
3. Cli code to activate the csv output.

Solves #1082

### DIFF
--- a/cli/src/main/java/de/jplag/cli/CLI.java
+++ b/cli/src/main/java/de/jplag/cli/CLI.java
@@ -81,7 +81,9 @@ public final class CLI {
                 ReportObjectFactory reportObjectFactory = new ReportObjectFactory();
                 reportObjectFactory.createAndSaveReport(result, cli.getResultFolder());
 
-                cli.printCsv(result);
+                if (cli.options.csv.print) {
+                    cli.printCsv(result);
+                }
             }
         } catch (ExitException exception) {
             logger.error(exception.getMessage()); // do not pass exception here to keep log clean
@@ -91,13 +93,10 @@ public final class CLI {
     }
 
     private void printCsv(JPlagResult result) {
-        if (options.csv.print) {
-            try {
-                CsvComparisonOutput.writeCsvResults(result.getAllComparisons(), options.csv.anonymize, new File(getResultFolder()),
-                        options.csv.fileName);
-            } catch (IOException e) {
-                logger.error("Could not write csv", e);
-            }
+        try {
+            CsvComparisonOutput.writeCsvResults(result.getAllComparisons(), options.csv.anonymize, new File(getResultFolder()), options.csv.fileName);
+        } catch (IOException e) {
+            logger.error("Could not write csv", e);
         }
     }
 
--- a/cli/src/main/java/de/jplag/cli/CLI.java
+++ b/cli/src/main/java/de/jplag/cli/CLI.java
@@ -81,19 +81,23 @@ public final class CLI {
                 ReportObjectFactory reportObjectFactory = new ReportObjectFactory();
                 reportObjectFactory.createAndSaveReport(result, cli.getResultFolder());
 
-                if (cli.options.csv.print) {
-                    try {
-                        CsvComparisonOutput.writeCsvResults(result.getAllComparisons(), cli.options.csv.anonymize, new File(cli.getResultFolder()),
-                                cli.options.csv.fileName);
-                    } catch (IOException e) {
-                        logger.error("Could not write csv", e);
-                    }
-                }
+                cli.printCsv(result);
             }
         } catch (ExitException exception) {
             logger.error(exception.getMessage()); // do not pass exception here to keep log clean
             finalizeLogger();
             System.exit(1);
+        }
+    }
+
+    private void printCsv(JPlagResult result) {
+        if (options.csv.print) {
+            try {
+                CsvComparisonOutput.writeCsvResults(result.getAllComparisons(), options.csv.anonymize, new File(getResultFolder()),
+                        options.csv.fileName);
+            } catch (IOException e) {
+                logger.error("Could not write csv", e);
+            }
         }
     }
 
--- a/cli/src/main/java/de/jplag/cli/CLI.java
+++ b/cli/src/main/java/de/jplag/cli/CLI.java
@@ -23,12 +23,12 @@ import de.jplag.Language;
 import de.jplag.cli.logger.CollectedLoggerFactory;
 import de.jplag.clustering.ClusteringOptions;
 import de.jplag.clustering.Preprocessing;
+import de.jplag.csv.comparisons.CsvComparisonOutput;
 import de.jplag.exceptions.ExitException;
 import de.jplag.merging.MergingOptions;
 import de.jplag.options.JPlagOptions;
 import de.jplag.options.LanguageOption;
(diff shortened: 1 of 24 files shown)


cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -78,6 +80,8 @@ public static void main(String[] args) {
                 JPlagResult result = JPlag.run(options);
                 ReportObjectFactory reportObjectFactory = new ReportObjectFactory();
                 reportObjectFactory.createAndSaveReport(result, cli.getResultFolder());
+
+                cli.printCsv(result);
```
Kr0nox, line 84: I feel like the code would be a bit easier to read, if the if from l. 93 was around this method and not inside it.
But if this is the way to keep it consistent in JPlag, keep it like it is

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -78,6 +80,8 @@ public static void main(String[] args) {
                 JPlagResult result = JPlag.run(options);
                 ReportObjectFactory reportObjectFactory = new ReportObjectFactory();
                 reportObjectFactory.createAndSaveReport(result, cli.getResultFolder());
```
Kr0nox, line 82: Do we still want to generate the zip report in CSV mode?

core/src/main/java/de/jplag/reporting/csv/comparisons/CsvComparisonOutput.java
```diff
@@ -0,0 +1,61 @@
...
+            String secondName = mapper.map(comparison.secondSubmission().getName());
+            printer.addRow(new CsvComparisonData(firstName, secondName, average, max));
+        }
+
+        printer.printToFile(new File(directory, fileName + ".csv"));
+
+        if (anonymize) {
+            List<Map.Entry<String, String>> nameMap = mapper.getNameMap();
+            CsvDataMapper<Map.Entry<String, String>> namesMapMapper = new HardcodedCsvDataMapper<>(2, it -> new String[] {it.getKey(), it.getValue()},
+                    new String[] {"realName", "id"});
```
Kr0nox, line 55: Would it be wise switching these cells around?
I feel like it is more relalistic that somebody is searching for the real name to a random ID than the other way around

core/src/main/java/de/jplag/reporting/csv/comparisons/NameMapper.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * @return The list of mappings.
+     */
+    List<Map.Entry<String, String>> getNameMap();
+
+    /**
+     * Simple implementation, that does not change the names.
+     */
+    class DirectMapper implements NameMapper {
```
Kr0nox, line 26: Maybe name this `IdentityMapper`?

core/src/main/java/de/jplag/reporting/csv/comparisons/NameMapperRandomIds.java
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * New instance
+     */
+    public NameMapperRandomIds() {
+        this.map = new HashMap<>();
+        this.random = new SecureRandom();
+    }
+
+    private String newId() {
+        String id = String.valueOf(this.random.nextInt(0, Integer.MAX_VALUE));
```
Kr0nox, line 25: Do we actually want to generate numbers of that big o a range?
This might make the csv hard to read

Kr0nox (changes requested):
Some small thinks I found

cli/src/main/java/de/jplag/cli/CliOptions.java
```diff
@@ -124,6 +127,18 @@ public static class Merging {
...
+        @Option(names = {"--csv-print"}, description = "If true, the comparisons will pre printed in a csv%n")
+        public boolean print;
+
+        @Option(names = {
+                "--csv-anonymize"}, description = "If true, the csv will contain anonymized data and a second csv will contain the actual names.%n")
+        public boolean anonymize;
+
+        @Option(names = {"--csv-file-name"}, description = "Overrides the base name of the csv file. Do not include .csv in here.%n")
+        public String fileName = "resultCsv";
+    }
```
tsaglam, line 140: Is this the path or the name? Only the name, right? Since the user cannot change the path, we could also think about a more simple solution in the CLI. For example, only one CSV flag. if this is enabled, both the anonymized and the unanonymized versions are written into two files. This flag can be part of the advanced group. This would keep the list of CLI options more concise and more readable.

core/pom.xml
```diff
@@ -37,6 +37,12 @@
             <version>${revision}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.scala-lang</groupId>
+            <artifactId>scala-library</artifactId>
+            <version>2.13.11</version>
+            <scope>test</scope>
+        </dependency>
```
tsaglam, line 45: Don't forget to remove this.

core/src/main/java/de/jplag/csv/CsvPrinter.java
```diff
@@ -0,0 +1,144 @@
...
+     * @param file The file to write
+     * @throws IOException on io errors
+     */
+    public void printToFile(File file) throws IOException {
+        try (Writer writer = FileUtils.openFileWriter(file)) {
+            this.printCsv(writer);
+        }
+    }
+
+    public String printToString() throws IOException {
```
tsaglam, line 88: Is there a reason this is not the toString() method?

core/src/main/java/de/jplag/csv/comparisons/NameMapperIncrementalIds.java
```diff
@@ -0,0 +1,42 @@
+package de.jplag.csv.comparisons;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Maps the real names of submissions to random ids to anonymize the data.
+ */
+public class NameMapperIncrementalIds implements NameMapper {
```
tsaglam, line 10: Are the duplicate name mappers still relics of the state before our discussion?
Right now, we just want simple incremental IDs in the order of the toplist.

tsaglam (changes requested):
I found a few issues.

cli/src/main/java/de/jplag/cli/OutputFileGenerator.java
```diff
@@ -0,0 +1,36 @@
...
+import java.io.File;
+import java.io.IOException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import de.jplag.JPlagResult;
+import de.jplag.csv.comparisons.CsvComparisonOutput;
+
+public class OutputFileGenerator {
```
tsaglam, line 12: Utility class should be final.

tsaglam (commented):
Nearly ready to merge, just one thing we overlooked.

tsaglam (approved):

tsaglam:
@Kr0nox ready to merge from your side?

Kr0nox (approved):

PR #1396, opened 2023-11-28
More similarity metrics

TwoOfTwelve:
Adds the similarity metrics described in #1134 
Also changes the report generation to export all metrics defined in the enum

As discussed the distributions in the report are not reversed anymore, this is an api breaking change.


core/src/main/java/de/jplag/options/SimilarityMetric.java
```diff
@@ -3,12 +3,19 @@
...
 public enum SimilarityMetric implements ToDoubleFunction<JPlagComparison> {
     AVG("average similarity", JPlagComparison::similarity),
     MIN("minimum similarity", JPlagComparison::minimalSimilarity),
     MAX("maximal similarity", JPlagComparison::maximalSimilarity),
-    INTERSECTION("matched tokens", it -> (double) it.getNumberOfMatchedTokens());
+    INTERSECTION("matched tokens", it -> (double) it.getNumberOfMatchedTokens()),
+    SYMMETRIC(
+            "symmetric similarity",
+            it -> 2.0 * it.getNumberOfMatchedTokens() / (it.firstSubmission().getNumberOfTokens() + it.secondSubmission().getNumberOfTokens())),
+    OVERLAP("overlap between both submissions (number of matched tokens)", JPlagComparison::getNumberOfMatchedTokens),
```
Kr0nox, line 16: Where is the difference between INTERSECTION and OVERLAP

core/src/main/java/de/jplag/options/SimilarityMetric.java
```diff
@@ -1,14 +1,27 @@
...
 import de.jplag.JPlagComparison;
+import de.jplag.Match;
 
 public enum SimilarityMetric implements ToDoubleFunction<JPlagComparison> {
     AVG("average similarity", JPlagComparison::similarity),
     MIN("minimum similarity", JPlagComparison::minimalSimilarity),
     MAX("maximal similarity", JPlagComparison::maximalSimilarity),
-    INTERSECTION("matched tokens", it -> (double) it.getNumberOfMatchedTokens());
+    INTERSECTION("matched tokens", it -> (double) it.getNumberOfMatchedTokens()),
+    SYMMETRIC("symmetric similarity", it -> {
```
SimDing, line 15: Isn't this equivalent to the implementation of the "average similarity", only ignoring base code?

Kr0nox:
With reversing the distribution, it now is the wrong way for the report-viewer. We should decide whether we want it to be reversed in the application and then written into the report the way the report-viewer currently expects it, or whether the report-viewer should turn it around

Kr0nox:
Replaced by #2058

PR #1408, opened 2023-11-30
Add universal character names in identifiers to cpp parser

nheir:
Starting from c99, C allows universal character names in identifiers see [6.4.2.1][1] and [Appendix D.1][2] for universal character name ranges

This is a valid code

```c
int main(void) {
	  int λ = 2;
	  int HÔNŒS = \u03bb - 12;
	  return H\u00d4NŒS;
}
```

[1]: https://www.iso-9899.info/n1570.html#6.4.3
[2]: https://www.iso-9899.info/n1570.html#D.

### DIFF
--- a/languages/cpp/src/main/javacc/CPP.jj
+++ b/languages/cpp/src/main/javacc/CPP.jj
@@ -9,6 +9,8 @@ options
   DEBUG_PARSER = false;
   DEBUG_LOOKAHEAD = false;
   DEBUG_TOKEN_MANAGER = false;
+  UNICODE_INPUT = true;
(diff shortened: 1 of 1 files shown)


tsaglam:
@nheir we are currently in the process of moving to the `cpp2` language module, which should already support special characters. We can still merge this PR, but in the future, the `cpp` language module will most likely be removed.
Is there a specific reason/feature why you are using the legacy `cpp` language module?

nheir:
I got errors while using cpp2 on C file
a quick test/analysis suggests that cpp2 assumes the code is C++ and that `new` is a keyword while it's not in C

tsaglam (approved):

PR #1409, opened 2023-11-30
Printable report viewer

### LINKED ISSUE #754: Feature request: option for printing/(option for additional) printable view
Unfortunately, in our school we have to print out (pairs of) code submissions that are suspected of being plagiarism, annotate them accordingly (by hand) and send them to the examination board as a "proof" of plagarism. This is a legal thing. With "printing out", I also mean generating a pdf or similar. With JPlag 2.x it was perfectly possible to print each of the two html files (each with the whole code)  belonging to a single plagiarism pair.
So the feature request for the report viewer would be to add, for example, an additional view for that.  Maybe changing the main view accordingly would destroy online usability or clarity, but I am not sure about that. Clarity was imho also pretty good in JPlag 2.x .

Thanks for developing JPlag!

Chris

Kr0nox:
This PR makes the report viewer printable. (closes #754)

The overview and cluster view print without the table, due to its items being loaded dynamically by a library.

In the comparison view the the list at the top gets replaced by a table containing the line numbers


sebinside (changes requested):
Thank you for making this primarily a CSS solution. I played with this PR using Edge and Chrome together with different printers and PDF export, and I noticed quite some whitespace and also cut off code parts (see the attached screenshot). While the former harms the planet, the latter could also make the solution unusable, especially regarding legal stuff. Please have another look at the solution (and also resolve the merge conflicts). Other than this, the solution is good, thank you!

image

Kr0nox:
I could not reproduce the cutoff with any browser. What browser and version did you experience this with?

Regarding the whitespace: Most of it is either spaced by the browser, when you have specific amounts of matches or because the otherwise diagramms wont be cut off

sebinside:
I had this problem in the following browsers:
* Edge Version 120.0.2210.133
* Chrome Version 120.0.6099.217
* Firefox 121.0.1 (here, even more things broke in the print view)

Let's discuss this in our next meeting.

sebinside (approved):

PR #1413, opened 2023-12-05
Sorting and filtering on cluster table

Kr0nox:
This PR adds the ability to search and filter the table in the cluster view.

For this code was moved around.
- The searchbar and sorting options were moved into the their own component.
  - This component is used in the comparison table component
- The filtering and sorting methods were moved from the overview to the comparison table component


tsaglam (approved):
As far as I can see, this looks good.

Issue #717, opened 2022-09-30
`Metric` data object stores UI description


JanWittler:
The [`Metric`](https://github.com/jplag/JPlag/blob/master/core/src/main/java/de/jplag/reporting/reportobject/model/Metric.java) record stores the description that should be shown in the report viewer. This breaks the intended separation between data and UI. As such, we should refactor `Metric` to only contain an identifier of the type of metric (e.g. modeled as an enum) and load the according description in the report viewer based on that identifier.

After this refactoring, we should move the [`Messages`](https://github.com/jplag/JPlag/blob/master/core/src/main/java/de/jplag/Messages.java) class and [all its resources](https://github.com/jplag/JPlag/tree/master/core/src/main/resources/de/jplag) to the CLI package and change the visibility to `protected`.

tsaglam:
Closed by #1415.

PR #1415, opened 2023-12-05
Deleted Metric record

TwoOfTwelve:
Deleted the Metric record, as it was not used anymore. #717 can be closed with this PR.


Kr0nox:
The issue also mentions moving the messages class
> After this refactoring, we should move the [Messages](https://github.com/jplag/JPlag/blob/master/core/src/main/java/de/jplag/Messages.java?rgh-link-date=2022-09-30T14%3A07%3A06Z) class and [all its resources](https://github.com/jplag/JPlag/tree/master/core/src/main/resources/de/jplag?rgh-link-date=2022-09-30T14%3A07%3A06Z) to the CLI package and change the visibility to protected.

Does this still apply?

tsaglam (approved):

PR #1416, opened 2023-12-05
ImmediatlyWriteZip

TwoOfTwelve:
Changed the report object factory to immediately write to a zip file instead of creating single files.

I tried to check how the runtime changed, by running jplag over 100 copies of the jplag code.
This is the result:

old:
8:10.19
7224.86s user 29.80s system

new:
7:46.55
6893.25s user 28.17s system

The majority of the time however was spent on comparing the submissions, so this might not be representative.

I build an artificial "worst-case" scenario to check the speed and both the old and the new version took the same time.


tsaglam:
As discussed, let's test how the performance is changed for a larger data set and include it in the PR description.

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -76,10 +77,10 @@ public static void main(String[] args) {
             if (!parseResult.isUsageHelpRequested() && !(parseResult.subcommand() != null && parseResult.subcommand().isUsageHelpRequested())) {
                 JPlagOptions options = cli.buildOptionsFromArguments(parseResult);
                 JPlagResult result = JPlag.run(options);
-                ReportObjectFactory reportObjectFactory = new ReportObjectFactory();
-                reportObjectFactory.createAndSaveReport(result, cli.getResultFolder());
+                ReportObjectFactory reportObjectFactory = new ReportObjectFactory(new File(cli.getResultFolder() + ".zip"));
```
tsaglam, line 80: Why is ".zip" hardcoded here? I think this should be part of the factory and probably a constant.

core/src/main/java/de/jplag/reporting/jsonfactory/ComparisonReportWriter.java
```diff
@@ -16,52 +16,51 @@
...
 public class ComparisonReportWriter {
 
-    private final JsonWriter fileWriter;
+    private final JPlagResultWriter resultWriter;
     private final Function<Submission, String> submissionToIdFunction;
     private final Map<String, Map<String, String>> submissionIdToComparisonFileName = new ConcurrentHashMap<>();
     private final Map<String, AtomicInteger> fileNameCollisions = new ConcurrentHashMap<>();
 
-    public ComparisonReportWriter(Function<Submission, String> submissionToIdFunction, JsonWriter fileWriter) {
+    public ComparisonReportWriter(Function<Submission, String> submissionToIdFunction, JPlagResultWriter jPlagResultWriter) {
```
tsaglam, line 32: better name it resultWriter

core/src/test/java/de/jplag/reporting/reportobject/ReportObjectFactoryTest.java
```diff
@@ -28,15 +26,30 @@ void testVersionLoading() {
...
+    }
+
+    /**
+     * Checks if the given file is a valid archive
+     * @param file The file to check
+     * @return True, if file is an archive
+     */
+    private static boolean isArchive(File file) {
+        int fileSignature = 0;
+        try (RandomAccessFile raf = new RandomAccessFile(file, "r")) {
```
tsaglam, line 47: Name it `file` or similar to avoid abbreviations.

core/src/test/java/de/jplag/reporting/reportobject/ReportObjectFactoryTest.java
```diff
@@ -28,15 +26,30 @@ void testVersionLoading() {
...
+            fileSignature = raf.readInt();
+        } catch (IOException e) {
+            // handle if you like
```
tsaglam, line 50: Shouldn't this be a test failure?

core/src/test/java/de/jplag/special/ReadmeCodeExampleTest.java
```diff
@@ -35,10 +36,12 @@ void testReadmeCodeExample() {
             JPlagResult result = JPlag.run(options);
 
             // Optional
-            ReportObjectFactory reportObjectFactory = new ReportObjectFactory();
-            reportObjectFactory.createAndSaveReport(result, "/path/to/output");
+            ReportObjectFactory reportObjectFactory = new ReportObjectFactory(new File("/path/to/output"));
```
tsaglam, line 39: This test is meant to remind us that we need to adapt the readme if we change the basic usage of JPlag!

tsaglam (changes requested):

Kr0nox:
What is the wanted behaviour, if jplag cant write the zip (can for example be forced by having a zip with the same name open).
Previously then you would still have the folder with the files. As I understand it, this will not happen now and the results are lost?

core/src/main/java/de/jplag/reporting/jsonfactory/ComparisonReportWriter.java
```diff
@@ -16,52 +16,51 @@
...
         List<JPlagComparison> comparisons = jPlagResult.getComparisons(numberOfComparisons);
-        writeComparisons(path, comparisons);
+        writeComparisons(comparisons);
         return submissionIdToComparisonFileName;
     }
 
-    private void writeComparisons(String path, List<JPlagComparison> comparisons) {
-        comparisons.parallelStream().forEach(comparison -> {
+    private void writeComparisons(List<JPlagComparison> comparisons) {
+        for (JPlagComparison comparison : comparisons) {
```
Kr0nox, line 54: I would suggest also performing a test with a lot of very small submissions (1000 maybe?, but each only a few small files), to see whether removing the parralelism actually decreased performance here

tsaglam (approved):

tsaglam:
@TwoOfTwelve we have one conflict here.

tsaglam (changes requested):
@TwoOfTwelve can you resolve the conflicts so I can merge?

tsaglam (approved):

PR #1433, opened 2023-12-12
Implemented progress bars for the cli

TwoOfTwelve:
Simplifies the logging output and add progress bars.

recording

The progress bars are implemented the following way:

The core module contains the infrastructure for progress bars. This includes interfaces for creating and manipulating progress bars (logging/ProgressBar and logging/ProgressBarProvider). It also contains some static code (logging/ProgressBarLogger), that allows the core to create new progress bars and the ui to set a new ProgressBarProvider in order to decide how the progress bars are implemented. The core module also has a dummy implementation for the interfaces, that simply writes to the log.

The CLI has been added implementations for the interfaces, that internally call a progress bar library (https://github.com/ctongfei/progressbar).

This way of implementing it, might behave strangely if JPlag is called multiple times simultaneously from the same JVM, but as this feature is currently only meant for the cli, I think the risk of that happening is negligible.


tsaglam:
@TwoOfTwelve can you add an example of the new log to the PR description?

core/src/main/java/de/jplag/UiHooks.java
```diff
@@ -0,0 +1,49 @@
+package de.jplag;
+
+/**
+ * Notifies the ui of state changes in JPlag
+ */
+public interface UiHooks {
```
tsaglam, line 6: Probably better named `ProgressListener`.

core/src/main/java/de/jplag/UiHooks.java
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * A null ui hook, that does nothing
+     */
+    UiHooks NullUiHooks = new UiHooks() {
```
tsaglam, line 10: Implicitly static final, so must be upper case with underscores.

cli/src/main/java/de/jplag/cli/CliUiHooks.java
```diff
@@ -0,0 +1,44 @@
...
+import de.jplag.UiHooks;
+
+import me.tongfei.progressbar.ProgressBar;
+import me.tongfei.progressbar.ProgressBarBuilder;
+import me.tongfei.progressbar.ProgressBarStyle;
+
+/**
+ * Provides progress bars for the cli
+ */
+public class CliUiHooks implements UiHooks {
```
tsaglam, line 12: Probably better named `ProgressBarListener`.

core/src/main/java/de/jplag/UiHooks.java
```diff
@@ -0,0 +1,49 @@
...
+    void multiStepDone();
+
+    /**
+     * The available processes
+     */
+    enum ProgressBarType {
+        LOADING,
+        PARSING,
+        COMPARING
+    }
```
tsaglam, line 48: Better put this into its own file.

core/src/main/java/de/jplag/JPlag.java
```diff
@@ -35,43 +37,49 @@ private static Version loadVersion() {
 
     /**
      * Creates and initializes a JPlag instance, parameterized by a set of options.
-     * @deprecated in favor of static {@link #run(JPlagOptions)}.
      * @param options determines the parameterization.
      */
-    @Deprecated(since = "4.3.0")
     public JPlag(JPlagOptions options) {
```
tsaglam, line 42: Should stay deprecated to tie the object creation to the run call. Just introduce a new run method that overloads the existing one.

core/src/main/java/de/jplag/JPlag.java
```diff
@@ -35,43 +37,49 @@ private static Version loadVersion() {
...
+        this(options, UiHooks.NullUiHooks);
+    }
+
+    /**
+     * Creates and initializes a JPlag instance, parameterized by a set of options.
+     * @param options determines the parameterization.
+     * @param uiHooks Used to notify the ui of state changes
+     */
+    public JPlag(JPlagOptions options, UiHooks uiHooks) {
```
tsaglam, line 51: As this constructor is new, it can be private.

core/src/main/java/de/jplag/JPlag.java
```diff
@@ -35,43 +37,49 @@ private static Version loadVersion() {
...
     }
 
     /**
      * Main procedure, executes the comparison of source code submissions.
-     * @param options determines the parameterization.
      * @return the results of the comparison, specifically the submissions whose similarity exceeds a set threshold.
      * @throws ExitException if JPlag exits preemptively.
      */
-    public static JPlagResult run(JPlagOptions options) throws ExitException {
+    public JPlagResult run() throws ExitException {
```
tsaglam, line 71: Should stay static, this would be a major API-breaking change.

tsaglam (commented):
Had a discussion with @dfuchss, here is some early feedback.

tsaglam:
@TwoOfTwelve we still have some comments from the last review here.

core/src/main/java/de/jplag/strategy/ParallelComparisonStrategy.java
```diff
@@ -28,8 +31,13 @@ public JPlagResult compareSubmissions(SubmissionSet submissionSet) {
         }
 
         List<SubmissionTuple> tuples = buildComparisonTuples(submissionSet.getSubmissions());
-        List<JPlagComparison> comparisons = tuples.stream().parallel().map(tuple -> compareSubmissions(tuple.left(), tuple.right()))
-                .flatMap(Optional::stream).toList();
+        ProgressBar progressBar = ProgressBarLogger.createProgressBar(ProgressBarType.COMPARING, tuples.size());
+        List<JPlagComparison> comparisons = tuples.stream().parallel().map(tuple -> {
+            Optional<JPlagComparison> result = compareSubmissions(tuple.left(), tuple.right());
+            progressBar.step();
```
tsaglam, line 37: Is there a reason this is done here and not in the compareSubmissions method? This ties the logging to the parallel strategy and makes the code a bit more cluttered.

tsaglam (changes requested):
The new classes are lacking documentation, thus making it harder to review this PR.
Please provide JavaDoc comments. Also: The design rationale of the interfaces in the core and the implementations in the CLI, as well as the separation between the provider and the progress bar, could be discussed in the PR description. Ideally, PR descriptions not only explain the feature and its importance but also how it was realized.

core/src/main/java/de/jplag/strategy/ParallelComparisonStrategy.java
tsaglam: As the progress bar behaves like a logger, I would create it in `AbstractComparisonStrategy` and store it as a (private) field. I would then call `progressBar.step()`  after the GST in  `AbstractComparisonStrategy.compareSubmissions()`. The advantage would be, that the concrete strategy never needs to bother with the progress bar. You can even get rid of the the calculation if you create the progress bar in the `buildComparisonTuples()` method.

tsaglam (changes requested):
Just a few minor issues. The solution regarding logging and the parallel strategy looks good!

olmokramer:
@TwoOfTwelve I'm sorry for the long delay, I've finally got some time to take a look at this PR. It looks like this will suit our needs of getting the progress. We would need to implement our own progress bar provider because we're running JPlag on some server and need to have the results show up in a web interface, but given the simplicity of the `ProgressBar` interface that should be no problem. Thanks for following up on this!

cli/src/main/java/de/jplag/cli/logger/TongfeiProgressBarProvider.java
```diff
@@ -0,0 +1,28 @@
...
+        me.tongfei.progressbar.ProgressBar progressBar = new ProgressBarBuilder().setTaskName(getProgressBarName(type)).setInitialMax(totalSteps)
+                .setStyle(ProgressBarStyle.UNICODE_BLOCK).build();
+        return new TongfeiProgressBar(progressBar);
+    }
+
+    private String getProgressBarName(ProgressBarType progressBarType) {
+        return switch (progressBarType) {
+            case LOADING -> "Loading Submissions  ";
+            case PARSING -> "Parsing Submissions  ";
+            case COMPARING -> "Comparing Submissions";
```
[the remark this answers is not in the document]
tsaglam: I prefer UI in core to the duplication, but I understand your design choice here.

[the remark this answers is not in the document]
dfuchss: I think it can stay in the CLI :)

languages/cpp2/src/main/java/de/jplag/cpp2/CPPListener.java
```diff
@@ -73,6 +55,35 @@ class CPPListener extends AbstractAntlrListener {
                 .onEnter((rule, varReg) -> varReg.setNextVariableAccessType(VariableAccessType.WRITE));
         visit(BracedInitListContext.class).map(BRACED_INIT_BEGIN, BRACED_INIT_END).withSemantics(CodeSemantics::new);
 
+        typeSpecifierRule();
```
Kr0nox, line 58: @TwoOfTwelve Did you mean to push these changes to this PR?

tsaglam (commented):
Looks good! Could you take a look at the sonar issues?

PR #1444, opened 2023-12-13
add zipfilename to title and overview

Kr0nox:
The name of the uploaded zip gets displayed:
- In the page title
- On the overview view
- On the information view


sebinside (approved):
I only made a small adjustment to better support smaller screens (or many open tabs :)).

PR #1446, opened 2023-12-14
Display all cli options on information view

Kr0nox:
I have added a new file called `options.json` to the zip. This file contains all values from the `JPlagOptions` object.
In the report viewer all run options are now accesible on the information view (reached by clicking more on the overview).

The report viewer stays backwards compatible with older versions of JPlag.

We should discuss how we want to display empty options (for example if no basecode was given)

@tsaglam Could you have a look at the order of the options and tell me where you would like to have it changed?


core/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -224,6 +229,24 @@ private void writeSubmissionIndexFile(JPlagResult result, String path) {
         jsonFileWriter.writeFile(fileIndex, path, SUBMISSION_FILE_INDEX_FILE_NAME);
     }
 
+    private void writeOptionsFiles(JPlagOptions options, String path) {
```
dfuchss, line 232: This could lead to errors in the future, if we forget to add new options. Do you have an idea to retrieve that from the options automatically?

dfuchss: Nevertheless, I like the idea of persisting this information :)

Kr0nox: We could add the anotations (`@JsonProp`) directly to the options and then write custom Serializers for all non native supported types (like the language) (see [https://www.baeldung.com/jackson-annotations#7-jsonserialize](https://www.baeldung.com/jackson-annotations#7-jsonserialize))

dfuchss: I think that would be a good idea :)

core/src/main/java/de/jplag/reporting/jsonfactory/serializer/LanguageSerializer.java
```diff
@@ -0,0 +1,28 @@
...
+public class LanguageSerializer extends StdSerializer<Language> {
+
+    /**
+     * Constructor used by the fasterxml.jackson
+     */
+    public LanguageSerializer() {
+        this(null);
+    }
+
+    public LanguageSerializer(Class<Language> t) {
```
tsaglam, line 20: Use a better param name here.

core/src/main/java/de/jplag/reporting/jsonfactory/serializer/LanguageSerializer.java
```diff
@@ -0,0 +1,28 @@
...
+        super(t);
+    }
+
+    @Override
+    public void serialize(Language value, JsonGenerator gen, SerializerProvider provider) throws IOException {
```
tsaglam, line 25: also here, avoid abbreviations.

tsaglam (changes requested):
Just two minor things. Regarding the order, is it as depicted in CliOptions.ts or do I have to check it in the viewer?

sebinside (approved):

Kr0nox:
@tsaglam For the order you can either start the report viewer and give it a result generated from this branch or look in InformationView.vue file starting from line 26

tsaglam (approved):
The order is fine for me. Currently, I have no change requests. Also, we can update it in the future.

PR #1448, opened 2023-12-15
Report Viewer Demo

Kr0nox:
This PR adds a demo for the report viewer. This demo will be hosted in a new repository (called demo) and will then be accessible on jplag.github.io/Demo/
The demo gets updated on each push to main. 
We build jplag from its sources and then run it on the progpedia dataset. The results are inserted into the report viewer

To preview the demo locally, you have to add a report called `example.zip` to the public folder in the report-viewer and then run:
`npx vite --mode demo`

@tsaglam Before the newly created workflow gets run (so either before the merge of develop into main or before it is run manually) please create an new repo called `Demo` and create the `gh-pages` branch.

@dfuchss Do you know whether we can reuse the token for the dev deploy in the workflow?


dfuchss:
You should be able to reuse the token

Kr0nox:
Demo can be viewed [here](https://kr0nox.github.io/JPlag/)

dfuchss (approved):

sebinside (approved):
I like the solution, let's see whether it will eventually break the build 😁 I just created an empty [Demo](https://github.com/jplag/Demo) repository.

For a future version, we should discuss whether the report viewer should directly open up the overview page in demo mode because the file upload view is a little bit pointless.

dfuchss:
@Kr0nox should be fixed now. I've retriggered the Workflow :)

dfuchss:
See https://jplag.github.io/Demo/

Issue #1028, opened 2023-04-13
Language-specific option for token string normalization


brodmo:
There is currently an unused `normalizeSubmissions` method in `SubmissionSet`. Instead, the `SubmissionSet` constructor should check whether token string normalization is enabled through the `JPlagOptions` object it has and perform the normalization if it is enabled. The token string normalization option would need to be language-specific as it is not supported for all languages.

tsaglam:
Relates to #920.

tsaglam:
Closed, as normalization can be controlled via `--normalize`.

PR #1479, opened 2024-01-13
Feature/normalization option

TwoOfTwelve:
Implemented the cli (and JPlag) option to enable normalization.
If normalization is enabled, but not supported for the selected language JPlag immediately throws an exception.

The help text also shows, if normalization is supported for a language:
```
  cpp         supports normalization: false
  cpp2        supports normalization: true
  csharp      supports normalization: false
  emf         supports normalization: false
  emf-model   supports normalization: false
  go          supports normalization: false
  java        supports normalization: true
  kotlin      supports normalization: false
  llvmir      supports normalization: false
  python3     supports normalization: false
  rlang       supports normalization: false
  rust        supports normalization: false
  scala       supports normalization: false
  scheme      supports normalization: false
  scxml       supports normalization: false
  swift       supports normalization: false
  text        supports normalization: false
  typescript  supports normalization: false
```

(relates to  #1028)


Kr0nox:
Since the README and wiki contains the help text, this PR should probably updates those

core/src/main/java/de/jplag/JPlag.java
```diff
@@ -96,4 +101,10 @@ private static void logSkippedSubmissions(SubmissionSet submissionSet, JPlagOpti
             }
         }
     }
+
+    private static void checkForConfigurationConsistency(JPlagOptions options) throws ConfigurationException {
+        if (options.normalize() && !options.language().supportsNormalization()) {
+            throw new ConfigurationException(String.format("The language %s cannot be used with normalization.", options.language().getName()));
```
tsaglam, line 107: I thought about it again and would prefer logging the error here instead of terminating it.
The error message should state that it is not available for that language and thus running without it.

docs/1.-How-to-Use-JPlag.md
```diff
@@ -84,22 +84,24 @@ Clustering
...
+  llvmir      supports normalization: false
+  python3     supports normalization: false
+  rlang       supports normalization: false
+  rust        supports normalization: false
+  scala       supports normalization: false
+  scheme      supports normalization: false
+  scxml       supports normalization: false
+  swift       supports normalization: false
+  text        supports normalization: false
+  typescript  supports normalization: false
```
tsaglam, line 104: After adding the hint to the option itself, I would leave this out.

tsaglam (changes requested):
A few minor issues here.

tsaglam (approved):

tsaglam:
@TwoOfTwelve Do you want to make the changes regarding the language interface in this PR, or should I merge?

tsaglam:
@TwoOfTwelve 

> Basically, in langauge-api/standardOptions the StandardOptionsLanguage and the NormalizatbleLanguage classes contain a bunch of ugly code. Also there will be runtime exceptions if the interface is used without the abstract class. 

Agreed, that is not ideal from my point of view.

> I think it is worth it, because it is the most flexible solution and also gives us an infrastructure for more options shared between languages.

I do not think so, as it introduces a lot of complexity just to prematurely plan for future use cases.
I think I prefer the lightweight option we discussed in the last meeting where we just provide a version of the parse method that supplies the language module with the information if normalization was enabled or not.

```java
List<Token> parse(Set<File> files, boolean normalizationRequested) throws ParsingException;

default List<Token> parse(Set<File> files) throws ParsingException {
	parse(files, false);
}
```

This is minimally invasive and should not be API breaking through the default option.

tsaglam (commented):
Just the minor issue we discussed regarding the language interface. Coverage is fine in this PR.

tsaglam (approved):

tsaglam:
@TwoOfTwelve now you know why I was thinking about replacing as much code as possible in the Scala module with Java code! :D

Issue #1386, opened 2023-11-21
Enhancement: print the file name if parse failed


hnwyllmm:
I use JPlag for checking [miniob](https://github.com/oceanbase/miniob) which use C++ 20 not supported by JPlag. So there are some exceptions when parsing the code.

I have two suggestions.
1. Print the file name if except occurs;
2. We can continue the parse even if except occurs.

Below is the stack trace.
```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because "next.children" is null
	at de.jplag.cpp2.CPPTokenListener.getDescendant(CPPTokenListener.java:375)
	at de.jplag.cpp2.CPPTokenListener.enterSimpleDeclaration(CPPTokenListener.java:326)
	at de.jplag.cpp2.grammar.CPP14Parser$SimpleDeclarationContext.enterRule(CPP14Parser.java:5639)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.enterRule(ParseTreeWalker.java:50)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:33)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:36)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:36)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:36)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:36)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:36)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:36)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:36)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:36)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:36)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:36)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:36)
[...]

PR #1480, opened 2024-01-13
Changed AbstractAntlrParserAdapter to catch all exceptions during par…

TwoOfTwelve:
This makes it, so all antlr exceptions are caught, so that the name of the file, that failed is always shown.
Addresses: #1386

### DIFF
--- a/language-antlr-utils/src/main/java/de/jplag/antlr/AbstractAntlrParserAdapter.java
+++ b/language-antlr-utils/src/main/java/de/jplag/antlr/AbstractAntlrParserAdapter.java
@@ -68,7 +68,7 @@ public abstract class AbstractAntlrParserAdapter<T extends Parser> extends Abstr
             for (ParseTree child : entryContext.children) {
                 treeWalker.walk(listener, child);
             }
-        } catch (Throwable exception) { // catching throwable to capture any exceptions thrown by ANTLR.
+        } catch (Exception exception) { // catching generic exception to capture any exceptions thrown by ANTLR.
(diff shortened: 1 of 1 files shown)


language-antlr-utils/src/main/java/de/jplag/antlr/AbstractAntlrParserAdapter.java
```diff
@@ -69,7 +68,7 @@ private void parseFile(File file, TokenCollector collector) throws ParsingExcept
             for (ParseTree child : entryContext.children) {
                 treeWalker.walk(listener, child);
             }
-        } catch (IOException exception) {
+        } catch (Throwable exception) { // catching throwable to capture any exceptions thrown by ANTLR.
```
dfuchss, line 71: Isn't Exception enough ?

dfuchss: I also had this kind of Errors in mind. Nevertheless, I won't catch this kind of Error since this kind of errors can't be handled in a good way. Catching a OutOfMemoryError won't solve the problem as far as I remember :)

But I don't have a strong opinion on what to catch here :)

tsaglam: One is better for API users, the other for application users. Let us go with `Exception` for now, we can always change it.

dfuchss (approved):

PR #1492, opened 2024-01-18
Remove comparison single file upload

Kr0nox:
Currently JPlag supports uploading a single comparison file json and then navigates to the comparison view. This can not work since the files of the submission are missing. Because of this this PR removes it.

The single file upload for overview files is untouched


sebinside (approved):
Works fine, thank you. If easily possible, we could think about showing a clearer error message when only dropping an `overview.json`. As this feature will soon not be required anyway, it's also fine as is.

PR #1493, opened 2024-01-19
Added a better error message, if no valid language was supplied.

TwoOfTwelve:
I an invalid language was supplied on the cli the following message is now printed:

```
Language asdf does not exists. Available languages are: cpp, cpp2, csharp, emf, emf-model, go, java, kotlin, llvmir, python3, rlang, rust, scala, scheme, scxml, swift, text, typescript
```

### DIFF
--- a/cli/src/main/java/de/jplag/cli/CLI.java
+++ b/cli/src/main/java/de/jplag/cli/CLI.java
@@ -6,6 +6,7 @@ import static picocli.CommandLine.Model.UsageMessageSpec.SECTION_KEY_SYNOPSIS;
 
 import java.io.File;
 import java.security.SecureRandom;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Random;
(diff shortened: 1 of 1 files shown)


Kr0nox (approved):

Kr0nox:
With this we could also rethink the available languages message printed at the start
```
 LanguageLoader - Available languages: '[C/C++ Scanner [basic markup], C/C++ Parser, C# 6 Parser, EMF metamodel, EMF models (dynamically created token set), Go Parser, Javac based AST plugin, Kotlin Parser, LLVMIR Parser, Python3 Parser, R Parser, Rust Language Module, Scala parser, SchemeR4RS Parser [basic markup], SCXML (Statechart XML), Swift Parser, Text Parser (naive), Typescript Parser]'
```
Since these are the available parsers, not languages as stated in the message. and with #1494 we have two languages that use the same parser and that new language would not be explicitly mentioned in this message

tsaglam (approved):

tsaglam:
@Kr0nox the language loader should check for language interface implementations, and thus the Javascript module should be mentioned when the dependency in the CLI module ist set...

PR #1494, opened 2024-01-20
JavaScript language module

Kr0nox:
Adds a language module for javascript.
It just extends TypeScript, since every typescript keyword is also a javascript keyword.

I moved the js specific tests from the ts language module to the js language module


Kr0nox:
> @Kr0nox the language loader should check for language interface implementations, and thus the Javascript module should be mentioned when the dependency in the CLI module ist set...

The Language Loader currently uses the `getName` method of a Language. Currently this returns the Parser for every language. As you mentioned it should mention the javascript method, but the parser still is the typescript parser. How do you want that handled?

tsaglam:
I would say this is an internal detail that is not important for users. So it is fine when it says javascript for now...

Kr0nox:
But it gets mentioned when starting the CLI in this message
```
LanguageLoader - Available languages: '[C/C++ Scanner [basic markup], C/C++ Parser, C# 6 Parser, EMF metamodel, EMF models (dynamically created token set), Go Parser, Javac based AST plugin, Kotlin Parser, LLVMIR Parser, Python3 Parser, R Parser, Rust Language Module, Scala parser, SchemeR4RS Parser [basic markup], SCXML (Statechart XML), Swift Parser, Text Parser (naive), Typescript Parser]
 ```

**Comment by tsaglam** [comment]
I think I may have misunderstood something, let's discuss it tomorrow, that is probably easier. :D

File: languages/javascript/pom.xml [hunk 1]
```diff
@@ -0,0 +1,52 @@
...
+        <dependency>
+            <groupId>de.jplag</groupId>
+            <artifactId>language-antlr-utils</artifactId>
+            <version>${revision}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>de.jplag</groupId>
+            <artifactId>typescript</artifactId>
+            <version>4.4.0-SNAPSHOT</version>
```
tsaglam, line 26: Use `${revision}` here, as this allows us to only change one variable to increment the version number.

languages/javascript/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>de.jplag</groupId>
+            <artifactId>language-testutils</artifactId>
+            <version>4.4.0-SNAPSHOT</version>
```
tsaglam, line 32: Also here.

tsaglam (changes requested):
Just one minor issue.

tsaglam (approved):

Issue #1445, opened 2023-12-14
Issues with same root directory name


Kr0nox:
If people use the same root directory two problems can occur.
1) The root directories have the same name, but the submissions have different names
Folder structur
```
└ jplag.jar
└ files
|   └ A
└ test
|   └ files
|       └ C
```
executed with `jplag.jar files -new test/files`
Here is the generated report:
[results_A_C.zip](https://github.com/jplag/JPlag/files/13671844/results_A_C.zip)
In the viewer we are not able to distinguish from wich files folder each submission came from. we should discuss wether this is an issue.
<br /><br />

2) Same root dir name and submission with same name.
Folder structur
```
└ jplag.jar
└ files
|   └ A
|       └ CLI.java
└ test
|   └ files
|       └ A
|           └ CLIOptions.java
```
executed with `jplag.jar files -new test/files`
Here is the generated report:
[results_A_A.zip](https://github.com/jplag/JPlag/files/13671875/results_A_A.zip)
Again, we cant tell from which folder each submission came from. This is defineatly an issue since they have the same name.
In addition only the files from one of the submissions are present in the files folder in the zip. Due to this matches cant be shown. This again is an issue.

All test were done with a build from the develop branch.

Kr0nox:
In the meeting we sumarized that the first issue is "ok" and will not need to be fixed in the next major version.
The second issue is important tho and should be fixed in v5

tsaglam:
Partially addressed by #1495, thus _not_ closed.

euberdeveloper:
> If people use the same root directory two problems can occur.
> 
> 1. The root directories have the same name, but the submissions have different names
>    Folder structur
> 
> ```
> └ jplag.jar
> └ files
> |   └ A
> └ test
> |   └ files
> |       └ C
> ```
> 
> executed with `jplag.jar files -new test/files` Here is the generated report: [results_A_C.zip](https://github.com/jplag/JPlag/files/13671844/results_A_C.zip) In the viewer we are not able to distinguish from wich files folder each submission came from. we should discuss wether this is an issue.
> 
> 2. Same root dir name and submission with same name.
>    Folder structur
> 
> ```
> └ jplag.jar
> └ files
> |   └ A
> |       └ CLI.java
> └ test
> |   └ files
> |       └ A
> |           └ CLIOptions.java
> ```
> 
[...]

PR #1495, opened 2024-01-22
Check for duplicate root directory names

Kr0nox:
This PR fixes the problems caused by having two root directories with the same same as described in #1445 by throwing an exception when two last folders have the same name


core/src/main/java/de/jplag/JPlag.java
```diff
@@ -96,4 +101,27 @@ private static void logSkippedSubmissions(SubmissionSet submissionSet, JPlagOpti
             }
         }
     }
+
+    private static void checkForConfigurationConsistency(JPlagOptions options) throws ConfigurationException {
+        List<String> duplicateNames = getDuplicateSubmissionFolderNames(options);
+        if (duplicateNames.size() > 0) {
+            throw new ConfigurationException(String.format("Duplicate root directory names found: %s", String.join(", ", duplicateNames)));
```
tsaglam, line 108: We already have a `RootDirectoryException`, that would fit here as well.

tsaglam (changes requested):
Just one small thing...

core/src/main/java/de/jplag/JPlag.java
Kr0nox: My plan was too keep it in line with #1479.
But if you prefer that exception I can also implement it that way

tsaglam: For that PR, I want to move away from using an exception and just log and error, so thus `RootDirectoryException` fits better in your PR.

tsaglam (approved):

PR #1506, opened 2024-01-24
Make cluster graph edge clickable

Kr0nox:
When clicking an Edge in the cluster graph, it now opens the coresponsing comparison


sebinside (commented):
The changes are fine, thank you. Only one minor comment: Could you change the mouse coursor when hovering over an edge to indicate that the user can click?

sebinside (approved):

PR #1510, opened 2024-01-25
Dont show cluster  graphs on too many members

Kr0nox:
When a cluster has too many members graph drawing can become very slow and also visually very hard to distinguish, since for example the circles start intersecting.
Because of this cluster graphics only get shown up to a certain member treshhold and a warning is displayed


tsaglam:
The following image shows what caused us to create this PR... 😄 

image

sebinside (approved):

PR #1515, opened 2024-01-30
Link to repo in Report Viewer

Kr0nox:
Adds a link to the issues in the report viewer


sebinside (commented):
We should quickly discuss the placement and the message in our next meeting. I'm fine with the placement but would add the information that JPlag is --in fact-- open source software. Timur might want the display message to be more visible, let's see :)

tsaglam (approved):
As discussed, it looks good.

Issue #1109, opened 2023-05-24
[Feature Suggestion] Separate percentages for multiple documents


erhanbicerr:
Hello,
Firstly, I am highly appreciated for the JPlag tool, it excels in minimizing the workload. My suggestion is to be able to see similarity percentages for each document pair in a folder structure such as multiple files exist within a folder. So, the aim is to compare multiple files within each folder by observing their own percentages. Following is an example:
Folder 1
|----->a1.py
|----->a2.py
Folder 2
|------>a1.py
|------>a2.py
Similarities (Desired Output):
%98.2 in a1.py between Folder 1 and Folder 2
%43.4 in a2.py between Folder 1 and Folder 2
Currently, it shows the average percentage of whole files.
Thanks.

PR #1516, opened 2024-01-30
File-based Similarity in the Comparison View

Kr0nox:
Displays the percentage of tokens of a file that are part of a match.
Also displays the total number of tokens in each submission.

The submission file index now holds data about the token count of each file

relates to #1109


Kr0nox:
We should discuss how the EOF Token should be handled. Currently it is excluded in the percent calculation but not in the tooltip, which may be confusing for the user.
If we would include it in the percent calculation a file that is entirely copied could not achive a 100% match. So one option would be to not count it in both numbers

sebinside (commented):
Regarding the UI, I would argue for not displaying decimal numbers to minimize visual clutter. Also, I would argue for a dark gray instead of black text color.

Kr0nox:
The performance impact of this is pretty heavy.
Zipping was increased from 13 seconds to 61 seconds for 64 copies of JPlag (Comparison time was about 3m 40s)

Because of this the token calculation should probably be moved somewhere in the core where it is more efficient. We should discus where this is and how this should be done

tsaglam:
> Using a parallel stream over the submissions we could reduce this to under a second. This would reduce readibility tho in my opinion. What is your opinion on that?

I would say performance > readability here.

> We should discuss whether this is fine or whether it should be moved. We should factor in how difficult it would be to optimize that. since parsing of submissions is not done in parallel there would be no performance improvement when moving this into the submission class and iterating over the list of tokens returned form the `Language:parse` method. So the only option I see to do this, is to do this directly in the parser of each language. This would lead to some big changes to the Language and Parser interface in my opinion tho. @tsaglam What is your opinion?

Doing it in each language is not something I am in favor of.
Another option would be to do it when creating the result object. e.g. either add a method to JPlagResult that returns the number of tokens per file, which is pre-calculated in the constructor and put in a map. Or add a method for that information to the submission class, but only calculate the info lazily when the method is first called. Thus, when writing the report, you can iterate over the submissions in parallel and call that method. This last solution is something I would actually prefer.

core/src/main/java/de/jplag/Submission.java
```diff
@@ -309,4 +311,21 @@ public Submission copy() {
...
+            return Collections.emptyMap();
+        }
+        Map<File, Integer> tokenCounts = new HashMap<>();
+        for (File file : this.files) {
+            tokenCounts.put(file, 0);
+        }
+        for (Token token : this.tokenList) {
+            tokenCounts.put(token.getFile(), tokenCounts.get(token.getFile()) + 1);
+        }
+        return tokenCounts;
```
tsaglam, line 329: This is done every time this method is called. For API users, it would be better if the map is stored as field, and calculated upon the first call of the method. As in: If empty, calculate, if not, return.

tsaglam (commented):
Just one small thing.

tsaglam (approved):

sebinside (approved):

PR #1554, opened 2024-02-13
Default to System Dark mode

Kr0nox:
The report viewer now tries to fetch the systems/browsers setting whether websites should be display in light or dark mode


sebinside (approved):
Worked for me in Edge, Chrome, and Firefox

PR #1566, opened 2024-02-16
Renamed the result-directory option to result-file

TwoOfTwelve:
The -r option is now called --result-file instead of --result-directory. This is, because JPlag does not interpret this option as a directory. To fit this, the code handling it has been updated to reflect this. Missing `.zip` endings will now be automatically added/ommited when using `-r`.

### DIFF
--- a/languages/scala/src/test/java/de/jplag/scala/ScalaLanguageTest.java
+++ b/languages/scala/src/test/java/de/jplag/scala/ScalaLanguageTest.java
@@ -19,7 +19,6 @@ import org.junit.jupiter.api.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import de.jplag.ParsingException;
 import de.jplag.SharedTokenType;
 import de.jplag.Token;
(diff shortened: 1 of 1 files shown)


Kr0nox (approved):

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -250,7 +252,17 @@ private String generateDescription() {
...
+            return optionValue;
+        } else {
+            return optionValue + DEFAULT_FILE_ENDING;
+        }
+    }
+
+    private String getResultFileBaseName() {
+        String defaultOutputFile = getResultFilePath();
+        return defaultOutputFile.substring(0, defaultOutputFile.length() - DEFAULT_FILE_ENDING.length());
     }
```
tsaglam, line 267: Could you add this to the PR description, right now this is a sneaky change (somewhat at least, as it is mentioned in the CLI help text)...

tsaglam (approved):
Looks good, just one meta issue.

PR #1496, opened 2024-01-23
Renamed cpp to c and cpp2 to cpp

TwoOfTwelve:
Renamed cpp to c and cpp2 to cpp as discussed.

PR #1567, opened 2024-02-16
Fix: Old cpp language names not found

Kr0nox:
Due to #1496 cpp results from <5.0.0 could not be read. this fixes it


uuqjz:
Can you open results from both datasets that i send you and view the actual comparisons?

uuqjz (approved):

PR #1568, opened 2024-02-16
Fixed a bug in the creation of the result due to missing slashes.

TwoOfTwelve:
For source roots without sub-directories, the files in the results.zip had the wrong name, because there was a slash missing.

### DIFF
--- a/core/src/main/java/de/jplag/reporting/FilePathUtil.java
+++ b/core/src/main/java/de/jplag/reporting/FilePathUtil.java
@@ -7,6 +7,7 @@ import java.util.function.Function;
 import de.jplag.Submission;
 
 public final class FilePathUtil {
+    private static final String ZIP_PATH_SEPARATOR = "/"; // Paths in zip files are always separated by a slash
 
     private FilePathUtil() {
         // private constructor to prevent instantiation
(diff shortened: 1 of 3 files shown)


core/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -112,7 +112,7 @@ private void copySubmissionFilesToReport(JPlagResult result) {
                 if (relativeFilePath.isEmpty()) {
                     relativeFilePath = file.getName();
```
Kr0nox, line 113: We could also just preface the files name with a `/`. Would be shorter, simpler solution. 
What do we prefer?

uuqjz (commented):
Issues are solved

core/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
tsaglam: To be honest, I have no real preference. But related to pre-release panic, I say fewer changes are better! :D

tsaglam (commented):
Just one question.

core/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
tsaglam: No, I would say it is fine as it is right now.

core/src/main/java/de/jplag/reporting/FilePathUtil.java
```diff
@@ -26,4 +26,23 @@ public static String getRelativeSubmissionPath(File file, Submission submission,
...
+        while (rightStripped.startsWith("/")) {
+            rightStripped = rightStripped.substring(1);
+        }
+
+        String leftStripped = left;
+        while (leftStripped.endsWith("/")) {
+            leftStripped = leftStripped.substring(0, leftStripped.length() - 1);
+        }
+
+        return leftStripped + "/" + rightStripped;
```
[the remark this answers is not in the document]
tsaglam: Thanks for clarifying. In that case, let us use a constant `ZIP_PATH_SEPARATOR` in that class for it, this allows us to clarify the concept and even add a comment along the lines of `// must always be a forward slash`.

tsaglam (approved):

PR #1607, opened 2024-02-23
Feature/cli module refactoring

TwoOfTwelve:
Contains a big refactoring of the cli module. In specific:

- Minor refactoring all over the module based on Intellijs code analysis.
- A rewrite of the CollectedLogger class, because the base interface we used is deprecated since 2020.
- The CLI class has been split up into multiple classes.

### DIFF
--- a/cli/src/main/java/de/jplag/cli/CLI.java
+++ b/cli/src/main/java/de/jplag/cli/CLI.java
@@ -26,18 +26,13 @@ public final class CLI {
 
     private static final String DEFAULT_FILE_ENDING = ".zip";
 
-    private final JPlagRunner runner;
     private final CliInputHandler inputHandler;
-    private final OutputFileGenerator outputFileGenerator;
 
     /**
      * Creates a cli.
-     * @param runner The way to run JPlag
      * @param args The command line arguments
      */
-    public CLI(JPlagRunner runner, OutputFileGenerator outputFileGenerator, String[] args) {
-        this.runner = runner;
-        this.outputFileGenerator = outputFileGenerator;
+    public CLI(String[] args) {
         this.inputHandler = new CliInputHandler(args);
     }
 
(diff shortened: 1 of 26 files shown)


cli/src/main/java/de/jplag/cli/picocli/CliInputHandler.java
```diff
@@ -0,0 +1,167 @@
...
+        this.options = new CliOptions();
+        this.commandLine = buildCommandLine();
+    }
+
+    private CommandLine buildCommandLine() {
+        CommandLine cli = new CommandLine(this.options);
+        cli.setHelpFactory(new HelpFactory());
+
+        cli.getHelpSectionMap().put(SECTION_KEY_OPTION_LIST, help -> help.optionList().lines().map(it -> {
+            if (it.startsWith("  -")) {
```
uuqjz, line 65: Can you replace these magic strings?

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -1,290 +1,120 @@
...
-import picocli.CommandLine;
-import picocli.CommandLine.Model.CommandSpec;
-import picocli.CommandLine.Model.OptionSpec;
-import picocli.CommandLine.ParseResult;
 
 /**
  * Command line interface class, allows using via command line.
  * @see CLI#main(String[])
  */
 public final class CLI {
```
uuqjz, line 24: In all other classes you use `Cli` but here it is capitalized

uuqjz (approved):
Looks got to me.

tsaglam (commented):
Have you tested that the CLI help text and the log output during the run is not changed?

cli/src/main/java/de/jplag/cli/CLI.java
tsaglam: No, `CliFooBar` is the standard way to work with established abbreviations in Java identifiers.

tsaglam (approved):

Issue #1612, opened 2024-02-26
JPlag is unable to handle 'ć' character


uuqjz:
When trying to check a Spanish cpp dataset almost all submissions are discarded for containing the char 'ć'.
I am using 5.1.0 from the dev branch.
The dataset is linked below:
[z5z5.zip](https://github.com/jplag/JPlag/files/14401388/z5z5.zip)

uuqjz:
More than removing non ascii is required.
After removing every non ascii char, I get this error:
line 1:3 token recognition error at: ''
The line in question is 
`/B2017/2018: Zadaa 5, Zadatak 4`

uuqjz:
Even after this preprocessing those files cannot be parsed:
```failed to parse 'student9307.cpp'Cannot invoke "de.jplag.cpp.grammar.CPP14Parser$DeclaratorContext.pointerDeclarator()" because the return value of "de.jplag.cpp.grammar.CPP14Parser$ParameterDeclarationContext.declarator()" is null```

uuqjz:
Got the dataset from here:
https://ieee-dataport.org/open-access/programming-homework-dataset-plagiarism-detection

uuqjz:
Gimme 5 minutes

PR #1613, opened 2024-02-26
Fixed a bug, where the cpp listener could throw a null pointer exception

TwoOfTwelve:
In some cases see #1612 the cpp listener threw a null pointer exception. This PR add a check to prevent that.


uuqjz:
Fix functions as intended

uuqjz:
@tsaglam If there is no objection I'll merge

uuqjz (approved):

PR #1632, opened 2024-03-05
fix unhiding through searchbar

### LINKED ISSUE #1629: Can not unhide Names with spaces in them
When you click hide all in the report viewer and then use the search bar to search for them and unname them, the name does not become visible.

Can be observed in the [demo](jplag.github.io/Demo)

Kr0nox:
Currently it was not possible to unhide names that contained spaces through the search bar.

This PR fixes it by also splitting names at spaces and checking if every part of the string is contained in the search string parts.

Fixes #1629


sebinside (approved):

SaiPreethamD20 (commented):
Everything is clear.

PR #1680, opened 2024-04-02
Added idle bars

TwoOfTwelve:
Adds Idle Bars, that behave like progress bars, but do not show a linear progress and instead only indicate that something is happening. They can be used, when the amount of operations is unknown.


Kr0nox:
Works good in windows terminals. 
Only thing I noticed was that the bouncer did not reach the left end full. There alsways was a space between `<` and `<+>`.

Kr0nox:
I could not see it hit the left side a single time. So maybe there actually is a windows issue here?

Kr0nox:
yes, with this delay it works

tsaglam (commented):
Tested the main method under macOS, both in IDE and in the terminal. Looks good, no issues. Adapts to terminal window size.

Kr0nox:
@TwoOfTwelve I dont get a progess bar during clustering is that on purpose?

core/src/main/java/de/jplag/clustering/ClusteringFactory.java
```diff
@@ -61,6 +66,8 @@ public static List<ClusteringResult<Submission>> getClusterings(Collection<JPlag
 
         // remove bad clusters
         result = removeBadClusters(result);
+
+        progressBar.dispose();
         logClusters(result);
```
Kr0nox, line 71: I would not printing all the clusters to the console

tsaglam (approved):

Issue #1667, opened 2024-03-27
Add additional progress bars


tsaglam:
- [ ] Match merging, text could be: "Merging matched subsequences"
- [ ] Token Sequence Normalization (TokenStringNormalizer), text could be "Normalizing Token Sequence"

tsaglam:
Closed by #1690.

PR #1690, opened 2024-04-09
Added progress bars for Match Merging and Token String Normalization.

TwoOfTwelve:
Adds the progress bars from #1667.
This PR also contains a method, that allows to create progress bars using a functional syntax.


Kr0nox:
https://github.com/jplag/JPlag/blob/main/core/src/main/java/de/jplag/SubmissionSetBuilder.java Line 83 also uses a for loop and could be transformed into the new syntax. Do we want to do that?

tsaglam (approved):

PR #1691, opened 2024-04-09
Added generic catch clause in the c language module

TwoOfTwelve:
Makes sure all exceptions thrown by the Java-CC parser are caught and wrapped in a JPlag exception. This ensures, that the exception shown to the user contains the file, that the parser to fail.

Similar to #1480

### DIFF
--- a/languages/c/src/main/java/de/jplag/c/Scanner.java
+++ b/languages/c/src/main/java/de/jplag/c/Scanner.java
@@ -26,7 +26,13 @@ public class Scanner extends AbstractParser {
         for (File file : files) {
             this.currentFile = file;
             logger.trace("Scanning file {}", currentFile);
-            CPPScanner.scanFile(file, this);
+            try {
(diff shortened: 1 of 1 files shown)


languages/c/src/main/java/de/jplag/c/Scanner.java
```diff
@@ -26,7 +26,13 @@ public List<Token> scan(Set<File> files) throws ParsingException {
         for (File file : files) {
             this.currentFile = file;
             logger.trace("Scanning file {}", currentFile);
-            CPPScanner.scanFile(file, this);
+            try {
+                CPPScanner.scanFile(file, this);
+            } catch (ParsingException e) {
+                throw e;
+            } catch (Exception e) {
+                throw new ParsingException(file, "Unexpected error during parsing.\n" + e.getMessage(), e);
```
tsaglam, line 34: Just noticed: System-dependent line separator.

PR #1696, opened 2024-04-09
Update develop from main

Kr0nox:
Updates develop from main after #1688


tsaglam:
For some reason it shows 0 files changed in the PR summary. The PR branch has `"vite": "^5.2.6"`. Probably a merging issue?

Kr0nox:
No, it is actually correct.
Develop is already on a never version of vite (5.2.6), while main just update to 5.1.7.
This means develop already has all the security updates main just got.
Merging develop into main, would still lead to merge conflicts, but since we dont need the vite update from main in develop for security reasons, I would leave it up to you whether we merge this now, or before the next release

tsaglam:
Ah, I missed the minor number. I will merge nonetheless, hoping the dependabot commit is then no longer counted as commit behind main for develop.

PR #1697, opened 2024-04-09
Fix Jar Upload

### LINKED ISSUE #1597: Error in build test workflow
The build workflow (`mayen.yml`) should append the build jar as an artifact to the workflow.

This does not happen, since the path the jar is not given correctly to the action.

Kr0nox:
fixes #1597 
Jar is now correctly attached to the workflow run


tsaglam (approved):
[Looks good!](https://github.com/jplag/JPlag/actions/runs/8619873379?pr=1697)

Issue #1670, opened 2024-03-27
Avoid overriding result files


tsaglam:
Try to check if a file can be written before executing JPlag. If a file already exists, add an increasingly incrementing number as a suffix.

PR #1716, opened 2024-04-18
Avoided overwriting of files

TwoOfTwelve:
Avoids overwriting of existing results files. Tries to find a new unique filename by adding (<number>). Aborts if the selected file is not writable. Also adds a cli option --overwrite, that ignores existing files. This still aborts if the file cannot be written.

Solved #1670

### DIFF
--- a/cli/src/test/java/de/jplag/cli/CheckResultFileWritableTest.java
+++ b/cli/src/test/java/de/jplag/cli/CheckResultFileWritableTest.java
@@ -29,7 +29,7 @@ public class CheckResultFileWritableTest extends CommandLineInterfaceTest {
     }
 
     @Test
-    public void testNonExistingWritableFile() throws Throwable {
+    void testNonExistingWritableFile() throws Throwable {
         File directory = Files.createTempDirectory("JPlagTest").toFile();
         File targetFile = new File(directory, "results.zip");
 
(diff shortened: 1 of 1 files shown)


tsaglam:
@TwoOfTwelve could you add a small test for this?

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -126,6 +131,34 @@ private String getResultFileBaseName() {
...
+            retryAttempt++;
+        }
+
+        String targetFileName = this.getOffsetFileName(retryAttempt);
+        File targetFile = new File(targetFileName);
+        if (!this.inputHandler.getCliOptions().advanced.overwrite && targetFile.exists()) {
+            throw new CliException(OUTPUT_FILE_EXISTS);
+        }
+
+        if (!(targetFile.canWrite() || (!targetFile.exists() && targetFile.getAbsoluteFile().getParentFile().canWrite()))) {
```
Kr0nox, line 155: I would suggest refactoring this expression to make it a bit more readable

cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -99,6 +99,9 @@ public static class Advanced {
 
         @Option(names = "--csv-export", description = "Export pairwise similarity values as a CSV file.")
         public boolean csvExport = false;
+
+        @Option(names = "--overwrite", description = "If this option is specified and a result file already exists it will be overwritten.")
```
Kr0nox, line 103: I would change this text to be more online with other flags. Meaning: Just state what it does without the `If this option is specified`

Kr0nox (changes requested):

tsaglam (changes requested):
One minor thing plus the 5 small sonar issues left todo.

tsaglam (approved):

Kr0nox (approved):

PR #1745, opened 2024-05-02
Add workflow for closing issues on develop merge

### LINKED ISSUE #1597: Error in build test workflow

Kr0nox:
This PR adds a workflow that closes issues when they get merged into the develop branch.

The workflow itself starts a python script for this.
This script uses the GitHub GraphQL API to get all issues that were linked using the sidebar. It then parses the text of the PR to get all issues that are linked using closing keywords.
It then uses the REST API to close the linked issues.

As an example we have linked this issue:
closes #1597

### DIFF
--- a/.github/workflows/closeDevIssues.yml
+++ b/.github/workflows/closeDevIssues.yml
@@ -1,0 +1,21 @@
+name: Close PR on develop merge
+
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  close_issues:
+    if: github.event.pull_request.merged == true && github.base_ref == 'develop'
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Checkout 🛎️
+      uses: actions/checkout@v4
+      
+    - name: Run script
+      working-directory: .github/workflows/scripts
+      run: |
+        pip install requests
+        python closeDevIssues.py ${{ secrets.GITHUB_TOKEN }} ${{ github.event.pull_request.number }}
--- a/.github/workflows/closeDevIssues.yml
+++ b/.github/workflows/closeDevIssues.yml
@@ -18,4 +18,4 @@ jobs:
       working-directory: .github/workflows/scripts
       run: |
         pip install requests
-        python closeDevIssues.py ${{ secrets.GITHUB_TOKEN }} ${{ github.event.pull_request.number }}
+        python closeDevIssues.py ${{ secrets.GITHUB_TOKEN }} jplag JPlag ${{ github.event.pull_request.number }}
--- a/.github/workflows/scripts/closeDevIssues.py
+++ b/.github/workflows/scripts/closeDevIssues.py
@@ -25,7 +25,7 @@ query = f"""
 }}
 """
 
-def run_query(query): # A simple function to use requests.post to make the API call. Note the json= section.
+def run_query(query):
     request = requests.post('https://api.github.com/graphql', json={'query': query}, headers=headers)
     if request.status_code == 200:
         return request.json()
--- a/.github/workflows/scripts/closeDevIssues.py
+++ b/.github/workflows/scripts/closeDevIssues.py
@@ -4,9 +4,9 @@ import sys
 
 headers = {"Authorization": f"Bearer {sys.argv[1]}"}
 
-owner = "jplag"
-repo = "JPlag"
-pr_number = int(sys.argv[2])
+owner = sys.argv[2]
+repo = sys.argv[3]
+pr_number = int(sys.argv[4])
 
 query = f"""
 {{
--- a/.github/workflows/scripts/closeDevIssues.py
+++ b/.github/workflows/scripts/closeDevIssues.py
@@ -61,7 +61,7 @@ result = run_query(query)['data']['repository']['pullRequest']
 issues = get_linked_issues(result)
 
 def close_issue(issue_number):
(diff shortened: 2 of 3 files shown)


tsaglam:
@dfuchss this is the PR I mentioned.

.github/workflows/closeDevIssues.yml
```diff
@@ -0,0 +1,21 @@
+name: Close PR on develop merge
```
tsaglam, line 1: Should probably be: 
```suggestion
name: Close linked issue on PR merge
```

.github/workflows/closeDevIssues.yml
```diff
@@ -0,0 +1,21 @@
...
+    
+    steps:
+    - name: Checkout 🛎️
+      uses: actions/checkout@v4
+      
+    - name: Run script
+      working-directory: .github/workflows/scripts
+      run: |
+        pip install requests
+        python closeDevIssues.py ${{ secrets.GITHUB_TOKEN }} ${{ github.event.pull_request.number }}
```
dfuchss, line 21: Maybe add owner & repo as parameters as well :) Then, we can recycle the action better :)

.github/workflows/scripts/closeDevIssues.py
```diff
@@ -0,0 +1,74 @@
...
+        nodes {{
+          number
+        }}
+      }}
+    }}
+  }}
+}}
+"""
+
+def run_query(query): # A simple function to use requests.post to make the API call. Note the json= section.
```
dfuchss, line 28: Why this comment?

.github/workflows/scripts/closeDevIssues.py
```diff
@@ -0,0 +1,74 @@
...
+
+def close_issue(issue_number):
+    requests.patch(f"https://api.github.com/repos/{owner}/{repo}/issues/{issue_number}", json={"state": "closed"}, headers=headers)
+
+if result['baseRefName'] != "develop":
+    print("PR not merged to develop, not closing issues")
+elif result['merged']:
+    print(f"Closing issues: {issues}")
+    for issue in issues:
+        close_issue(issue)
```
dfuchss, line 71: Could you add a comment, why the issue is closed to the closing process :)

dfuchss (changes requested):
Sounds good. Just two little improvement suggestions and then it should be ready to merge from my point of view. Nevertheless, I think it has to be merged to main too, that it works.

.github/workflows/scripts/closeDevIssues.py
Kr0nox: Do you mean just sending a message before closing, like `Closed by #PR_NUMBER`?

dfuchss: Yes

dfuchss:
> > Nevertheless, I think it has to be merged to main too, that it works.
> 
> In my testing it always worked when I did not have the workflow on main and also always took the newest version of the workflow, that was not on main. This would mean it takes the workflow from either the target branch or the branch the PR comes from

Ok. Then it might not be needed.

dfuchss (approved):

Kr0nox:
Now that we have the opportunity to close the issues with a custom message, do we want to add that the feature or bug fix will be released with the next release, so users do not think the feature is there already.
Currently, the bot just says `Closed by #NUMBER`. We could have them say something like:
```
Closed by #NUMBEr
The fix/feature will be part of the next release. If you want it now, you can build JPlag from the sources of the develop branch.
```

What are your opinions on that?

tsaglam:
I think I would this is probably not needed, as the users who actually follow the development here are aware that there is a delay between the PR merge and the feature release. I would rather just stick to `Closed by #NUMBER.`

.github/workflows/scripts/closeDevIssues.py
```diff
@@ -0,0 +1,75 @@
...
+        closing_issues.extend(subsequences_matching_regex(result['body'].lower(), f'{keyword} #([0-9]+)'))
+    for k in result['closingIssuesReferences']['nodes']:
+        closing_issues.append(k['number'])
+    return list(set(closing_issues))
+
+result = run_query(query)['data']['repository']['pullRequest']
+issues = get_linked_issues(result)
+
+def close_issue(issue_number):
+    requests.post(f"https://api.github.com/repos/{owner}/{repo}/issues/{issue_number}/comments", json={"body": f"Closed by #{pr_number}"}, headers=headers)
```
tsaglam, line 64: Just adds the dot.
```suggestion
    requests.post(f"https://api.github.com/repos/{owner}/{repo}/issues/{issue_number}/comments", json={"body": f"Closed by #{pr_number}."}, headers=headers)
```

Kr0nox:
> the users who actually follow the development here

My though was that users that post issues might not do that and expect closed issues to be solved and ready for use, but in the end its up to you

tsaglam:
At least from my experience, we have not yet received questions regarding such a case. We can always add it if this appears more frequently than in the past.

tsaglam:
@Kr0nox The workflow affected more PRs than expected, looks like a bug? (see referenced issues above)

dfuchss:
Yes .. 15 -- 159 .. 1597 .. there is probably a \b or similar missing :)

Kr0nox:
Yes, we did not check for a non number to be after the issue number, so it found all substrings of numbers of the actual issue.
#1748 provides a fix for this

Issue #1669, opened 2024-03-27
Revisit Python Language module


tsaglam:
Our python module seems to be frequently used, so we should take a look at it again:

- [x] Update grammar based on https://github.com/antlr/grammars-v4/pull/3540 (probably only the expression change)
- [x] Discuss if we want to close this: https://github.com/jplag/JPlag/issues/1481
- [x] Check what language features are new since 3.6 and whether we need them?

PR #1750, opened 2024-05-03
Update python language module

TwoOfTwelve:
Added python grammar changes from the antlr repository.
Also migrated the language module to the new framework.

The token creation has not been altered, except for two new tokens for "async" and "await". That feature has been added to python since the language module was created. Except for that I did not find any big changes that JPlag needs to reflect.

Addresses #1669


Kr0nox:
Since after this the state of the python language module is no longer that of the legacy version do we want to change that in the README and docs?

tsaglam:
@TwoOfTwelve could you elaborate in the PR description whether this PR changes anything regarding the tokenization or support of python language features?

tsaglam (approved):

tsaglam:
> Since after this the state of the python language module is no longer that of the legacy version do we want to change that in the README and docs?

Yes, we should change it to `beta`. @TwoOfTwelve can you adapt the docs?

Issue #1719, opened 2024-04-20
Progessbars cut off error messages


Kr0nox:
When failing to parse a submission, the Progess bar can cut off the error message.

How it was displayed in my terminal:
![grafik](https://github.com/jplag/JPlag/assets/39801116/79b0f5ac-bd00-44d5-836d-0cedfd76c074)

After trying to expand the terminal to see more:
![grafik](https://github.com/jplag/JPlag/assets/39801116/288d0d48-a64c-40e8-9c52-e6e95ed06b4c)

PR #1751, opened 2024-05-03
Added command line option for log-level

TwoOfTwelve:
Added a cli option to set the log-level


Kr0nox:
Do we want the progress bars on level that shows less than info? (WARN, ERROR)
Do we want to keep the progress bars on DEBUG and TRACE? They are barely visible anyway and due to #1719 the additional messages are cut off sometimes

cli/src/main/java/de/jplag/cli/logger/JPlagLoggerBase.java
```diff
@@ -16,14 +16,12 @@ public abstract class JPlagLoggerBase extends AbstractLogger {
 
     private static final Level LOG_LEVEL_FOR_EXTERNAL_LIBRARIES = LOG_LEVEL_ERROR;
 
-    private final Level currentLogLevel;
+    public static Level currentLogLevel = LOG_LEVEL_INFO;
```
tsaglam, line 19: I would prefer to keep this private and use a static setter instead.

tsaglam (changes requested):
One small thing and a sonar issue.

Kr0nox (changes requested):
Readding my comment from above as requested changes, to not forget it before merging:
> Do we want the progress bars on level that shows less than info? (WARN, ERROR) Users that select these levels probably actually only want warnings and errors

> Do we want to keep the progress bars on DEBUG and TRACE? They are barely visible anyway and due to https://github.com/jplag/JPlag/issues/1719 the additional messages are cut off sometimes

In the meeting we decided we dont want the bar on any level other than info

tsaglam:
One conflict has to be solved before I can merge.

Kr0nox (approved):

tsaglam (approved):

Issue #1497, opened 2024-01-23
Use of getName method of Langauge


Kr0nox:
Currently the `getName` method of the `Language` class returns the name of the Parser for most languages. This could be changed to just return the name of the language.
This would also shorten the language list message printed on the start of the CLI

Mintype:
I think my pull request will solve this issue.

PR #1767, opened 2024-05-16
Update getName method in used in Language class

Mintype:
Hi! This pull request does everything that issue #1497 asks for, which is to clean up the getName() methods of many of the languages. Now they just will return the language name and not the parser name. Please tell me if I did anything wrong or if I have to change anything. Thanks a lot!


languages/csharp/src/main/java/de/jplag/csharp/CSharpLanguage.java
```diff
@@ -25,7 +25,7 @@ public String[] suffixes() {
 
     @Override
     public String getName() {
-        return NAME;
+        return "C#";
```
tsaglam, line 28: With this change, you produce inconsistencies with the `name` constant. You should probably just adapt the constant value instead.

tsaglam:
Just as general feedback, while granular committing is great, your commits are very atomic, meaning they encapsulate very few changes, which would make them generally considered a bit too fine-grained. This is not a problem for this PR though, as I can squash on merge.

tsaglam:
>  Where do you think this PR will be included?

I see this as a v6.0.0 feature, we can probably just merge this after the v5.1.0 release.

Mintype:
> You will also need to update all names here https://github.com/jplag/JPlag/blob/develop/report-viewer/src/model/Language.ts
> 
> Depending on whether this will be a 5.1.0 or a 6.0.0 feature, we should keep the old names, to not break highlighting in these languages. @tsaglam Where do you think this PR will be included?

Ok I'll do this!

Mintype:
wait a second. i forked the main branch not develop so I cant find https://github.com/jplag/JPlag/blob/develop/report-viewer/src/model/Language.ts 

Should I close this pull request and make a new one with the right branch?

PR #1769, opened 2024-05-18
Shorten Languages in getName() method.

Mintype:
This is for issue #1497 and I think this should be all.

### DIFF
--- a/languages/csharp/src/main/java/de/jplag/csharp/CSharpLanguage.java
+++ b/languages/csharp/src/main/java/de/jplag/csharp/CSharpLanguage.java
@@ -9,7 +9,7 @@ import de.jplag.antlr.AbstractAntlrLanguage;
  */
 @MetaInfServices(de.jplag.Language.class)
 public class CSharpLanguage extends AbstractAntlrLanguage {
-    private static final String NAME = "C#";
+    private static final String NAME = "C# 6 Parser";
     private static final String IDENTIFIER = "csharp";
     private static final String[] FILE_ENDINGS = new String[] {".cs", ".CS"};
     private static final int DEFAULT_MIN_TOKEN_MATCH = 8;
@@ -25,7 +25,7 @@ public class CSharpLanguage extends AbstractAntlrLanguage {
 
     @Override
     public String getName() {
-        return NAME;
+        return "C#";
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


languages/csharp/src/main/java/de/jplag/csharp/CSharpLanguage.java
```diff
@@ -25,7 +25,7 @@ public String[] suffixes() {
 
     @Override
     public String getName() {
-        return NAME;
```
Kr0nox, line 28: I would suggest either removing the name constant or changing the constant itself

languages/scheme/src/main/java/de/jplag/scheme/SchemeLanguage.java
```diff
@@ -26,7 +26,7 @@ public String[] suffixes() {
 
     @Override
     public String getName() {
-        return "SchemeR4RS Parser [basic markup]";
+        return "SchemeR4RS [basic markup]";
```
Kr0nox, line 29: @tsaglam Is it wanted that we have `[basic markup]` here?

languages/text/src/main/java/de/jplag/text/NaturalLanguage.java
```diff
@@ -31,7 +31,7 @@ public String[] suffixes() {
 
     @Override
     public String getName() {
-        return "Text Parser (naive)";
+        return "Text (naive)";
```
Kr0nox, line 34: @tsaglam Do we want it to say `(naive)`?

report-viewer/src/model/Language.ts
Kr0nox, line 1: Some names do not reflect the names in the java classes. 
- Text
- Scheme
But with that we can wait for @tsaglam opinion on my above comments

Kr0nox (changes requested):
Some small things I found. Some of them need @tsaglam opinion before changing

languages/csharp/src/main/java/de/jplag/csharp/CSharpLanguage.java
Mintype: Ok I made one commit right now for this.

dfuchss:
There are also some CI issues :)

languages/scheme/src/main/java/de/jplag/scheme/SchemeLanguage.java
tsaglam: This is a legacy thing, I am not even sure what exactly Scheme basic markup is. I am both fine with removing and leaving it.

tsaglam:
Closed due to inactivity, happy to open it again in case you want to continue.

Mintype:
> Closed due to inactivity, happy to open it again in case you want to continue.

Sure, I can continue, but what needs to be done specifically?

tsaglam:
The CI build is failing due to multiple reasons.

- Spotless style check: Always run `mvn spotless:apply` before commiting
- CI Build: You code does not compile, click the red X next to the latest commit above to find out why.
- When the code passes the CI, I will review it.

PR #1796, opened 2024-06-04
Refactor Json Match

Kr0nox:
Cleans up the Json for matches and removes duplication


core/src/main/java/de/jplag/reporting/reportobject/model/CodePosition.java
```diff
@@ -0,0 +1,7 @@
+package de.jplag.reporting.reportobject.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record CodePosition(@JsonProperty("line") int lineNumber, @JsonProperty("column") int column,
```
dfuchss, line 5: Please document stuff like "zero-indexed" etc. for the new records :)

tsaglam (approved):

PR #1838, opened 2024-07-02
Fix bugs for WSL

Kr0nox:
When opening the local report viewer on WSL, JPlag terminated since no browser was found. Now there is an info message that the browser could not be opened.
grafik

~~Another issue was that WSL did not access the windows localhost with the given IP Address. Thus, I changed it to the universal Address `0.0.0.0`~~

### DIFF
--- a/cli/src/main/java/de/jplag/cli/JPlagRunner.java
+++ b/cli/src/main/java/de/jplag/cli/JPlagRunner.java
@@ -43,7 +43,11 @@ public final class JPlagRunner {
         ReportViewer reportViewer = new ReportViewer(zipFile, port);
         int actualPort = reportViewer.start();
         logger.info("ReportViewer started on port http://localhost:{}", actualPort);
-        Desktop.getDesktop().browse(URI.create("http://localhost:" + actualPort + "/"));
+        if (Desktop.isDesktopSupported()) {
+            Desktop.getDesktop().browse(URI.create("http://localhost:" + actualPort + "/"));
+        } else {
+            logger.info("Could not open browser. You can open the Report Viewer here: http://localhost:{}/", actualPort);
+        }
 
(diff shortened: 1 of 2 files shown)


Kr0nox:
@TwoOfTwelve @tsaglam Could you quickly verify that the local mode still works on your os

tsaglam:
> Could you quickly verify that the local mode still works on your os

Works for me (macOS Sonoma).

tsaglam (approved):

dfuchss (changes requested):
Could you test binding to `127.0.0.1` instead of `localhost` or `0.0.0.0` ? Even in the subsystem `python -m http.server -b 127.0.0.1` is accessible for me. So the binding to `0.0.0.0` might not be needed

Kr0nox:
> Could you test binding to 127.0.0.1 instead of localhost or 0.0.0.0 ? Even in the subsystem python -m http.server -b 127.0.0.1 is accessible for me. So the binding to 0.0.0.0 might not be needed

`127.0.0.1` does not work for me

dfuchss:
> > Could you test binding to 127.0.0.1 instead of localhost or 0.0.0.0 ? Even in the subsystem python -m http.server -b 127.0.0.1 is accessible for me. So the binding to 0.0.0.0 might not be needed
> 
> `127.0.0.1` does not work for me

Ok .. I'll test it later :) Still, as default I'd not bind it to 0.0.0.0

dfuchss:
@Kr0nox I've updated the PR. The trick is to ensure that Java understands that the loopback address is IPv4. Please test it on your WSL again :) At my machine that works now :)

Kr0nox:
Yes, works on my system now

dfuchss (approved):

PR #1839, opened 2024-07-03
Enhance search bar of comparison table

Kr0nox:
This table adds new features to the search bar of the comparison table.

It is now possible to search for index. If a number is written, it will take all indexes containing this number. E.g. when `5` is written in the search bar 5, 15 and 51 are displayed.
To get a specific index only, the number can be prefaced with `index:`. E.g. `index:5` will only filter the 5th comparison in the current sorting.

It is now possible to filter for similarities. All these filters need to be prefaced with either >, <, >= or <= followed by the number of percentages to filter for. This searches over all similarities. A percentage symbol is optional. E.g. `<100` will show all comparisons where any metric (so either avg or max) is less than 100%.
It is possible to filter for a specific metric by prefacing the comparison symbol with its name in the table header, followed by a double colon. E.g. `max:>=90%` will filter for submission where the maximum similarity is greater than or equal to 90%.

This is also explained in the tooltip of the search bar. 
Due to the tooltip being cutoff, its direction changed.

### DIFF
--- a/report-viewer/src/components/ComparisonTableFilter.vue
+++ b/report-viewer/src/components/ComparisonTableFilter.vue
@@ -2,7 +2,7 @@
   <div class="space-y-2">
     <div class="flex flex-row flex-wrap items-center gap-x-8 gap-y-2">
       <h2>{{ header }}</h2>
-      <ToolTipComponent direction="bottom" class="min-w-[50%] flex-grow">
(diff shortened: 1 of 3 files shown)


tsaglam (changes requested):
One small thing.

tsaglam (approved):
Looks good!

Issue #1848, opened 2024-07-08
Feature requeset: same name submissions


euberdeveloper:
I have a case in which I have submissions (or more precisely old submissions) with the same name.

[This](https://github.com/jplag/JPlag/blob/570efab64f89ad368bf262d1c1e57710557c0fed/core/src/main/java/de/jplag/JPlag.java#L121) piece of code blocks them, but this is problematic, I have to do a directory refactoring just to be able to run JPlag.

My directory structure looks like that:

```
2023
	ms1
		gr1
		gr2 
		gr3
	ms2
		gr1 
		gr2
		gr3
	ms3
		gr1
		gr2
		gr3
	ms4
		gr1
		gr2
		gr3


2022
	ms1
		gr1
		gr2 
		gr3
	ms2
		gr1
		gr2
		gr3
	ms3
		gr1
		gr2
		gr3
	ms4
		gr1
		gr2
		gr3


2021
	ms1
		gr1 
		gr2 
		gr3
	ms2
		gr1
		gr2
		gr3
	ms3
		gr1
		gr2
		gr3
	ms4
		gr1
		gr2
		gr3
```

If I am checking in the current submissions "2023/ms1" (milestone1 of all groups in 2023), I would put as old submissions ["2022/ms1", "2021/ms1", "2020/ms1"] (milestone1 of all groups of the previous years 2022,2021,2020).

Hence, ms1 would be "duplicate" for all the previous years.
A solution would be, in case the path is different but the name is the same, to take the shorter parent path to make them different, in this case taking as namess ["2022/ms1", "2021/ms1] instead of ["ms1","ms1"]

Kr0nox:
This sounds like it is related to #1445

tsaglam:
Closed, as it is already documented in #1445.

PR #1850, opened 2024-07-08
feat: add possibility of using submissions with same name

euberdeveloper:
Ref to #1445 and  #1848


tsaglam:
Thanks for the PR; I will take a look at it. To fix the formatting issues, just run `mvn spotless:apply` and commit again.

euberdeveloper:
Thank you, I'll do it

Issue #1849, opened 2024-07-08
Feature request: single submission against "older submissions"


euberdeveloper:
I have a particular usecase.

Starting the current semester 2023 like this:

```
2023
	ms1
	ms2
        ms3
        ms4

2022
	ms1
		gr1
		gr2 
		gr3
	ms2
        ms3
        ms4
2021
	ms1
		gr1 
		gr2 
		gr3
	ms2
        ms3
        ms4
```

The first group "gr1" does a submission on a website. The WebServer automatically clones the code into the path "2023/ms1/gr1" and obtains:

```
2023
	ms1
		gr1
	ms2
        ms3
        ms4

2022
	<same as before>
2021
	<same as before>
```

This submission "2023/ms1/gr1" **is to be tested only against the older submissions** "2022/ms1" and "2021/ms1". There are no other current submissions so the current submissions is a **Set of size 1**. 

**This gets refused by JPlag**

Still, when gr2 submits "2023/ms1/gr2" I obtain:

```
2023
	ms1
		gr1
		gr2
	ms2
        ms3
        ms4

2022
	<same as before>
2021
	<same as before>
```

Here I still want only one group, the submitting one which is "gr2", to be tested against the gr1 submission of 2023 and all the previous submissions of the previous years. So I would have as current submissions the set "2023/ms1/gr2" and as older submissions ["2023/ms1/gr1", "2022/ms1", "2021/ms1"]

Again, this is not possible because the current submissions set is of size 1 and **Jplag doesn't allow it**

euberdeveloper:
I have done an implementation of a possible solution (more like a proof of concept) into this branch: https://github.com/euberdeveloper/JPlag/tree/feature/allow-duplicate-submission-names

PR #1852, opened 2024-07-09
Feature/allow single submissions

euberdeveloper:
Ref to #1445 and #1848, fixes #1850

### DIFF
--- a/core/src/main/java/de/jplag/JPlag.java
+++ b/core/src/main/java/de/jplag/JPlag.java
@@ -1,11 +1,8 @@
 package de.jplag;
 
 import java.io.File;
-import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
 import java.util.ResourceBundle;
-import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


euberdeveloper:
It fixes also this feature request: #1849.

The code probably is to be reviewed by the maintainers and re-adjusted, but it works as a proof of concept

tsaglam:
@euberdeveloper reviews will follow. For now, please take a look at the Sonar issues and see which ones you can resolve.

Kr0nox (changes requested):
Could you add a simple e2e case to this pr, so we can make sure, it is always correctly exported.
Here are references on where they are implemented currently:
https://github.com/jplag/JPlag/blob/develop/report-viewer/tests/e2e/OpenComparisonTest.spec.ts
https://github.com/jplag/JPlag/blob/develop/.github/workflows/complete-e2e.yml

euberdeveloper:
Ok, I will do it next week

tsaglam:
@euberdeveloper can you also extend the PR description so it becomes clearer what aspects this PR addresses and how it does so? Also how we can test it. We may need to discuss what features we want to include and which ones we don't want to include. Finally, please switch your target branch to develop, the main branch is only for releases.

euberdeveloper:
@tsaglam I've created a new branch that forks that:
* branches develop instead of master
* uses spotless to avoid style issues
* adds tests that prove that the new behaviour works and work also as an example to understand it

Please take a look at it, I think that it is a good improvement that fixes #1849 and #1848 .

For a description of what it does, you can take a look at those issues (#1849  and #1848 ), namely:
* If I want to test a single new submission against many old ones, it works instaed of throwing an error
* If I have, for example, a new submission called "submission" and an old submission called also "submission", instead of throwing an error, requiring me to manually change the names of the folders, it gets as submission name the min. different path, e.g. "submissions/submission" and "olds/submission".
* I find the old behaviour of throwing errors in those 2 cases limiting the tool and requiring users to find tricks to avoid those errors

The pull request will reference this old one, that I'm closing

PR #1853, opened 2024-07-09
Enable Preview-Features only if Java Version matches the Java Version of JPlag

dfuchss:
See https://github.com/jplag/JPlag/discussions/1851


dfuchss:
Another solution has been found :)

PR #1854, opened 2024-07-09
FIx bug with --enable-preview on Java 22

tsaglam:
Fix a bug occurring on Java 22 caused by the `--enable-preview` option that is enabled for the Java language module (see #1851):

- Use the runtime Java version as the release version for the JavaC compilation task instead of the hardcoded version 21.
- Only use hardcoded language version in the Java feature test.
- Use constants for options flags.
- Use a short link in the StackOverflow comment.

### DIFF
--- a/languages/java/src/main/java/de/jplag/java/JavacAdapter.java
+++ b/languages/java/src/main/java/de/jplag/java/JavacAdapter.java
@@ -29,6 +29,10 @@ import com.sun.source.util.Trees;
 
 public class JavacAdapter {
 
+    private static final String NO_ANNOTATION_PROCESSING = "-proc:none";
+    private static final String PREVIEW_FLAG = "--enable-preview";
+    private static final String RELEASE_VERSION_FLAG = "--release=";
+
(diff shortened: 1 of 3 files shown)


languages/java/src/main/java/de/jplag/java/JavacAdapter.java
```diff
@@ -29,6 +29,10 @@
 
 public class JavacAdapter {
 
+    private static final String NO_ANNOTATION_PROCESSING = "-proc:none";
+    private static final String PREVIEW_FLAG = "--enable-preview";
+    private static final String RELEASE_VERSION_FLAG = "--release=";
```
dfuchss, line 34: tbh, I think it's not a "flag" :D

dfuchss (approved):

languages/java/src/main/java/de/jplag/java/JavacAdapter.java
tsaglam: Fair! :D

PR #1864, opened 2024-07-10
Refactored RLang language module to new framework

TwoOfTwelve:
I refactored the code for the rlang language module for the new framework. In doing so I noticed that the ANTLR grammar isn't able to handle the test source. I've created a PR in the ANTLR repository (https://github.com/antlr/grammars-v4/pull/4160) that should fix these issues. I suggest we wait for that PR to be resolved before proceeding with this.

There is a second PR in the ANTLR repository regarding the usability of the generated AST: [ANTLR PR](https://github.com/antlr/grammars-v4/pull/4215)


languages/rlang/src/main/java/de/jplag/rlang/RListener.java
```diff
@@ -0,0 +1,105 @@
...
+import de.jplag.rlang.grammar.RParser.TrueContext;
+import de.jplag.rlang.grammar.RParser.WhileContext;
+
+/**
+ * Contains mapping for RLang to create JPlag tokens from ANTLR tokens
+ */
+public class RListener extends AbstractAntlrListener {
+    public RListener() {
+        header();
+        functions();
```
Kr0nox, line 59: imo these function names do not express what the functions do. It might be clear from the context of the class, but me should discuss whether we want to have more explicit names

languages/rlang/src/main/java/de/jplag/rlang/RListener.java
```diff
@@ -0,0 +1,105 @@
...
+        visit(FalseContext.class).map(BOOL);
+    }
+
+    private void controlStructures() {
+        visit(IfContext.class).map(IF_BEGIN, IF_END);
+        visit(IfElseContext.class).map(IF_BEGIN, IF_END);
+        visit(ForContext.class).map(FOR_BEGIN, FOR_END);
+        visit(WhileContext.class).map(WHILE_BEGIN, WHILE_END);
+        visit(RepeatContext.class).map(REPEAT_BEGIN, REPEAT_END);
+
```
Kr0nox, line 95: Do we want to have three separate loop tokens here or combine them into one?
But these changes could also be subject of a different PR.

Kr0nox (changes requested):
Some things to discuss in the meeting

Kr0nox (approved):

tsaglam (approved):

tsaglam:
@TwoOfTwelve only the conflicts are left.

PR #1866, opened 2024-07-11
Avoid progress bar overlap

TwoOfTwelve:
Avoids overlaps of the progress bar with log outputs by delaying the log outputs while a progress bar is being shown.

Log outputs that happen during the progress bar are instead shown below it, after it is done:
```
Loading Submissions   100% [=========================] 2/2 (0:00:00 / 0:00:00) 
Parsing Submissions     0% [                               ] 0/2 (0:00:00 / ?) line 5:12 mismatched input '=' expecting '{'
line 5:12 mismatched input '=' expecting '{'
Parsing Submissions   100% [=========================] 2/2 (0:00:00 / 0:00:00) 
2024-07-11-12:01:45_008 [WARN] Submission - Failed to parse submission test1.cs:
failed to parse 'test1.cs'class de.jplag.csharp.grammar.CSharpParser$Namespace_or_type_nameContext cannot be cast to class de.jplag.csharp.grammar.CSharpParser$Local_variable_declarationContext (de.jplag.csharp.grammar.CSharpParser$Namespace_or_type_nameContext and de.jplag.csharp.grammar.CSharpParser$Local_variable_declarationContext are in unnamed module of loader 'app')
2024-07-11-12:01:45_012 [WARN] Submission - Failed to parse submission test2.cs:
failed to parse 'test2.cs'class de.jplag.csharp.grammar.CSharpParser$Namespace_or_type_nameContext cannot be cast to class de.jplag.csharp.grammar.CSharpParser$Local_variable_declarationContext (de.jplag.csharp.grammar.CSharpParser$Namespace_or_type_nameContext and de.jplag.csharp.grammar.CSharpParser$Local_variable_declarationContext are in unnamed module of loader 'app')
```

There is still some overlap with antlr, because antlr writes directly to System.err. That problem is addressed in #1868.


cli/src/main/java/de/jplag/cli/logger/TerminalPrinter.java
```diff
@@ -0,0 +1,87 @@
+package de.jplag.cli.logger;
+
+import java.io.PrintStream;
+import java.util.PriorityQueue;
+import java.util.Queue;
+
+/**
+ * Prints strings to stdout. Provides the option to delay the actual printing.
+ */
+public class TerminalPrinter {
```
tsaglam, line 10: `DelayablePrinter` is probably more fitting, what do you think?

cli/src/main/java/de/jplag/cli/logger/TerminalPrinter.java
```diff
@@ -0,0 +1,87 @@
...
+        synchronized (TerminalPrinter.class) {
+            if (instance == null) {
+                synchronized (TerminalPrinter.class) {
+                    instance = new TerminalPrinter();
+                }
+            }
+        }
+
+        return instance;
+    }
```
tsaglam, line 27: I would either synchronize the whole method (draconian synchronization) or do the double-checked locking pattern here instead of synchronizing twice.

cli/src/main/java/de/jplag/cli/logger/TerminalPrinter.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private final Queue<String> outputQueue;
+    private PrintStream targetStream;
+
+    private boolean isDelayed;
```
tsaglam, line 32: Move those at the top with the other fields for consistency's sake.

cli/src/main/java/de/jplag/cli/logger/TerminalPrinter.java
```diff
@@ -0,0 +1,87 @@
...
+        this.outputQueue = new PriorityQueue<>();
+        this.targetStream = System.out;
+        this.isDelayed = false;
+    }
+
+    /**
+     * Prints the given string to the terminal appending a line-break
+     * @param output The string to print
+     */
+    public void println(String output) {
```
tsaglam, line 44: If you are synchronizing the whole method body anyways and the class is the monitor, just add `synchronized` as method keyword.

cli/src/main/java/de/jplag/cli/logger/TerminalPrinter.java
```diff
@@ -0,0 +1,87 @@
...
+    public void delay() {
+        synchronized (TerminalPrinter.class) {
+            this.isDelayed = true;
+        }
+    }
+
+    /**
+     * Resumes printing if {@link #delay()} was called
+     */
+    public void unDelay() {
```
tsaglam, line 63: I think I prefer `resume()`...

tsaglam (changes requested):
There are a few issues regarding sync and naming here. Also, please check the Sonar issues.

cli/src/main/java/de/jplag/cli/logger/TerminalPrinter.java
tsaglam: Interesting, did it give you input as to why it wants this implementation?

tsaglam (approved):

Kr0nox:
grafik
Do we want the errors to be printed like they are in the screenshot or after a specific progress bar has run to 100%.
With the current way, a specific progress bar may be printed multiple times with different percentages.

tsaglam:
> Do we want the errors to be printed like they are in the screenshot or after a specific progress bar has run to 100%. With the current way, a specific progress bar may be printed multiple times with different percentages.

Both would be possible. Without the duplication, it looks cleaner. With the duplication, you get feedback earlier. For large datasets, this allows you to already check out the erroneous submissions while JPlag is still running.

@jplag/hiwi @jplag/maintainer any preferences?

Kr0nox:
I think printing them after would be better.
1) Then all errors are accumulated together and not split between multiple instances of the progress bar
2) It is easier to follow the process of the current taks, when you dont have to switch progress bars.
But that is just my opinion. Feel free to object

Kr0nox:
grafik
There also seems to be an issue with printing. These two failed submission are exact copies of each other, but the reason is only printed for one of them

Kr0nox:
> There also seems to be an issue with printing. These two failed submission are exact copies of each other, but the reason is only printed for one of them

This seems to be fixed now

>Do we want the errors to be printed like they are in the screenshot or after a specific progress bar has run to 100%.
With the current way, a specific progress bar may be printed multiple times with different percentages.

@TwoOfTwelve Did you also change it to this behavior? Or was it the default in your PR anyways or is my dataset just small enough that the errors are printed so late?

PR #1871, opened 2024-07-16
Fix overscroll when clicking match

Kr0nox:
When clicking on a match, it was possible to overscroll the entire page and thus making important information inaccessible.

This PR solves this by:
- Scrolling to the line number and not the text line
- Vertical scrolling is now performed by manually calculating the scroll offset


tsaglam (approved):
Looks good.

PR #1881, opened 2024-07-18
Added pre-commit hook checking spotless

TwoOfTwelve:
Added pre-commit hook checking spotless


Kr0nox:
Works good. Although the package lock differs when i run `npm i`. Did you remove husky manually or using `npm uninstall husky`

gitHooks/installHooks.py
```diff
@@ -0,0 +1,5 @@
+#!/bin/python
+
+import subprocess
+
+subprocess.run(["git", "config", "--local", "core.hooksPath", "gitHooks/hooks"])
```
dfuchss, line 5: You don't need python for this :)

dfuchss (changes requested):

gitHooks/installHooks.py
dfuchss: I assume, you can't :( Create a .bat and .sh file. Otherwise, you require the users to have python installed.

gitHooks/hooks/pre-commit
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
Kr0nox: yes, worked on windows for me

gitHooks/scripts/javaPreCommit
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+if ! command -v <the_command> &> /dev/null #checks if maven is installed
+then
+    exit 0
```
dfuchss, line 5: Probably a warning would be good here :)

Kr0nox:
> @Kr0nox I removed husky through npm uninstall. The changes look identical on my system.

Then I would probably attribute that to a different node version. But everything still works with the current package-lock

gitHooks/installHooks.py
```diff
@@ -0,0 +1,5 @@
+#!/bin/python
+
+import subprocess
```
dfuchss, line 3: I don't get why we need bat,sh **and** python :) ?

gitHooks/scripts/reportViewerPreCommit
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+npx lint-staged
```
dfuchss, line 3: Same as for java. You should check whether npx is installed

dfuchss (changes requested):

gitHooks/installHooks.py
dfuchss: I simply don't get in which cases you dont have a terminal?

gitHooks/scripts/reportViewerPreCommit
dfuchss: Not sure here. I could still change one line without the need for the full npm stack ..

gitHooks/installHooks.py
dfuchss: After talking to @tsaglam : The clean way to go is just to remove all the installHooks scripts and put it into the development wiki :) We would prefer this way :)

dfuchss: Yes :) Otherwise, we have to maintain three files.

README.md
```diff
@@ -182,6 +182,11 @@ try {
 We're happy to incorporate all improvements to JPlag into this codebase. Feel free to fork the project and send pull requests.
 Please consider our [guidelines for contributions](https://github.com/jplag/JPlag/wiki/3.-Contributing-to-JPlag).
 
+### Git hooks
```
dfuchss, line 185: @tsaglam has to say whether https://github.com/jplag/JPlag/wiki/3.-Contributing-to-JPlag is a better place for the information :)

dfuchss (approved):

README.md
tsaglam: We discussed this in the DevMeeting, we concluded the wiki is the right place.

dfuchss (approved):

Kr0nox (approved):

tsaglam (approved):

PR #1904, opened 2024-08-05
Added a warning when a newer version is available in the cli.

TwoOfTwelve:
Added a warning when a newer version is available in the cli.

Accesses the GitHub api and compares the current version against the newest.
If the connection fails, a warning is printed.

### DIFF
--- a/cli/src/main/java/de/jplag/cli/JPlagVersionChecker.java
+++ b/cli/src/main/java/de/jplag/cli/JPlagVersionChecker.java
@@ -26,7 +26,7 @@ public class JPlagVersionChecker {
     private static final Logger logger = LoggerFactory.getLogger(JPlagVersionChecker.class);
     private static final String EXPECTED_VERSION_FORMAT = "v\\d\\.\\d\\.\\d+";
     private static final String WARNING_UNABLE_TO_FETCH = "Unable to fetch version information. New version notification will not work.";
-    private static final String NEWER_VERSION_AVAILABLE = "There is a newer version ({}) available. You can fetch the newest version here: https://github.com/jplag/JPlag/releases.";
+    private static final String NEWER_VERSION_AVAILABLE = "There is a newer version ({}) available. You can download the newest version here: https://github.com/jplag/JPlag/releases";
     private static final String UNEXPECTED_ERROR = "There was an unexpected error, when checking for new versions. Please report this on: https://github.com/jplag/JPlag/issues";
 
     private JPlagVersionChecker() {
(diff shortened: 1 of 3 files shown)


cli/src/main/java/de/jplag/cli/JPlagVersionChecker.java
```diff
@@ -0,0 +1,84 @@
...
+        return parseVersion(apiResult.getJsonObject(0).getString("name"));
+    }
+
+    /**
+     * Parses the version name.
+     * @param versionName The version name. The expected format is: v[major].[minor].[patch]
+     * @return The parsed version
+     */
+    private static Version parseVersion(String versionName) {
+        String withoutPrefix = versionName.substring(1);
```
dfuchss, line 76: Please add a regex here. Maybe in the future there will be some alpha versions. Just to be sure that this call will not cause an exception.

cli/src/main/java/de/jplag/cli/JPlagVersionChecker.java
```diff
@@ -0,0 +1,84 @@
...
+    private static Optional<Version> checkForNewVersion() {
+        try {
+            JsonArray array = fetchApi();
+            Version newest = getNewestVersion(array);
+            Version current = JPlag.JPLAG_VERSION;
+
+            if (newest.compareTo(current) > 0) {
+                return Optional.of(newest);
+            }
+        } catch (IOException | URISyntaxException e) {
```
dfuchss, line 50: This call should never fail. Maybe catch more in this case. You don't want to crash JPlag because of an unknown scenario. E.g., Server returns 502 -> I don't know the exception here .. Or JSON invalid -> Jackson will throw an exception ..

tsaglam: In both cases, no warning or error, just to debug the log. If the user is the a train, he has no connection, then it should not fail as this is technically not a unexpected failure but a scenario that can be expected. Also if the latter case is true, JPlag still works as intended.

cli/src/main/java/de/jplag/cli/JPlagVersionChecker.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Handles the check for newer versions.
+ */
+public class JPlagVersionChecker {
+    private static final String API_URL = "https://api.github.com/repos/jplag/JPlag/releases";
+    private static final Logger logger = LoggerFactory.getLogger(JPlagVersionChecker.class);
+    private static final String EXPECTED_VERSION_FORMAT = "v\\d\\.\\d\\.\\d+";
+    private static final String WARNING_UNABLE_TO_FETCH = "Unable to fetch version information. New version notification will not work.";
+    private static final String NEWER_VERSION_AVAILABLE = "There is a newer version ({}) available. You can fetch the newest version here: https://github.com/jplag/JPlag/releases.";
```
Kr0nox, line 29: Imo `download` would be a more fitting word than `fetch`, but i guess that is just preference

dfuchss, line 29:
```suggestion
    private static final String NEWER_VERSION_AVAILABLE = "There is a newer version ({}) available. You can download the newest version here: https://github.com/jplag/JPlag/releases";
```

PR #1914, opened 2024-08-07
Added structure for token position tests.

TwoOfTwelve:
This adds the structure for token position tests.

The tests are written by adding test sources with a specific syntax, see languages/java/src/test/resources/de/jplag/java/tokenPositions/VarDef_1.java. The actual source is given with each line prefixed by '>'. Lines with '$' as a prefix contain information for one expected token each. The '|' marks the spot where the token should be extracted in the source line above. After that the token type follows and last the length of the token.

### DIFF
--- a/language-testutils/src/test/java/de/jplag/testutils/datacollector/TokenPositionTestData.java
+++ b/language-testutils/src/test/java/de/jplag/testutils/datacollector/TokenPositionTestData.java
@@ -12,12 +12,20 @@ import de.jplag.Token;
 import de.jplag.testutils.TmpFileHolder;
 import de.jplag.util.FileUtils;
 
+/**
+ * Test sources with token information Reads token position test specifications form a file and provides the token
+ * information for tests. The sources cna be used as regular test sources.
+ */
 public class TokenPositionTestData implements TestData {
     private final List<String> sourceLines;
     private final List<TokenData> expectedTokens;
 
     private final String descriptor;
 
+    /**
+     * @param testFile The file containing the test specifications
+     * @throws IOException If the file cannot be read
+     */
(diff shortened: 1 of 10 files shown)


language-testutils/src/test/java/de/jplag/testutils/datacollector/TokenPositionTestData.java
```diff
@@ -0,0 +1,92 @@
...
+        int currentLine = 0;
+
+        for (String sourceLine : testFileLines) {
+            if (sourceLine.charAt(0) == '>') {
+                this.sourceLines.add(sourceLine.substring(1));
+                currentLine++;
+            }
+
+            if (sourceLine.charAt(0) == '$') {
+                int col = sourceLine.indexOf('|');
```
tsaglam, line 47: `column`. Generally, there are a lot of unnecessary abbreviations throughout the code; try to avoid them.

language-testutils/src/test/java/de/jplag/testutils/datacollector/TokenPositionTestData.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Information about a single token
+     * @param typeName The name of the token type
+     * @param line The line the token is in
+     * @param col The column the token is in
+     * @param length The length of the token
+     */
+    public record TokenData(String typeName, int line, int col, int length) {
```
tsaglam, line 90: lineIndex, columnIndex?

tsaglam (changes requested):
I found a few issues, but generally it looks good.

tsaglam (approved):

PR #1935, opened 2024-08-26
Feature/allow single submissions and path names

### LINKED ISSUE #1848: Feature requeset: same name submissions

euberdeveloper:
This pull request add minor updates described in the last comment of #1852. It fixed #1848  and #1849


Kr0nox (changes requested):
As mentioned on the last PR, since you are changing/adding features on how submissions can be given to Jplag, please add an e2e case for each of them.

Here are references on where they are implemented currently:
[develop/report-viewer/tests/e2e/OpenComparisonTest.spec.ts](https://github.com/jplag/JPlag/blob/develop/report-viewer/tests/e2e/OpenComparisonTest.spec.ts?rgh-link-date=2024-07-16T08%3A38%3A00Z)
[develop/.github/workflows/complete-e2e.yml](https://github.com/jplag/JPlag/blob/develop/.github/workflows/complete-e2e.yml?rgh-link-date=2024-07-16T08%3A38%3A00Z)

Please add one for every Problem solved by this PR

euberdeveloper:
Ok, the tests I did so far weren't in the e2e, I'll add some

euberdeveloper:
The functioning of the e2e tests seems to be more complicated than the normal Java tests.
If I've understook correctly, for each featur I should:
1. Add a zip dataset to `.github/workflows/files`
2. During the action, `java -jar jplag.jar ${{ matrix.dataset.folder }} -l ${{ matrix.dataset.language }} -r ${{ matrix.dataset.name }}-report ${{ matrix.dataset.cliArgs }}` will execute it
3. The report result will be added to an artifact
4. I should add [here](https://github.com/jplag/JPlag/blob/develop/report-viewer/tests/e2e/OpenComparisonTest.spec.ts?rgh-link-date=2024-07-16T08%3A38%3A00Z) an object with the expected first two submission names that will be compared
5. The e2e tests of Vue will do the rest

Is it correct?

euberdeveloper:
I have added the e2e test to one of the two features, how can I test the action?

euberdeveloper:
Any hint on how to debug the playwright e2e issue? I am sure the problem is on how the test is written, not on what JPlag does. Unfortunately I don't have any experience with Playwright...

euberdeveloper:
The e2e tests should be fixed

euberdeveloper:
In [this check](https://github.com/jplag/JPlag/actions/runs/10666536761/job/29562226385?pr=1935), the error is `Set the SONAR_TOKEN env variable.`. Is it something that should be done on your side?

Otherwise all checks seem to have passed!

euberdeveloper:
Ok, so should I divide the pull request in two pull requests?

euberdeveloper:
Also for the second feature, we could make it that it stops until the current dir where it is executed JPlag CLI or something similar so that it doesn't give further than needed

tsaglam:
> Ok, so should I divide the pull request in two pull requests?

Yes, the first feature should be its own PR.

> Also for the second feature, we could make it that it stops until the current dir where it is executed JPlag CLI or something similar so that it doesn't give further than needed

I will take that idea into our internal discussions.

euberdeveloper:
Ok, then:
1. I will close this PR and make a new PR only with the accepted feature
2. I will open another PR with the unaccepted one, in which in case some changes could be proposed 
    - e.g. stopping at cwd or something similar when proceeding to the parents
    - e.g. adding a flag "unsafe" or "reveal paths" and only if active the feature will act
    
I think with a specific flag to enable the feature the security issue is solved, because one that wants to use it must explicitly specify that the behaviour is wanted

tsaglam:
Closed, as discussed above.

PR #1955, opened 2024-08-30
Fix exporting of last token in match

Kr0nox:
Currently, the last token in a match gets chosen, based on its start column. Now it takes the last column plus the tokens length


dfuchss:
Could you add a mini-test that ensures that the issue will not occur in the future again :) Or is it complicated to write a test for this ?

tsaglam (approved):

dfuchss (approved):

Issue #1865, opened 2024-07-10
Java Language Frontend: J_CONSTR_BEGIN, J_CONSTR_END and J_VOID never used


euberdeveloper:
It can be seen from the code that these token types for Java are never used:

![image](https://github.com/jplag/JPlag/assets/33126163/54beb248-f8d5-4329-b7a9-14fe8c805d8e)

euberdeveloper:
The same can be said for other tokens

![image](https://github.com/jplag/JPlag/assets/33126163/a9f5ce42-5375-475e-a0cf-792639122175)

tsaglam:
Yes, the unused tokens and the comments after each line are very much legacy. Some of them should not be used during tokenization as they decrease detection quality by enlarging the attack surface. For others, it is unclear if they have an impact at all.

Kr0nox:
@TwoOfTwelve I think you mentioned something about unused tokens in one of our meetings? Is this the same? Did they get deleted now?

TwoOfTwelve:
It is the same. They are not deleted yet. I'll investigate why they are here and open a PR

PR #1960, opened 2024-09-03
Added token position tests for annotations

TwoOfTwelve:
Added some token position tests for annotations. The tests currently fail, because I decided on different lengths than the java module currently reports.

The lengths I assume for these test:

-Annotation: from @ to end of annotation name
-Annotation type begin: from the first modifier to the end of the name
All other tokens lengths are the same as in the current implementation 

There are also some considerations for the module:

Most of the annotation token types seem to be unused. Maybe we should remove them?
I guess ANNO_M was supposed to be annotation members. They are currently extracted as Method tokens. Do we want to keep it like that?
Annotation arguments are currently not extracted at all. Do we want some tokens for that?

### DIFF
--- a/languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java
+++ b/languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java
@@ -147,7 +147,7 @@ final class TokenGeneratingTreeScanner extends TreeScanner<Void, Void> {
         } else if (node.getKind() == Tree.Kind.RECORD) {
             addToken(JavaTokenType.J_RECORD_BEGIN, start, 1, semantics);
         } else if (node.getKind() == Tree.Kind.ANNOTATION_TYPE) {
-            addToken(JavaTokenType.J_ANNO_T_BEGIN, start - 2 /*@ is final modifier for annotations*/, (start - 2) + 11 + nameLength, semantics);
+            addToken(JavaTokenType.J_ANNO_T_BEGIN, start - 2 /* @ is final modifier for annotations */, (start - 2) + 11 + nameLength, semantics);
         } else if (node.getKind() == Tree.Kind.CLASS) {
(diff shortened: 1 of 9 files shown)


languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java
```diff
@@ -146,7 +147,7 @@ public Void visitClass(ClassTree node, Void unused) {
         } else if (node.getKind() == Tree.Kind.RECORD) {
             addToken(JavaTokenType.J_RECORD_BEGIN, start, 1, semantics);
         } else if (node.getKind() == Tree.Kind.ANNOTATION_TYPE) {
-            addToken(JavaTokenType.J_ANNO_T_BEGIN, start, 10, semantics);
+            addToken(JavaTokenType.J_ANNO_T_BEGIN, start - 2 /* @ is final modifier for annotations */, (start - 2) + 11 + nameLength, semantics);
```
uuqjz, line 150: This inline comment look awful between two parameters

uuqjz (changes requested):
Please move the comment

tsaglam (approved):
Looks good.

My initial thoughts on the open points:
> Most of the annotation token types seem to be unused. Maybe we should remove them?

Yes, we can do that, probably in a separate PR to also cover #1865.

> I guess ANNO_M was supposed to be annotation members. They are currently extracted as Method tokens. Do we want to keep it like that?

During the declaration of the annotation members in the annotation definition? There I would stick to method tokens.

> Annotation arguments are currently not extracted at all. Do we want some tokens for that?

We could actually do this. However, ideally, it would only be for mandatory ones. If we do it for optional ones, you can affect the tokens by removing the argument or adding an argument with the same value as the default.

Let me know what you guys think!

tsaglam:
@uuqjz ready from your side?

uuqjz (approved):
LGTM

PR #1975, opened 2024-09-06
Feature/allow single submissions

euberdeveloper:
Please refer the previous pull request: #1935
Regards issue #1849

This pull requests contains only the wanted behaviour and removes the fact that multiple submissions with same name should not be accepted and enlarged in their names with their parents


Kr0nox:
I tested the example you attached for the e2e tests, and it worked fine for me in version 5.1.0. Does the example not work for you in 5.1.0? Or could you provide us an example that does not work in 5.1.0

tsaglam:
> I tested the example you attached for the e2e tests, and it worked fine for me in version 5.1.0. Does the example not work for you in 5.1.0? Or could you provide us an example that does not work in 5.1.0

To add to this, this PR does not change the behavior of JPlag; it just makes some cosmetic changes.

tsaglam:
Closed due to inactivity.

PR #2075, opened 2024-11-20
Added the option to specify a result file with --move VIEW

TwoOfTwelve:
Added the option to specify a result file with `--mode VIEW`

Allows to specify exactly one file with either the positional argument, --new, --old or -r

The documentation has been adapted in the CLI and in the README

### DIFF
--- a/cli/src/main/java/de/jplag/cli/options/CliOptions.java
+++ b/cli/src/main/java/de/jplag/cli/options/CliOptions.java
@@ -22,7 +22,7 @@ import picocli.CommandLine.Parameters;
 public class CliOptions implements Runnable {
     public static final Language defaultLanguage = new JavaLanguage();
 
-    @Parameters(paramLabel = "root-dirs", description = "Root-directory with submissions to check for plagiarism. If mode is set to VIEW, this parameter can be used to specify a file to open. In that case only a single file may be specified.", split = ",")
+    @Parameters(paramLabel = "root-dirs", description = "Root-directory with submissions to check for plagiarism. If mode is set to VIEW, this parameter can be used to specify a report file to open. In that case only a single file may be specified.", split = ",")
     public File[] rootDirectory = new File[0];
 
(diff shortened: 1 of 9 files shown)


cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -53,7 +53,8 @@ public class CliOptions implements Runnable {
             "--result-file"}, description = "Name of the file in which the comparison results will be stored (default: ${DEFAULT-VALUE}). Missing .zip endings will be automatically added.")
     public String resultFile = "results";
 
-    @Option(names = {"-M", "--mode"}, description = "The mode of JPlag. One of: ${COMPLETION-CANDIDATES} (default: ${DEFAULT_VALUE})")
+    @Option(names = {"-M",
+            "--mode"}, description = "The mode of JPlag. If VIEW is chosen, you can specify a result file to display with either the positional argument, '--new', '--old' or '-r'. Make sure to only specify one if you do that. One of: ${COMPLETION-CANDIDATES} (default: ${DEFAULT_VALUE})")
```
tsaglam, line 57: I was wondering whether we should simplify this and not tell the users that there are multiple options. Just document one, but leave the functionality for all others in.

tsaglam (changes requested):
Just a few minor issues (+ Sonar issues).

cli/src/main/java/de/jplag/cli/options/CliOptions.java
tsaglam, line 57:
```suggestion
            "--mode"}, description = "The mode of JPlag. One of: ${COMPLETION-CANDIDATES} (default: ${DEFAULT_VALUE}). If VIEW is chosen, you can optionally specify a result file.")
```

tsaglam: And for subm dir argument add a short sentence: if mode is view, you specify path report here.

PR #2076, opened 2024-11-20
Implemented GreedyStringTiling workaround

TwoOfTwelve:
Implemented a check for the index out of bounds exception, that occurs occasionally.
It prevents the normal exception and throws one with more debug information

### DIFF
--- a/core/src/main/java/de/jplag/GreedyStringTiling.java
+++ b/core/src/main/java/de/jplag/GreedyStringTiling.java
@@ -30,6 +30,8 @@ public class GreedyStringTiling {
     private final Map<Submission, int[]> cachedTokenValueLists = new IdentityHashMap<>();
     private final Map<Submission, SubsequenceHashLookupTable> cachedHashLookupTables = new IdentityHashMap<>();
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/de/jplag/GreedyStringTiling.java
```diff
@@ -30,6 +31,13 @@ public class GreedyStringTiling {
     private final Map<Submission, int[]> cachedTokenValueLists = new IdentityHashMap<>();
     private final Map<Submission, SubsequenceHashLookupTable> cachedHashLookupTables = new IdentityHashMap<>();
 
+    private static final String ERROR_INDEX_OUT_OF_BOUNDS = """
```
Kr0nox, line 34: Elaborate what index and length is printed.
Add information what two submissions are being compared (names)

tsaglam: I would second that.

Kr0nox (approved):

tsaglam (approved):

PR #2086, opened 2024-12-01
Eslint upgrade

Kr0nox:
This PR upgrades eslint to version 9 and moves to the new flat config format.
Due to several dependencies being upgraded in the process new rules emerged. The code was updates to match these new rules.
The vast majority of these changes are switches in the order parmeters are given in markup


Kr0nox:
We first need to discuss the issue.
The problem is that e is not used in this case, but if it is just left, eslint will complain.
voiding it however makes sonarcloud complain.
One solution would be to ignore the eslint rule for that line

tsaglam:
> One solution would be to ignore the eslint rule for that line

It would be fine here, I think.

tsaglam (approved):

PR #2087, opened 2024-12-02
Multi Language Program Support

TwoOfTwelve:
This PR adds the capability to parse multiple languages with JPlag in on run.

A new language module is added, which identifies a module for each file and delegates the parsing to that module. The resulting tokens are concatenated and passed back to JPlag. This does not allow comparing different languages with each other, but it allows for projects that contain multiple languages (multi-language projects).

The languages are discovered on runtime using the same mechanism as the cli. This made it necessary to move the LanguageLoader class from the cli module to the language-api module.
The language modules are identified using the suffixes defined in the individual language modules. This matches the way JPlag already identifies code files in a submission.
The user has to select all language modules that should be used manually using a language specific option.

If multiple language modules are selected for the same suffix, a module is chosen arbitrarily.

Future considerations:
* Change Cli to allow configuration for multiple language-modules. Right now it's not possible to set language specific options for the selected languages. Changing that would be a major change to the cli though.
* Allow implicitly selecting all languages. This would make JPlag easier to use, since a user would not have to select a language module anymore, but makes handing language modules with the same suffix harder
* Find a better way to identify the correct language module. This could be done by adding priorities to language modules, or by adding function to analyze the contents of a file before.

Usage example:
```
jplag multi --languages java,cpp <path to submissions>
```

### DIFF
--- a/language-api/src/main/java/de/jplag/options/LanguageOptions.java
+++ b/language-api/src/main/java/de/jplag/options/LanguageOptions.java
@@ -56,7 +56,7 @@ public abstract class LanguageOptions {
      * @return The new option
      */
     protected <T> LanguageOption<T> createOption(OptionType<T> type, String name, String description) {
-        LanguageOption<T> option = new DefaultLanguageOption<>(type, description, name);
+        LanguageOption<T> option = new DefaultLanguageOption<>(type, name, description);
         this.options.add(option);
         return option;
     }
--- a/language-api/src/main/java/de/jplag/options/LanguageOptions.java
+++ b/language-api/src/main/java/de/jplag/options/LanguageOptions.java
@@ -56,7 +56,7 @@ public abstract class LanguageOptions {
      * @return The new option
      */
     protected <T> LanguageOption<T> createOption(OptionType<T> type, String name, String description) {
-        LanguageOption<T> option = new DefaultLanguageOption<>(type, name, description);
+        LanguageOption<T> option = new DefaultLanguageOption<>(type, description, name);
         this.options.add(option);
         return option;
     }
--- a/languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageOptions.java
+++ b/languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageOptions.java
@@ -11,20 +11,24 @@ import de.jplag.options.OptionType;
 
 public class MultiLanguageOptions extends LanguageOptions {
     private static final String ERROR_LANGUAGE_NOT_FOUND = "The selected language %s could not be found";
-    private static final String ERROR_NOT_ENOUGH_LANGUAGES = "To use multi language specify at least 2 languages";
+    private static final String ERROR_NOT_ENOUGH_LANGUAGES = "To use multi language specify at least 1 language";
 
-    public LanguageOption<String> languageNames = createOption(OptionType.string(), "langs",
+    public LanguageOption<String> languageNames = createOption(OptionType.string(), "languages",
             "The languages that should be used. This is a ',' separated list");
     private List<Language> languages = null;
 
     public List<Language> getLanguages() {
         if (this.languages == null) {
+            if (languageNames.getValue() == null) {
+                throw new IllegalArgumentException(ERROR_NOT_ENOUGH_LANGUAGES);
+            }
+
             List<Language> languages = Arrays.stream(languageNames.getValue().split(","))
                     .map(name -> LanguageLoader.getLanguage(name)
                             .orElseThrow(() -> new IllegalArgumentException(String.format(ERROR_LANGUAGE_NOT_FOUND, name))))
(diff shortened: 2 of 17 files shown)


language-api/src/main/java/de/jplag/options/LanguageOptions.java
```diff
@@ -56,7 +56,7 @@ protected <T> LanguageOption<T> createDefaultOption(OptionType<T> type, String n
      * @return The new option
      */
     protected <T> LanguageOption<T> createOption(OptionType<T> type, String name, String description) {
-        LanguageOption<T> option = new DefaultLanguageOption<>(type, name, description);
+        LanguageOption<T> option = new DefaultLanguageOption<>(type, description, name);
```
uuqjz, line 59: Why did you change the order? The parameters of the function are name and description

languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageOptions.java
```diff
@@ -0,0 +1,40 @@
...
+import de.jplag.options.LanguageOption;
+import de.jplag.options.LanguageOptions;
+import de.jplag.options.OptionType;
+
+public class MultiLanguageOptions extends LanguageOptions {
+    private static final String ERROR_LANGUAGE_NOT_FOUND = "The selected language %s could not be found";
+    private static final String ERROR_NOT_ENOUGH_LANGUAGES = "To use multi language specify at least 1 language";
+
+    public LanguageOption<String> languageNames = createOption(OptionType.string(), "languages",
+            "The languages that should be used. This is a ',' separated list");
```
uuqjz, line 17: Extract as constant please

uuqjz (changes requested):
Looks good. Just found two small things.

uuqjz (approved):
LGTM

tsaglam (commented):
Could you adapt the PR name to be more descriptive, details like the background (BA alex) of the PR should not be in the PR name.

robinmaisch (approved):
Neat!

uuqjz:
@TwoOfTwelve could you adress the Sonor issue please?

languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageParser.java
```diff
@@ -0,0 +1,36 @@
...
+            if (language.isPresent()) {
+                results.addAll(language.get().parse(Set.of(file), normalize));
+            }
+        }
+        return results;
+    }
+
+    private Optional<Language> findLanguageForFile(File file) {
+        return this.languages.stream().filter(language -> Arrays.stream(language.suffixes()).anyMatch(suffix -> file.getName().endsWith(suffix)))
+                .findFirst();
```
tsaglam, line 34: This will currently pick the first language for cases where multiple languages support the same file type. Have you discussed this behavior? This affects the C/C++ modules and also the EMF modules.

tsaglam (changes requested):
For conceptually larger PRs or PRs that add a lot of new functionality the PR description should sufficiently describe the intent of the PR and how it works. Also edge cases like the one mentioned below.

languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageParser.java
tsaglam: (also in future Java vs. Java-CPG)

tsaglam: Ah I understand; I overlooked that the users specify the languages when using the module.
In the long run, I considered making the language module the default, but that would only be user-friendly if users do not need to specify languages. This would mean the multi-language module automatically parses all code that JPlag supports. Then we need prioritization.

With the current solution, we add yet another cli argument, which is less likely to be used by many users. For now, let us leave it as is, but before the release, we need to think about which mode we truly want. If we want more people to try out the language module, we probably need to implement the unparameterized version. However, in all cases, I would not make it the default language straight away.

tsaglam (approved):
Looks good; we should discuss in the future if we prefer specifying languages or if all supported languages should be used by default.

tsaglam:
> Can you look at the sonar issue with the cyclic dependency? I think it's fine in this case.

I briefly looked at the cycles.
The multi-language cycle could be resolved by avoiding the .class comparison and adding a method to the language interface. However, I would not do that, as this could also be resolved when changing the things discussed above in a later PR. So let us leave that cycle as is.

The other one was not immediately clear to me, I will have a look at it later.

tsaglam:
I think the other one is: `Submission --> JPlagOptions --> SimilarityMetric --> JPlagComparison --> Submission`
Which is not mentioned by Sonar. Should be fine as well, I think structurally SimilarityMetric is the culprit and needs a redesign, I think it takes over the responsibility of the JPlagComparison class via the method references. But this has been part of the code for a long time. I would say, fix only the public modifier and we ignore the cycles for now.

PR #2100, opened 2024-12-09
Shared token types

TwoOfTwelve:
Added shared token types and description for how they should be used

### DIFF
--- a/language-api/src/main/java/de/jplag/tokentypes/AnnotationTokenTypes.java
+++ b/language-api/src/main/java/de/jplag/tokentypes/AnnotationTokenTypes.java
@@ -1,0 +1,18 @@
+package de.jplag.tokentypes;
+
+import de.jplag.TokenType;
+
+public enum AnnotationTokenTypes implements TokenType {
+    ANNOTATION("ANNO");
+
+    private String description;
+
+    AnnotationTokenTypes(String description) {
+        this.description = description;
+    }
+
+    @Override
+    public String getDescription() {
+        return this.description;
+    }
+}
--- a/language-api/src/main/java/de/jplag/tokentypes/ImperativeTokenType.java
+++ b/language-api/src/main/java/de/jplag/tokentypes/ImperativeTokenType.java
@@ -1,0 +1,41 @@
+package de.jplag.tokentypes;
+
+import de.jplag.TokenType;
+
+public enum ImperativeTokenType implements TokenType {
+    VARIABLE_DEFINITION("VAR_DEF"),
+    ASSIGNMENT("ASSIGN"),
+
+    STRUCTURE_DEFINITION("STRUCT_DEF"),
+    STRUCTURE_END("STRUCT_END"),
+    STRUCTURE_MEMBER("STRUCT_MEMBER"),
+
+    IF("IF"),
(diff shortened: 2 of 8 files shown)


language-api/src/main/java/de/jplag/tokentypes/ObjectOrientationTokens.java
```diff
@@ -0,0 +1,22 @@
+package de.jplag.tokentypes;
+
+import de.jplag.TokenType;
+
+public enum ObjectOrientationTokens implements TokenType {
+    CONSTRUCTOR("CONST"),
+    CONSTRUCTOR_END("CONT_END"),
+    METHOD("METHOD"),
+    METHOD_END("METHOD_END"),
```
tsaglam, line 9: Are methods really that object-oriented? I would say no, as this token would correspond to any function/routine.

language-api/src/main/java/de/jplag/tokentypes/ObjectOrientationTokens.java
```diff
@@ -0,0 +1,22 @@
+package de.jplag.tokentypes;
+
+import de.jplag.TokenType;
+
+public enum ObjectOrientationTokens implements TokenType {
```
tsaglam, line 5: I would expect the tokens for classes, interfaces, and enums here or maybe in a separate enum for types.

language-api/src/main/java/de/jplag/tokentypes/AnnotationTokenTypes.java
```diff
@@ -0,0 +1,18 @@
+package de.jplag.tokentypes;
+
+import de.jplag.TokenType;
+
+public enum AnnotationTokenTypes implements TokenType {
```
tsaglam, line 5: I would try avoid token types with single tokens.

language-api/src/main/java/de/jplag/tokentypes/ImperativeTokenType.java
```diff
@@ -0,0 +1,41 @@
...
+    IF_END("IF_END"),
+    SWITCH("SWITCH"),
+    SWITCH_END("SWITCH_END"),
+    CASE("CASE"),
+    CASE_END("CASE_END"),
+    LOOP("LOOP"),
+    LOOP_END("LOOP_END"),
+    GOTO("GOTO"), // TODO different context?
+    BREAK("BREAK"),
+    CONTINUE("CONTINUE"),
```
tsaglam, line 24: I would move those into a token type for control flow tokens.

language-api/src/main/java/de/jplag/tokentypes/ImperativeTokenType.java
```diff
@@ -0,0 +1,41 @@
...
+
+    FUNCTION_DEFINITION("FUNCTION_DEF"),
+    FUNCTION_END("FUNCTION_END"),
```
tsaglam, line 27: Why differentiate from methods here? Do we have languages that distinguish methods and functions during tokenization?

tsaglam (commented):
A few thoughts, maybe we need to discuss this in a meeting.

PR #2148, opened 2025-01-17
Reviewable version of 6.0.0 deploy report-viewer

Kr0nox:
> [!warning]
> Do not merge this PR
> It is not meant to be merged into main. It only exists to enable review 

With version 6.0.0 there will no longer be an updated deployed version of the report viewer.
The build version of this branch should be pushed to the gh-pages branch on release of the 6.0.0 version.

The currently deployed version should be modified to notify the user about the newer version and the fact, that the report viewer will no longer be updated

Image


tsaglam (changes requested):
(This is just to prevent merging)

Kr0nox:
Closed now that it was published on pages branch

PR #2158, opened 2025-01-23
Add link for how to use Jplag

Kr0nox:
Adds a link to the wiki page on how to use jplag onto the landing page.

We should discuss if we want to have the link somewhere else too, since when opening a report the user does not see this page anymore.

<details>
  <summary>View</summary>
	

</details>


tsaglam:
I would say, for now, the landing screen is fine for the link.
If the user is already viewing a report, they have already succeeded in using JPlag.

PR #2160, opened 2025-01-24
Give results as local in e2e test

Kr0nox:
Updates the e2e tests to now use provide the files through the request for a local result file instead of the file chooser option


dfuchss (approved):

Kr0nox:
@tsaglam Is there a reason for the PISE tag?

tsaglam:
> @tsaglam Is there a reason for the PISE tag?

Ah, no, I misclicked! :D

PR #2168, opened 2025-01-29
Added automatic mode selection

TwoOfTwelve:
Adds an implicit mode selection. The CLI will now detect the mode based on the input files.

From the documentation of the mode flag:
By default JPlag will automatically select the mode based on your input files. If none are selected the viewer will open on the file select screen. If a single result zip is selected it will be opened in the viewer directly. Otherwise JPlag will run on the submissions in the input files and show the result in the viewer.


cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -115,6 +120,30 @@ public void runViewer(File zipFile) throws IOException {
         JPlagRunner.runInternalServer(zipFile, this.inputHandler.getCliOptions().advanced.port);
     }
 
+    private void selectModeAutomatically() throws IOException, ExitException {
```
Kr0nox, line 123: We should quickly discuss if we want to move calls like `runViewer(runJPlag())` into their own method like `runModeRunAndView()`, to make it clearer which mode got auto selected.

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -115,6 +120,30 @@ public void runViewer(File zipFile) throws IOException {
...
+        }
+
+        this.runViewer(this.runJPlag());
+    }
+
+    private List<File> getAllInputs() {
+        List<File> inputs = new ArrayList<>();
+        inputs.addAll(List.of(this.inputHandler.getCliOptions().newDirectories));
+        inputs.addAll(List.of(this.inputHandler.getCliOptions().oldDirectories));
+        inputs.addAll(List.of(this.inputHandler.getCliOptions().rootDirectory));
```
Kr0nox, line 143: I would say we add the root directory first, just because it seems more logical

cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -53,8 +53,9 @@ public class CliOptions implements Runnable {
             "--result-file"}, description = "Name of the file in which the comparison results will be stored (default: ${DEFAULT-VALUE}). Missing .zip endings will be automatically added.")
     public String resultFile = "results";
 
-    @Option(names = {"-M", "--mode"}, description = "The mode of JPlag. One of: ${COMPLETION-CANDIDATES} (default: ${DEFAULT_VALUE})")
-    public JPlagMode mode = JPlagMode.RUN_AND_VIEW;
+    @Option(names = {"-M",
+            "--mode"}, description = "The mode of JPlag. By default JPlag will automatically select the mode based on your input files. If none are selected the viewer will open on the file select screen. If a single result zip is selected it will be opened in the viewer directly. Otherwise JPlag will run on the submissions in the input files and show the result in the viewer. One of: ${COMPLETION-CANDIDATES} (default: ${DEFAULT_VALUE})")
```
Kr0nox, line 57:
```suggestion
            "--mode"}, description = "The mode of JPlag. By default JPlag will automatically select the mode based on the given input files. If none are given the report viewer will open on the file upload page. If a single result zip is given it will be opened in the report viewer directly. Otherwise JPlag will run on the submissions with the input files and show the result in the report viewer. One of: ${COMPLETION-CANDIDATES} (default: ${DEFAULT_VALUE})")
```

Discuss with @tsaglam before accepting

Kr0nox (changes requested):
Blocker, as a reminder that the above comments are change requests

Kr0nox:
Also update the documentation

https://github.com/jplag/JPlag/blob/10306ccb2ac1363d8ad70ec2d4f22586eb59f2fa/docs/1.-How-to-Use-JPlag.md?plain=1#L120-L130

Inspiration may be found here #2171 or just redirect to that section

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -115,6 +129,30 @@ public void runViewer(File zipFile) throws IOException {
...
 
+    private void selectModeAutomatically() throws IOException, ExitException {
+        List<File> inputs = this.getAllInputs();
+
+        if (inputs.isEmpty()) {
+            this.runViewer(null);
+            return;
+        }
+
+        if (inputs.size() == 1 && inputs.getFirst().getName().endsWith(ZIP_FILE_ENDING)) {
```
Kr0nox, line 140: add a small comment what happens here, why we give an input file to the that is normally a submission to the viewer

docs/1.-How-to-Use-JPlag.md
```diff
@@ -9,59 +9,95 @@ The language can either be set with the -l parameter or as a subcommand. If both
 When using the subcommand, language-specific arguments can be set.
 A list of language-specific options can be obtained by requesting the help page of a subcommand (e.g., "jplag java -h").
 
+To open an existing report run: `java -jar jplag.jar </path/to/report.zip>`
```
Kr0nox, line 12: I would also name the command you have above again here, to have them grouped better
`java -jar jplag.jar path/to/the/submissions`

README.md
Kr0nox, line 1: There are many changes here (and in the docs) due to the formatting of it changing. Is that due to the help text generation of picocli having changed?

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -32,6 +34,8 @@ public final class CLI {
     private static final String OUTPUT_FILE_EXISTS = "The output file (also with suffixes e.g. results(1).zip) already exists. You can use --overwrite to overwrite the file.";
     private static final String OUTPUT_FILE_NOT_WRITABLE = "The output file (%s) cannot be written to.";
 
+    private static final String ZIP_FILE_ENDING = ".zip";
```
tsaglam, line 37: File extension would be a bit less colloquial.

cli/src/main/java/de/jplag/cli/CLI.java
tsaglam: Yes, that is a good idea.

cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -54,8 +54,8 @@ public class CliOptions implements Runnable {
     public String resultFile = "results";
 
     @Option(names = {"-M",
-            "--mode"}, description = "The mode of JPlag. One of: ${COMPLETION-CANDIDATES} (default: ${DEFAULT_VALUE}). If VIEW is chosen, you can optionally specify a path to an existing report.")
-    public JPlagMode mode = JPlagMode.RUN_AND_VIEW;
+            "--mode"}, description = "The mode of JPlag. By default JPlag will automatically select the mode based on the given input files. If none are given the report viewer will open on the file upload page. If a single result zip is given it will be opened in the report viewer directly. Otherwise, JPlag will check the given submissions and show the result in the report viewer. One of: ${COMPLETION-CANDIDATES} (default: ${DEFAULT_VALUE})")
+    public JPlagMode mode = JPlagMode.AUTO;
```
tsaglam, line 58: What was the consensus here from the last meeting? Less details, right? @jplag/hiwi @jplag/maintainer

tsaglam (changes requested):
Looks good! Just a few minor things.

Kr0nox (approved):

docs/1.-How-to-Use-JPlag.md
```diff
@@ -152,13 +133,15 @@ The report will always be zipped unless there is an error during the zipping pro
 Starting with version v6.0.0, the report viewer is bundled with JPlag and will be launched automatically. The `--mode` option controls this behavior.
 By default, JPlag will process the input files and produce a zipped result file. After that, the report viewer is launched (on localhost), and the report will be shown in your browser.
```
Kr0nox, line 134: This is not correct anymore, right? @TwoOfTwelve

Kr0nox:
```suggestion
By default, JPlag tries to guess the best mode. When calling it with submission files, it executes the run_and_view mode. When the positional argument is a zip file including a report, it starts the view mode and opens the given report. When no arguments are given, the report viewer is started in view mode without any given file,
```

We could just copy the report viewer text for this

tsaglam (approved):

PR #2179, opened 2025-02-05
Greedy String Tiling out of bounds

TwoOfTwelve:
Moved the token value mapping outside greedy string tiling to prevent racing conditions

The out of bounds exception that occasionally occurs is presumably caused by calculation the token values in multiple threads and using non-thread-safe methods. This PR solves that problem by calculating the lists before starting greedy string tiling.

Previously there was a method in GreedyStringTiling, that mapped the submissions to token values lazily whenever needed. This is likely the cause of the OutOfBounds exception that occasionally occurs, because GreecyStringTiling runs parallel and the data-structure used isn't thread-safe. To fix this I created a new Class (TokenValueMapper), that creates these mappings in a single thread before calling GreedyStringTiling. The process of creating the token values is unchanged from the original method. When creating the token values the class will also show a progress bar on the cli.

I took some measurements and the total time for JPlag seems to be unaffected

On the progpedia data I got The following measurement:
old: 2.053 s ±  0.014 s
new: 2.080 s ±  0.016 s

Using 100 copies of JPlag core code as input:
old: 13.006 s ±  0.201 s
new: 12.997 s ±  0.150 s

EDIT: _This PR also overhauls the core comparison classes, merging the ComparisonStrategy interface, the AbstractComparisonStrategy class, and the ParallelComparison class into a single one. This separation was historical and serves no longer any purpose. With this, the package structure is also cleaned up._


core/src/main/java/de/jplag/JPlag.java
```diff
@@ -39,8 +39,8 @@ private static Version loadVersion() {
 
     /**
      * Creates and initializes a JPlag instance, parameterized by a set of options.
-     * @deprecated in favor of static {@link #run(JPlagOptions)}.
      * @param options determines the parameterization.
+     * @deprecated in favor of static {@link #run(JPlagOptions)}.
```
Kr0nox, line 43: did this move down due to spotless?

core/src/main/java/de/jplag/JPlag.java
```diff
@@ -49,9 +49,9 @@ public JPlag(JPlagOptions options) {
 
     /**
      * Main procedure, executes the comparison of source code submissions.
-     * @deprecated in favor of static {@link #run(JPlagOptions)}.
      * @return the results of the comparison, specifically the submissions whose similarity exceeds a set threshold.
      * @throws ExitException if JPlag exits preemptively.
+     * @deprecated in favor of static {@link #run(JPlagOptions)}.
```
Kr0nox, line 54: same here

core/src/main/java/de/jplag/TokenValueMapper.java
```diff
@@ -0,0 +1,49 @@
...
+            tokenValues[i] = tokenTypeValues.get(type);
+        }
+        this.tokenValueMap.put(submission, tokenValues);
+    }
+
+    public int[] getTokenValuesFor(Submission submission) {
+        return this.tokenValueMap.get(submission);
+    }
+
+    public static TokenValueMapper generateTokenValueMapper(SubmissionSet submissionSet) {
```
Kr0nox, line 41: Is there a special reason to have this generator method and not use the constrcutor?

core/src/main/java/de/jplag/TokenValueMapper.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.HashMap;
+import java.util.IdentityHashMap;
+import java.util.List;
+import java.util.Map;
+
+import de.jplag.logging.ProgressBarLogger;
+import de.jplag.logging.ProgressBarType;
+
+public class TokenValueMapper {
```
tsaglam, line 11: This class definitely needs exhaustive documentation. First, the complexity in the GST algorithm is already high, making it harder to understand this part of the code base. Second, this class here is currently very generically named, making hard to understand what it does, why it is needed, and what is actually models conceptually.

Please add _helpful_ JDoc comments here, also include your design decision in the PR description.

core/src/main/java/de/jplag/GreedyStringTiling.java
```diff
@@ -219,33 +216,14 @@ private boolean[] calculateInitiallyMarked(Submission submission) {
 
     private SubsequenceHashLookupTable subsequenceHashLookupTableForSubmission(Submission submission, boolean[] marked) {
         return cachedHashLookupTables.computeIfAbsent(submission,
-                (key -> new SubsequenceHashLookupTable(minimumMatchLength, tokenValueListFromSubmission(key), marked)));
-    }
-
-    /**
-     * Converts the tokens of the submission to a list of values.
-     * @param submission The submission from which to convert the tokens.
-     */
```
tsaglam, line 228: Your are removing documention here (by replacing documented code with undocumented code). This is often a giveaway that newly written code should be documented.

core/src/main/java/de/jplag/logging/ProgressBarType.java
```diff
@@ -6,6 +6,7 @@
 public enum ProgressBarType {
     LOADING("Loading Submissions  ", false),
     PARSING("Parsing Submissions  ", false),
+    HASH_CREATION("Preparing Submissions", false),
```
tsaglam, line 9: Enum naming not consistent with CLI message text. Do we really create hashes here?

tsaglam (changes requested):
As the core algorithm is the most long-living and complex part of our code base, it is especially important to make PRs as well as the JDoc for these parts especially descriptive, as the may be referenced frequently in the future.

There are a few issues I see here, mostly regarding high-level design.

Could you also specify explicitly in the PR what changes regarding the point in time of the token mapping. If I understand it correctly, we map all tokens at the beginning now instead of on demand, right?

tsaglam:
Since a restructuring of the core classes was well overdue, I took the opportunity to address a few overarching issues, which also partly addressed issues mentioned in my feedback.

tsaglam (approved):

core/src/main/java/de/jplag/comparison/TokenValueMapper.java
```diff
@@ -0,0 +1,66 @@
...
+ * Maps the tokens in a submission names to integers so they can be used by {@link GreedyStringTiling}. Each token type
+ * will be assigned a unique number. The token lists in that form can be queried by calling
+ * {@link TokenValueMapper#getTokenValuesFor(Submission)}.
+ */
+public class TokenValueMapper {
+    private final Map<TokenType, Integer> tokenTypeValues;
+    private final Map<Submission, int[]> tokenValueMap;
+
+    /**
+     * Crates the mapping for a single submission. This will also show the progress to the user using the
```
robinmaisch, line 26: Documentation is inaccurate, right? Also: typo.

tsaglam: Good catch!

PR #2188, opened 2025-02-14
Enforce mimum merges when using match merging.

### LINKED ISSUE #2077: Subsequence match merging: implement minimum number of merges.
After merging each comparison, we should check the number of neighboring matches merged. When this number is below a threshold, the merged matches should be dropped and the original ones used. This safeguard reduces the impact on pairs or unrelated programs.

tsaglam:
For pairs of unrelated programs, match merging may increase the similarity as it is a heuristic-based approach to revert the effects of obfuscation. To reduce this effect on false positives, this PR introduces a threshold for the minimum number of merges. Below this threshold, the merged matches are discarded, and the original matches are used. Resolves #2077. 

TODO:
- [ ] Adapt tests
- [ ] Introduce CLI parameter
- [ ] tune default value

### DIFF
--- a/core/src/main/java/de/jplag/merging/MatchMerging.java
+++ b/core/src/main/java/de/jplag/merging/MatchMerging.java
@@ -25,6 +25,8 @@ import de.jplag.options.JPlagOptions;
  */
 public class MatchMerging {
     private final JPlagOptions options;
+    private final static int MINIMUM_REQUIRED_MERGES = 3;
+    private int numberOfMerges;
 
     /**
      * Instantiates the match merging algorithm for a comparison result and a set of specific options.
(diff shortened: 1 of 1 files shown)


core/src/test/java/de/jplag/merging/MergingTest.java
```diff
@@ -201,31 +203,27 @@ void testCorrectMerges() {
...
-        }
-
-        List<Match> expectedBefore = new ArrayList<>();
-        expectedBefore.add(new Match(5, 3, 6));
-        expectedBefore.add(new Match(11, 12, 6));
-        expectedBefore.add(new Match(0, 0, 3));
-        expectedBefore.add(new Match(3, 18, 2));
-        expectedBefore.add(new Match(17, 20, 2));
+        List<Match> expectedBefore = List.of( //
+                new Match(5, 3, 6), //
```
uuqjz, line 210: What are the comments for?

uuqjz (approved):
Looks good to me

Kr0nox (changes requested):
I will add the new option to the report viewer. Blocking because of that

Kr0nox (approved):

core/src/test/java/de/jplag/merging/MergingTest.java
tsaglam: They enforce formatting; without them, the Block would be in two lines. What is your preference here?

tsaglam:
Details from the default value tests:

<details><summary>Toggle me!</summary>

image

</details>

6 was chosen as a suitable default value.

PR #2244, opened 2025-03-12
Switch GreedyStringTiling from IdentityHashMap to ConcurrentHashMap

TwoOfTwelve:
Changed to IdentityHashMaps in GreedyStringTiling to ConcurrentHashMaps. This has been done because the comparisons are called parallel and IdentityHashMap is not thread-safe.

I checked to end-to-end tests to make sure the comparison still works as intended.

I measured the runtime impact on the progpedia dataset:
```
Benchmark 1: java -jar new.jar ACCEPTED --mode RUN --overwrite
  Time (mean ± σ):      1.631 s ±  0.029 s    [User: 9.458 s, System: 0.837 s]
  Range (min … max):    1.579 s …  1.709 s    20 runs

Benchmark 2: java -jar old.jar ACCEPTED --mode RUN --overwrite
  Time (mean ± σ):      1.640 s ±  0.026 s    [User: 9.599 s, System: 0.846 s]
  Range (min … max):    1.592 s …  1.681 s    20 runs

Summary
  java -jar new.jar ACCEPTED --mode RUN --overwrite ran
    1.01 ± 0.02 times faster than java -jar old.jar ACCEPTED --mode RUN --overwrite
```

### DIFF
--- a/core/src/main/java/de/jplag/comparison/GreedyStringTiling.java
+++ b/core/src/main/java/de/jplag/comparison/GreedyStringTiling.java
@@ -3,10 +3,10 @@ package de.jplag.comparison;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashSet;
-import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 import de.jplag.JPlagComparison;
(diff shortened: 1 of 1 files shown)


core/src/main/java/de/jplag/comparison/GreedyStringTiling.java
```diff
@@ -26,9 +26,9 @@
 public class GreedyStringTiling {
     private final int minimumMatchLength;
     private final JPlagOptions options;
-    private final Map<Submission, Set<Token>> baseCodeMarkings = new IdentityHashMap<>();
```
dfuchss, line 29: ConcurrentHashMaps work differnetly from identityHashMaps. I'd suggest to use Collections.synchronizedMap() to ensure that behavior stays the same.

dfuchss (changes requested):

core/src/main/java/de/jplag/comparison/GreedyStringTiling.java
tsaglam: We also discussed this PR in the last meeting; the other option would be ConcurrentHashMap + add an equals/hashcode method to Submission. @dfuchss, do you have any opinions architecture-wise? I guess IdentityHashMap will preserve behavior no matter what happens in the future with a potential equals method.

dfuchss: Since there is already a equals / hashcode impl. it can be tricky to ensure the same handling as before.

Another option is not to use a map, but to give the submissions some indices. But I think a synchronized map will be fine

tsaglam:
As far as I can tell, the `baseCodeMarkings` are not accessed concurrently; thus, they do not need synchronization.

tsaglam:
@TwoOfTwelve could you recheck the performance for the current state?

dfuchss (approved):

tsaglam (approved):

PR #2274, opened 2025-03-31
Improve short submission error.

tsaglam:
Improves the following error:

`[ERROR] Submission - Submission first.c contains 8 tokens, which is below the minimum match length 12!`

Before this PR, the number of actual tokens included the end-of-file token.
If a user reads this error and sets the min token match value via `-t` to 8, JPlag will accept the file but will calculate 0% similarity, as there are actually only 7 tokens that can be matched (the comparison algorithm uses the end-of-file tokens as pivot tokens).

With this PR, the end-of-file token is deducted.

### DIFF
--- a/core/src/main/java/de/jplag/Submission.java
+++ b/core/src/main/java/de/jplag/Submission.java
@@ -247,7 +247,7 @@ public class Submission implements Comparable<Submission> {
 
         if (tokenList.size() < minimalTokens) {
             // print the number of tokens without the file-end token to help users choose the right parameters:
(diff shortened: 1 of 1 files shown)


robinmaisch:
Should we account for the case that the submission is a collection of too-short files, so that the total token count is greater than the minimum token match length, but matches are still impossible?

tsaglam:
> Should we account for the case that the submission is a collection of too-short files, so that the total token count (excluding the end-of-file tokens) is greater than the minimum token match length, but matches are still impossible?

Good point, I have not thought about that. We have to think about what we want to happen in this case. Currently, the submission would pass, but it would get 0% with other submissions

tsaglam:
It should be fine for now, and we can always change that behavior if we want to.

PR #2278, opened 2025-04-02
Improve the JPlag API

tsaglam:
improves several aspects of the JPlag API:

- Remove redundant code in the similarity calculation (and add a test case to ensure correctness)
- Correct naming of `Submission::hasBaseCodeMatches` (only checks if there is a basecode comparison, which does not imply that there are matches)
- Make the setter for clustering results package private
- Specify submission naming scheme
- Improve documentation of the similarity scores, JPlagComparison, JPlagResult and Match

### DIFF
--- a/core/src/main/java/de/jplag/JPlagResult.java
+++ b/core/src/main/java/de/jplag/JPlagResult.java
@@ -10,7 +10,15 @@ import de.jplag.options.JPlagOptions;
 import de.jplag.options.SimilarityMetric;
 
 /**
- * Encapsulates the results of a comparison of a set of source code submissions.
+ * Encapsulates the results of a pairwise comparison of program structure among a set of source code submissions.
+ * <p>
+ * Provides access to:
+ * <ul>
+ * <li>Pairwise comparison results sorted by similarity</li>
+ * <li>Similarity distribution data</li>
+ * <li>Execution duration and configuration options</li>
+ * <li>Optional clustering results over submissions</li>
+ * </ul>
  */
 public class JPlagResult {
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/de/jplag/JPlagResult.java
```diff
@@ -45,7 +53,12 @@ public void dropComparisons(int limit) {
         this.comparisons = this.getComparisons(limit);
     }
 
-    public void setClusteringResult(List<ClusteringResult<Submission>> clustering) {
+    /**
+     * Sets the clustering results for the current set of submissions. This can be used to attach the output of one or more
+     * clustering algorithms to this result object.
+     * @param clustering is the list of clustering results.
+     */
+    /* package-private */ void setClusteringResult(List<ClusteringResult<Submission>> clustering) {
```
uuqjz, line 61: Clean up please

uuqjz (changes requested):

uuqjz (approved):
Looks good to me then

core/src/main/java/de/jplag/JPlagResult.java
```diff
@@ -10,7 +10,15 @@
...
 
 /**
- * Encapsulates the results of a comparison of a set of source code submissions.
+ * Encapsulates the results of a pairwise comparison of program structure among a set of source code submissions.
+ * <p>
+ * Provides access to:
+ * <ul>
+ * <li>Pairwise comparison results sorted by similarity</li>
+ * <li>Similarity distribution data</li>
+ * <li>Execution duration and configuration options</li>
```
robinmaisch, line 19: I remember a case where we said we would rather not have HTML in the documentation. I think it is helpful here, though!

tsaglam: Good point! I will restructure it to omit the HTML-style syntax.

PR #2289, opened 2025-04-06
Move sorting to table headers

Kr0nox:
Removes the sorting bar of the table
Sorting is now performed by clicking the table headers.
Sort icons indicate the possible columns and the direction
grafik


tsaglam (approved):
Looks good. I tested it, works well in both the overview and cluster view.

PR #2290, opened 2025-04-06
Add expand all option for files

Kr0nox:
Files are expanded by default
Adds an option to expand all files
grafik


tsaglam (commented):
Could this be the same button that just changes the icon and text?
While this does not allow expanding all, collapsing a few, and then reexpanding all, I think it might still be sufficient and looks cleaner.

PR #2293, opened 2025-04-08
Change report file ending to .jplag

Kr0nox:
The cli now produces `.jplag` files

To ensure some backwards compatibility:
When starting the cli with `jplag.jar SOURCE -r results.zip` it will produce a `results.jplag` file
When starting the cli with `jplag.jar results.zip` it still serves it as `results.jplag` to the report viewer


tsaglam (approved):
The code looks good. I tested it for different use cases, and it behaves as expected. I noticed that the check if a file ending is part of the `-r` input is case sensitive, but that was already the previous behavior for zip files.

Kr0nox:
Should I change it to be case insensitive or delay that for a future pr?

tsaglam:
Good question. We could also say we leave it as is, users will notice anyway, as their file is then named `name.JPLAG.jplag`. No strong opinion here.

PR #2303, opened 2025-04-10
Fix Java module token tracing bug for classes without modifers

### LINKED ISSUE #2292: Class Token not extracted at correct position when no modifiers (java)
When a java class (or enum, interface, record) has no modifiers, e.g.:
```Java
class A {}
```
it is extracted position is `Line 1, Column 1`.

I believe the cause of this is this line:
https://github.com/jplag/JPlag/blob/998557e9da212a4e1c3311e377f695bc895da9f0/languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java#L139
As there are not modifiers to take for the start position.
It could be enough to take the start position of the node when there are no modifiers.

tsaglam:
Fixes #2292: Previously, the modifiers were used to calculate which line a class starts in. This leads to a bug when a class has no modifiers. This is now fixed by using the class position itself when no modifiers are present.

This PR also adds corresponding test cases and a JUnit assumption for preconditions in the test framework.


Kr0nox (commented):
I think we should add a token position test (https://github.com/jplag/JPlag/pull/1914/files#diff-0510ecf52be2ae53bb15c2fbfb5e09afbe3d40097f4dfa517e395b40ffa65f7c) for the class begin token, to make sure the bug does not get reimplemented

Kr0nox (approved):

Issue #2304, opened 2025-04-10
Open tasks regarding the multi-language module


tsaglam:
The following usability improvements are planned:

- [x] When using the multi-language module, a warning should be given that this does not involve cross-language checking
- [x] The language interface should be extended:
	- [x] method `supportsMultiLanguage()` with default: true
	- [x] method `hasPriority()` with default: false
	- [x] set `supportsMultiLanguage()` false for the text and EMF-model module (but true for emf)
	- [x] priority for C++ and typescript
- [x] The multi-language module should delegate to all languages that support it
- [x] When checking which modules support a file of a certain type, priority shall be given according to `hasPriority()`
- [x] If multiple modules that support a single file type have priority, produce an IlligealStateException
- [x] Same if multiple exist but none have priority
- [x] Add multi-language to readme table
- [x] Add dedicated readme for multi-language module
- [x] Add test checking that there are no duplicate file endings in different language modules

EDIT:

- [x] Check report viewer compatibility.

PR #2315, opened 2025-04-17
multi language open tasks

TwoOfTwelve:
Fixes the open tasks from #2304, which improve the usability of the multi-language module.

### DIFF
--- a/cli/src/test/java/de/jplag/cli/FileExtensionConflictsTest.java
+++ b/cli/src/test/java/de/jplag/cli/FileExtensionConflictsTest.java
@@ -7,7 +7,7 @@ import org.junit.jupiter.api.Test;
 import de.jplag.multilang.MultiLanguageOptions;
 import de.jplag.multilang.MultiLanguageParser;
 
-public class FileExtensionConflictsTest {
+class FileExtensionConflictsTest {
     @Test
     void testNoConflictsBetweenFileExtensions() {
         assertDoesNotThrow(() -> new MultiLanguageParser(new MultiLanguageOptions()),


languages/multi-language/README.md
```diff
@@ -0,0 +1,7 @@
+# Multi-Language module
+
+This module allows parsing multiple languages in one run of JPlag. The parsing will be delegated to a different language-module per file.
+
+This does not allow comparing code from different languages, but it can be used to compare multiple languages in the same run.
```
robinmaisch, line 5: Let's be even more precise here to avoid confusion. I suggest this:
"This does not entail comparing implementations of the same functionality in different languages to each other, but allows comparing submissions that use more than one language."

languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageParser.java
```diff
@@ -7,18 +7,28 @@
...
     public MultiLanguageParser(MultiLanguageOptions options) {
         this.languages = options.getLanguages();
+        hasPrintedWarning = false;
     }
 
     public List<Token> parseFiles(Set<File> files, boolean normalize) throws ParsingException {
+        this.printWarning();
         List<Token> results = new ArrayList<>();
         for (File file : files) {
             Optional<Language> language = findLanguageForFile(file);
```
robinmaisch, line 34: We discussed that we can find out at compile time whether the mapping of file extensions to language modules is well-defined. I would argue then that we should set up an actual look up table instead of repeating the same steps for each individual file.

robinmaisch, line 34: We discussed that we can find out at compile time whether the mapping of file extensions to language modules is well-defined. I would argue then that we should set up an actual look up table instead of repeating the same steps for each individual file.

languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageParser.java
```diff
@@ -29,8 +39,45 @@ public List<Token> parseFiles(Set<File> files, boolean normalize) throws Parsing
...
 
-    private Optional<Language> findLanguageForFile(File file) {
-        return this.languages.stream().filter(language -> Arrays.stream(language.suffixes()).anyMatch(suffix -> file.getName().endsWith(suffix)))
-                .findFirst();
+    private Optional<Language> findLanguageForFile(File file) throws ParsingException {
+        List<Language> normalLanguages = new ArrayList<>();
+        List<Language> priorityLanguages = new ArrayList<>();
+
+        for (Language language : this.languages) {
+            if (Arrays.stream(language.suffixes()).anyMatch(it -> file.getName().endsWith(it))) {
```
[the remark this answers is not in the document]
tsaglam: I created one, see #2320.

languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageParser.java
dfuchss: Yes, I meant it as general information . I would not implement it in this PR :)

robinmaisch: @TwoOfTwelve Sorry for the unclear positioning of my comment. I was not referencing the warning, but the call to `findLanguageForFile(file)`, which repeats the same tedious steps for each file. I propose that, instead, a `Map<String, Language>` of each file extension to the respective language module is set up during initialization, so that this "number of language modules" x "number of files" effort is reduced to a simple lookup of the file extension. The check for ambiguous mapping can also be done there one-time.

tsaglam:
@TwoOfTwelve let us know when the remaining points of #2304 are implemented, then I will review. Also, take a look at the sonar issues.

language-api/src/main/java/de/jplag/Language.java
```diff
@@ -123,4 +123,18 @@ default boolean supportsNormalization() {
...
+
+    /**
+     * @return True, if the language module can be used by the multi-language module
+     */
+    default boolean supportsMultilanguage() {
+        return true;
+    }
+
+    /**
+     * @return True, if the language module should be used in case of ambiguities
```
tsaglam, line 135: Should be clearer, e.g. 
```suggestion
     * @return True if the language module should be prioritized in the multi-language case, if multiple language modules support the same file extension.
```

languages/cpp/src/main/java/de/jplag/cpp/CPPLanguage.java
```diff
@@ -50,4 +50,9 @@ public boolean supportsNormalization() {
     public List<Token> parse(Set<File> files, boolean normalize) throws ParsingException {
         return new CPPParserAdapter().parse(files);
     }
+
+    @Override
+    public boolean hasPriority() {
+        return true;
```
tsaglam, line 56:
```suggestion
        return true; // Priority over the C language module.
```
Let us document the interdependencies.

languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageOptions.java
```diff
@@ -20,14 +20,14 @@ public class MultiLanguageOptions extends LanguageOptions {
     public List<Language> getLanguages() {
         if (this.languages == null) {
             if (languageNames.getValue() == null) {
-                throw new IllegalArgumentException(ERROR_NOT_ENOUGH_LANGUAGES);
+                this.languages = LanguageLoader.getAllAvailableLanguages().values().stream().filter(Language::supportsMultilanguage).toList();
+            } else {
+                this.languages = Arrays.stream(languageNames.getValue().split(","))
```
tsaglam, line 25: Use a constant for the separator.

languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageParser.java
```diff
@@ -29,8 +43,46 @@ public List<Token> parseFiles(Set<File> files, boolean normalize) throws Parsing
         return results;
     }
 
+    private void registerLanguageMaps(List<Language> languages) {
```
tsaglam, line 46: High complexity, as [marked by sonar](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=2315&id=jplag_JPlag&open=AZbN4ypZmLwy2Ak0-sjM). Can you refactor this/split it up to improve the structure?

languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageParser.java
```diff
@@ -29,8 +43,46 @@ public List<Token> parseFiles(Set<File> files, boolean normalize) throws Parsing
...
+        }
+
+        return Optional.empty();
+    }
+
+    private static void printWarning() {
+        if (!hasPrintedWarning) {
+            hasPrintedWarning = true;
+            LOG.warn(WARNING);
+        }
```
tsaglam, line 86: Ideally, we would like to avoid keeping static state here. We could move the warning to the `MultiLanguage.parse()` call, which is only called once.

tsaglam (changes requested):
Just a few minor things here.

tsaglam:
@TwoOfTwelve, regarding the test case, a good spot would be the CLI module, as we already include all languages there.
It is probably enough to instantiate all languages and get all the file extensions, and if there are duplicates, check that exactly one module that supports the extensions has priority. Else, fail.

languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageParser.java
```diff
@@ -2,23 +2,37 @@
...
+    private final Map<String, Language> languageMap;
 
     public MultiLanguageParser(MultiLanguageOptions options) {
-        this.languages = options.getLanguages();
+        this.languageMapPriority = new HashMap<>();
+        this.languageMap = new HashMap<>();
+        this.registerLanguageMaps(options.getLanguages());
     }
 
     public List<Token> parseFiles(Set<File> files, boolean normalize) throws ParsingException {
```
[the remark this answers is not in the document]
tsaglam: It makes sense, so we leave it as is.

cli/src/test/java/de/jplag/cli/LanguageSuffixConflictsTest.java
```diff
@@ -0,0 +1,14 @@
...
+
+import de.jplag.multilang.MultiLanguageOptions;
+import de.jplag.multilang.MultiLanguageParser;
+
+public class LanguageSuffixConflictsTest {
+    @Test
+    void testNoConflictsBetweenLanguageSuffixes() {
+        new MultiLanguageParser(new MultiLanguageOptions()); // triggers internal check for conflicts automatically.
+        // For conflicting language, see error message
+    }
```
tsaglam, line 13: Use `assertDoesNotThrow(() -> new MultiLanguageParser(new MultiLanguageOptions()), message)` here to make it a proper test case. For the message, write a descriptive message that someone in 10 years understands and has no clue about the underlying assumptions of the multi-language module. For example, along the lines of "There is conflict regarding file extensions between two or more language modules. Ensure exactly one module has priority."

languages/javascript/src/main/java/de/jplag/javascript/JavaScriptLanguage.java
```diff
@@ -25,4 +25,9 @@ public String[] suffixes() {
     public String getName() {
         return "JavaScript";
     }
+
+    @Override
+    public boolean hasPriority() {
+        return false;
```
tsaglam, line 31:
```suggestion
        return false; // Since this module extends Typescript, it has no priority.
```

languages/multi-language/README.md
```diff
@@ -0,0 +1,7 @@
+# Multi-Language module
+
+This module allows parsing multiple languages in one run of JPlag. The parsing will be delegated to a different language-module per file.
+
+This does not entail comparing implementations of the same functionality in different languages to each other, but allows comparing submissions that use more than one language each.
+
+By default, all supported languages (except text) are parsed. This can be changed by specifying the language modules to use with the --languages parameter: `java -jar jplag.jar <root folder> multi --languages <firstLanguage>,<secondLanguage>,...`
```
tsaglam, line 7:
```suggestion
This module allows submissions to be compared with files in multiple programming languages. The parsing will be delegated to a different language module per file extension.

This does not entail comparing implementations of the same functionality in different languages to each other, but allows comparing submissions that use more than one language each.

By default, all supported languages (except text) are parsed. This can be changed by specifying the language modules to use with the `--languages` parameter: `java -jar jplag.jar <root folder> multi --languages <firstLanguage>,<secondLanguage>,...`
```

tsaglam (changes requested):

languages/javascript/src/main/java/de/jplag/javascript/JavaScriptLanguage.java
dfuchss: tbh, I do not 100% understand the priority. What would happen if I'd have three languages; where 2 have priority?

cli/src/test/java/de/jplag/cli/LanguageSuffixConflictsTest.java
tsaglam: As discussed in the meeting, it is enough, just use my wording in the exception.

languages/multi-language/src/main/java/de/jplag/multilang/MultiLanguageParser.java
```diff
@@ -29,8 +36,42 @@ public List<Token> parseFiles(Set<File> files, boolean normalize) throws Parsing
...
+        if (language.hasPriority()) {
+            registerLanguageSuffixTo(this.languageMapPriority, language, suffix, true);
+        } else {
+            registerLanguageSuffixTo(this.languageMap, language, suffix, !this.languageMapPriority.containsKey(suffix));
+        }
+    }
+
+    private void registerLanguageSuffixTo(Map<String, Language> map, Language language, String suffix, boolean additionalConditionForException) {
+        if (map.containsKey(suffix) && additionalConditionForException) {
+            throw new IllegalStateException(String.format(ERROR_MULTIPLE_LANGUAGES, suffix, language.getName(), map.get(suffix).getName()));
```
robinmaisch, line 58: The condition "more than one non-prioritized language module but no language module with priority" is not necessarily a problem, as a language module with priority might still be found afterwards. Put like this, depending on the order in which the language modules are loaded, an exception might be thrown or not.

robinmaisch: Concept of alternative implementation:
 - only one `Map<String, Language> languageMap` + one `Set<String> needsPriority`
 - for each `Language language` for each `String fileExtension`:
 - if `!languageMap.containsKey(fileExtension))`: set language for that fileExtension
 - else: check whether `language` and the existing language for `fileExtension` have priority (`newPriority`, `oldPriority`)
 - `newPriority && !oldPriority`: overwrite
 - `newPriority && oldPriority`: throw exception
 - `!newPriority && oldPriority`: do nothing
 - `!newPriority && !oldPriority`: `needsPriority.add(fileExtension)`

In the end, assert that for all file extensions in `needPriority`, a language module with priority has been found, or else throw an exception.

robinmaisch (changes requested):

robinmaisch (approved):

PR #2316, opened 2025-04-18
Restructure report

Kr0nox:
This PR restructures the report file
- comparison files are in their own folder
- split overview into
  - RunInformation (execution time, ...)
  - Clusters
  - Top comparisons
  - Distributions
  - Cli Options
- Make all properties in the report camel case aligning with the JPlag core 
  
Added a central data getter to extract information in the correct order and check the version


tsaglam (approved):
Looks good. I also sanity checked the report viewer UI with a larger dataset. Only the conflict is left to do.

PR #2318, opened 2025-04-20
Warning when submission parsing failed

Kr0nox:
When the report contains failed submissions, it displays a visible message in the top of the overview

Also adds the names of failed submissions to the information view

grafik
grafik
grafik
grafik


report-viewer/src/model/factories/OverviewFactory.ts
```diff
@@ -34,6 +34,24 @@ export class OverviewFactory extends BaseFactory {
       }
     }
 
+    // send a warning the first time the report is opened with the failed submissions
+    const failedSubmissionNames = json.failed_submission_names as string[]
+    if (failedSubmissionNames.length > 0) {
+      const hasSendWarning = sessionStorage.getItem('failedSubmissionWarning')
```
dfuchss, line 40: What happens if I load another report with failed parsing ? Or is this not possible anymore ? Does this reset the session store ?

Kr0nox: Yes that could be an issue. 
We could always reset the token when opening the landing page, although that might not cover some edge cases.
Alternativly we could store some (nearly) unique identifier for a report/failed submission list to store what the last report we sent the warning for is. 
What do you think is the best way? @jplag/maintainer

tsaglam: Do I understand it correctly that this scenario would always arise when opening more than one report at the same time, where at least one submission failed? Then we definitely need a solution here. I'm not sure what the drawbacks of both approaches here are, so hard to judge from my end.

sebinside: We should discuss in the next meeting(s) whether this is a one-time warning needed and what is the best way to inform users that there have been problems without making the warning a) too easy to overlook and b) too annoying. I would guess there is still more to experiment with than we can achieve until the next release. This feature is too important to rush a change that makes everything worse.

Kr0nox: In the meeting we concluded that we want to omit showing an alert and rather have a very visible entry in the top container of the overview, next to the "JPlag Report" Text, stating that comparisons were missing

tsaglam (approved):
Looks great! I proposed a follow-up idea in #2382. But I would do that in a separate PR.

report-viewer/src/views/OverviewView.vue
```diff
@@ -6,7 +6,22 @@
     class="grid grid-cols-1 grid-rows-[auto_800px_90vh] gap-5 md:grid-cols-2 md:grid-rows-[auto_1fr] md:overflow-hidden print:grid-cols-1 print:grid-rows-[auto_auto]"
   >
     <Container class="col-start-1 row-start-1 md:col-end-3 md:row-end-2">
-      <h2>JPlag Report</h2>
+      <div class="flex flex-col gap-x-5 md:flex-row md:items-center">
+        <h2>JPlag Report</h2>
+        <ToolTipComponent v-if="overview.failedSubmissionNames.length > 0" direction="bottom">
+          <template #default>
+            <p class="text-error font-bold">
+              {{ overview.failedSubmissionNames.length }} submissions could be parsed. They will not
```
dfuchss, line 14:
```suggestion
              {{ overview.failedSubmissionNames.length }} submissions couldn't be parsed. They will not
```

dfuchss (changes requested):
Hope that I don't oversee a thing here :D

report-viewer/src/views/OverviewView.vue
```diff
@@ -123,6 +138,22 @@ const submissionPathValue = computed(() =>
...
+  let i = 0
+  while (i < props.overview.failedSubmissionNames.length && names.length < 600) {
+    names += props.overview.failedSubmissionNames[i]
+    i++
+    if (i < props.overview.failedSubmissionNames.length) {
+      names += ', '
+    }
+  }
+  if (i < props.overview.failedSubmissionNames.length) {
+    names += '... (click More to see all names)'
```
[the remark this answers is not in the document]
tsaglam: Good point, maybe even more in italic.

dfuchss:
Sorry, I'll test something here :D

report-viewer/src/views/OverviewView.vue
```diff
@@ -123,6 +138,22 @@ const submissionPathValue = computed(() =>
     : props.overview.submissionFolderPath[0]
 )
 
+const failedNamesToolTip = computed(() => {
+  let names = ''
+  let i = 0
+  while (i < props.overview.failedSubmissionNames.length && names.length < 600) {
```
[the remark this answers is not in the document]
Kr0nox: Changed it to use 20 submissions. This could lead to very inconsistent lengths of the tooltip with different name lengths tho

report-viewer/src/views/InformationView.vue
```diff
@@ -124,6 +124,9 @@
         <TextInformation label="Missing Comparisons" class="pb-1">{{
           overview.missingComparisons
         }}</TextInformation>
+        <TextInformation label="Failed Submissions" class="pb-1">{{
+          overview.failedSubmissionNames.join(', ')
+        }}</TextInformation>
```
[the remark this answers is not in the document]
Kr0nox: This is actually a valid concern and we should discuss how we want to handle this.
1) Leave as it is and have the test behind `Failed Submissions: ` empty
2) Write a text like `Failed Submissions: No failed submissions` (what should be the excact content?)
3) Not show this at all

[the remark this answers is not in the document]
tsaglam: I would vote for 1. or 2.

dfuchss (approved):

robinmaisch (approved):

PR #2331, opened 2025-04-24
Fix file ending of result file in dev demo workflow.

tsaglam:
TODO


Kr0nox:
@tsaglam I also found a test file were the endings were not updated and the file format for the report viewer was not set correctly (even though it should work the way it is currently). Should I append that to this PR?

tsaglam:
Which file was it?

Kr0nox:
The ModeTest in the cli tests

uuqjz (approved):
LGTM

tsaglam:
Let us do a separate PR for this one.

PR #2333, opened 2025-04-24
Set JPlag revision to v6.1.0

robinmaisch:
For the upcoming release, set JPlag revision to v6.1.0

### DIFF
--- a/report-viewer/src/version.json
+++ b/report-viewer/src/version.json
@@ -1,7 +1,7 @@
 {
   "report_viewer_version": {
-    "major": 6,
-    "minor": 1,
+    "major": 0,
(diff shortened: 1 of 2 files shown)


report-viewer/src/version.json
```diff
@@ -1,7 +1,7 @@
 {
   "report_viewer_version": {
```
Kr0nox, line 2: I believe we do not set this version, as the release action does that automatically

Kr0nox (changes requested):

tsaglam (approved):

tsaglam:
Note to future us, we only set the pom.xml revision nothing else.

PR #2339, opened 2025-04-28
Improve code quality and fix minor issues

tsaglam:
This PR fixes various issues raised by Sonar:

- Visibility issues
- Performance-related issues (logging, initialization of data structures)
- Naming issues
- Syntactic sugar

### DIFF
--- a/core/src/main/java/de/jplag/SubmissionSetBuilder.java
+++ b/core/src/main/java/de/jplag/SubmissionSetBuilder.java
@@ -111,7 +111,7 @@ public class SubmissionSetBuilder {
             throw new RootDirectoryException("No root directories specified with submissions to check for plagiarism!");
         }
 
-        Set<File> canonicalRootDirectories = new HashSet<>(rootDirectoryNames.size());
+        Set<File> canonicalRootDirectories = HashSet.newHashSet(rootDirectoryNames.size());
         for (final File rootDirectory : rootDirectoryNames) {
             if (!rootDirectory.exists()) {
                 throw new RootDirectoryException(String.format("Root directory \"%s\" does not exist!", rootDirectory));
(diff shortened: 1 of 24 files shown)


core/src/main/java/de/jplag/SubmissionSetBuilder.java
```diff
@@ -111,7 +111,7 @@ private Set<File> verifyRootDirectories(Set<File> rootDirectoryNames, boolean ar
             throw new RootDirectoryException("No root directories specified with submissions to check for plagiarism!");
         }
 
-        Set<File> canonicalRootDirectories = new HashSet<>(rootDirectoryNames.size());
+        Set<File> canonicalRootDirectories = HashSet.newHashSet(rootDirectoryNames.size());
```
uuqjz, line 114: How is this an improvement?

core/src/main/java/de/jplag/clustering/IntegerMapping.java
```diff
@@ -15,7 +15,7 @@ public class IntegerMapping<T> {
     private int size = 0;
 
     public IntegerMapping(int initialCapacity) {
-        mapping = new HashMap<>(initialCapacity);
+        mapping = HashMap.newHashMap(initialCapacity);
```
uuqjz, line 18: Same here

core/src/main/java/de/jplag/SubmissionSetBuilder.java
tsaglam: Initializing hash sets with a custom size does not mean they are immediately initialized at that size. Rather, they are resized later depending on the load. `HashSet.newHashSet` ensures the size is always as specified and the resizing does not occur at a later time. See also: https://sonarcloud.io/organizations/jplag/rules?open=java%3AS6485&rule_key=java%3AS6485.

uuqjz (approved):
LGTM

core/src/main/java/de/jplag/clustering/IntegerMapping.java
tsaglam: See above. However, it could really affect performance here.

PR #2350, opened 2025-04-30
Fix typos and extend some documentation

robinmaisch:
This PR fixes some typos, primarily in the JavaDoc documentation, as well as in a class name, and adds some detail here and there.
Naturally, this covers only a small part of the project, and other parts need to be revised in the future. Still, every step helps!

### DIFF
--- a/core/src/main/java/de/jplag/SubmissionFileData.java
+++ b/core/src/main/java/de/jplag/SubmissionFileData.java
@@ -3,11 +3,10 @@ package de.jplag;
 import java.io.File;
 
 /**
- * Contains the information about a single file in a submission. For single file submissions the submission file is the
- * same as the root.
- * @param submissionFile The file, that is part of a submission
+ * Captures information about an individual submission.
+ * @param submissionFile The file that is part of a submission
  * @param root The root of the submission
- * @param isNew Indicates weather this follows the new or the old syntax
+ * @param isNew Indicates whether this follows the new or the old syntax
  */
 public record SubmissionFileData(File submissionFile, File root, boolean isNew) {
 }
--- a/core/src/main/java/de/jplag/SubmissionFileData.java
+++ b/core/src/main/java/de/jplag/SubmissionFileData.java
@@ -3,10 +3,11 @@ package de.jplag;
 import java.io.File;
 
 /**
- * Captures information about an individual submission.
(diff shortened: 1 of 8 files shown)


core/src/main/java/de/jplag/SubmissionFileData.java
```diff
@@ -3,11 +3,10 @@
 import java.io.File;
 
 /**
- * Contains the information about a single file in a submission. For single file submissions the submission file is the
- * same as the root.
```
robinmaisch, line 7: Note: I was unable to reproduce the behavior previously described here. It seems that the object `root` is just the same for all submissions, regardless of whether they are single-file submissions or submission directories.

tsaglam: I think the comment means that both fields of type `File` are identical if it is a single file submission.

E.g. for `folder/file.java`:

- submissionFile = `file.java`
- root = `folder`

And for `file.java`

- submissionFile = `file.java`
- root  = `file.java`

However, I think the sentences here in the documentation are not very understandable! :D

robinmaisch: I agree that this is what the comment means, but the code does not seem to do that. Instead, the `root` attribute always points to the directory where the submission (and possibly others) are located, which was given in the CLI. This kind of root cannot be a proper file, the CLI will not allow it.
That's the notion of a root of the CLI.

Other times, like in the `Submission` class, the `submissionRootFile` actually refers to the top-level `File` of the individual submission, which may be a directory or a proper file.
That's the notion of a root of a Submission.

We might want to clear that up a bit.

tsaglam: Oh, so this is not the root file of the submission, but rather the root directory specified in the CLI? This is indeed highly confusing!

tsaglam: I tried to improve it in https://github.com/jplag/JPlag/pull/2350/commits/63148bbda38d136226f8e6b89fdc4323d5eb1a32, do you think the updated version is clear enough?

robinmaisch: I would think so, thanks!

tsaglam (approved):

PR #2362, opened 2025-05-07
Add Q&A section to the Wiki

tsaglam:
Introduces a Q&A section to the wiki for the most important user questions. It is intentionally kept short.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,11 @@
 [![Java Version](https://img.shields.io/badge/java-SE%2021-yellowgreen)](#download-and-installation)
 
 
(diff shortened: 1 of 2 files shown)


tsaglam:
@jplag/maintainer @jplag/devteam let me know if you think an essential question is missing.

tsaglam (commented):
Feedback from our meeting:
Q: To obfuscation resilience?
Q: Who is behind JPlag?
Q: How to archive reports?

uuqjz:
For 7: Maybe state that both "techniques" can be combined. I read it as either or

tsaglam:
> For 7: Maybe state that both "techniques" can be combined. I read it as either or

Good point. I included it.

uuqjz (approved):

Issue #2366, opened 2025-05-09
Multiple SLF4J bindings


robinmaisch:
## Observation ##

During testing or building JPlag, SLF4J brings up a warning indicating that multiple bindings for `SLF4JServiceProvider`s have been found.

![Image](https://github.com/user-attachments/assets/0a0f3ca8-94fe-4615-8b14-f5f42a83ad42)

Even though this does not indicate any failure or fatal error, the warning can be avoided.

## Origin of the warning ##

The warning occurs when more than one `SLF4JServiceProvider` is found in the classpath. In our case, it is these two:
- our very own `CollectedLoggerServiceProvider` (part of the `cli` package, provides the `CollectedLogger`) and 
- the `SimpleServiceProvider` (part of the `slf4j-simple` package, listed as a test dependency in the aggregator module, provides the `SimpleLogger`) 

This happens in the end-to-end tests and the CLI tests, which, naturally, are also executed when building the project.

Any time both providers are found, SLF4J chooses the one that was found first, which is JVM-dependent. We might want to avoid this since that should be possible fairly easily.

## Approaches ##

(A) Since we have the `CollectedLogger`, we could use it instead of the `SimpleLogger` during testing. This could be done via a dependency from the `aggregator` module to the `cli` module. We could also extract a new `logging` package to clear up the dependency structure.

(B) If it turns out we need the SimpleLogger specifically, the warning can be suppressed by changing the verbosity of SLF4J internal messages.

PR #2367, opened 2025-05-09
Fix multiple SLF4J providers during cli tests

robinmaisch:
This PR is a working solution for #2366, moving `Logger`s and `ProgressBar`s and the related tests into a new module and removing the dependency to `slf4j-simple`.
I do not claim this solution is perfect;	especially the new dependencies need a good look by someone who knows their Maven stuff better than I do.


dfuchss:
I'd either remove the slf4j simple or keep it as it was as this only effects the testing stage. You could also use a custom class loader for the tests it this is really needed. Let's talk next week :)

PR #2392, opened 2025-05-21
Frequency determination

heylissi:
This pull request contains methods for determining the frequency of matches across all submissions. Particularly rare matches may provide indications of plagiarism or frequent matches may indicate frequently used data structures that may be less important for plagiarism detection. Therefore, methods were implemented in the following to determine a frequency distribution across all matches contained in the submissions with different strategies.


tsaglam:
You can also remove the CLI integration for now, which avoids the need to update the help text. I think for the PR, we should remove it, as we currently do not want it to be a user-facing feature yet.

core/src/test/java/de/jplag/highlight_extraction/frequencyDetermination/StrategyIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+class StrategyIntegrationTest extends TestBase {
+
+    private static JPlagResult result;
+
+    @BeforeEach
+    void prepareMatchResult() throws ExitException {
+
+        JPlagOptions options = getDefaultOptions("PartialPlagiarism");
+        System.out.println(options);
```
tsaglam: Try to always use a logger in JPlag instead of System.out:

```
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
```
and
```
private static final Logger logger = LoggerFactory.getLogger(YOUR_CLASS.class);
logger.info(...)
```

core/src/test/java/de/jplag/highlight_extraction/frequencyDetermination/StrategyIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+        SubmissionSetBuilder builder = new SubmissionSetBuilder(options);
+        SubmissionSet submissionSet = builder.buildSubmissionSet();
+        LongestCommonSubsequenceSearch strategy = new LongestCommonSubsequenceSearch(options);
+        result = strategy.compareSubmissions(submissionSet);
+        System.out.println("result: !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!");
+        System.out.println(result);
```
tsaglam: Also, try to avoid excessive printing. If test cases log some information, it should be concise and useful.

core/src/test/java/de/jplag/highlight_extraction/frequencyDetermination/StrategyIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+import de.jplag.JPlagResult;
+import de.jplag.SubmissionSet;
+import de.jplag.SubmissionSetBuilder;
+import de.jplag.TestBase;
+import de.jplag.comparison.LongestCommonSubsequenceSearch;
+import de.jplag.exceptions.ExitException;
+import de.jplag.highlight_extraction.*;
+import de.jplag.options.JPlagOptions;
+
+class StrategyIntegrationTest extends TestBase {
```
tsaglam: Add Javadoc for all public classes and members.

core/src/main/java/de/jplag/highlight_extraction/ContainedStrategy.java
```diff
@@ -0,0 +1,81 @@
...
+     * @param tokens Token list of the match.
+     * @param comparisonId Identifier for the comparison.
+     * @param map Map that contains token subsequences and how often they occur across comparisons.
+     * @param size Minimum length of the considered submatches.
+     */
+
+    @Override
+    public void create(List<String> tokens, String comparisonId, Map<String, List<String>> map, int size) {
+        if (tokens.size() >= size) {
+            for (int j = size; j <= tokens.size(); j++) {
```
tsaglam: Try to stick to conventional variable names. `j` is usually used for two nested loops; here I would either use `i` or a speaking name.

core/src/main/java/de/jplag/highlight_extraction/ContainedStrategy.java
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Calculates the frequency of all matches and adds them to the map.
+     * @param tokens List of tokens representing the match.
+     * @param comparisonId Identifier for the comparison.
+     * @param map Map that associates token subsequences with how often they occur across comparisons.
+     * @param size The minimum sub length considered in other strategies.
+     */
+
+    @Override
+    public void check(List<String> tokens, String comparisonId, Map<String, List<String>> map, int size) {
```
tsaglam: As discussed, try to find speaking names for classes, methods, and variables, so future developers have an easier time reading your code if they do not know what it does.

core/src/main/java/de/jplag/highlight_extraction/FrequencyDetermination.java
```diff
@@ -0,0 +1,77 @@
...
+            List<Token> tokensNames = left.getTokenList();
+            List<TokenType> tokens = new ArrayList<>();
+            for (Token token : tokensNames) {
+                tokens.add(token.getType());
+            }
+            String comparisonId = comparison.toString();
+
+            for (Match match : comparison.matches()) {
+                int start = match.startOfFirst();
+                int len = match.lengthOfFirst();
```
tsaglam: Try to avoid abbreviations, both in the tests and the main code. See also the [article on contributing](https://github.com/jplag/JPlag/wiki/3.-Contributing-to-JPlag) in the wiki.

core/src/main/java/de/jplag/highlight_extraction/CompleteMatchesStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Adds all matches to a map using the token sequence as the key.
+     * @param tokens List of tokens representing the match.
+     * @param comparisonId Identifier for the comparison.
+     * @param map Map that associates token subsequences with how often they occur across comparisons.
+     * @param size The minimum sub length considered in other strategies.
+     */
+
+    @Override
+    public void create(List<String> tokens, String comparisonId, Map<String, List<String>> map, int size) {
```
tsaglam: We need to find more suitable representations in these data structures here. Currently, you are encoding most information with strings. For the token list `tokens`, the type should be `List<Token>` throughout your code. The map should also be mapping from and to specific types instead of strings. For the key, you want to check if the token sequence is identical. Instead of concatenating strings, use the token types. You could, for example, see if you can work with Objects.hash(). Or map the list to an immutable list of token types, then you can use that as a key (the second solution is probably better). For the values, also do not use a list of strings; instead, why not just keep an integer for the frequency?

core/src/main/java/de/jplag/highlight_extraction/ContainedStrategy.java
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Builds the map by adding token subsequences from a match.
+     * @param map Map that contains token subsequences and how often they occur across comparisons.
+     * @param tokens Token list of the match.
+     */
+
+    private void addToMap(Map<String, List<String>> map, List<String> tokens) {
+        String key = String.join(" ", tokens);
+        map.computeIfAbsent(key, k -> new ArrayList<>());
```
tsaglam: Is it on purpose that you never add anything to the list here?

core/src/main/java/de/jplag/highlight_extraction/FrequencyStrategyFactory.java
```diff
@@ -0,0 +1,27 @@
...
+     * @param strategy the applied frequency-strategy from the input
+     * @return An instance of the corresponding FrequencyStrategy implementation.
+     */
+    public static FrequencyStrategy create(FrequencyStrategies strategy) {
+        return switch (strategy) {
+            case COMPLETE_MATCHES -> new CompleteMatchesStrategy();
+            case SUB_MATCHES -> new SubMatchesStrategy();
+            case WINDOW_OF_MATCHES -> new WindowOfMatchesStrategy();
+            case CONTAINED_MATCHES -> new ContainedStrategy();
+            default -> throw new IllegalArgumentException("Unknown strategy: " + strategy);
```
tsaglam: You can move this method to the enum; this makes the code cleaner and removes the need for a factory class.
As numerals can have fields, each enum value can return its FrequencyStrategy, thus eliminating the switch.

core/src/main/java/de/jplag/highlight_extraction/SubMatchesStrategy.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Strategy that uses submatches of matches from the comparisons and calculates their frequency in the matches across
+ * all submissions.
+ */
+
```
tsaglam: After class comments, you often seem to have a superfluous empty line.

core/src/main/java/de/jplag/highlight_extraction/SubMatchesStrategy.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Builds the map by adding token subsequences from a match.
+     * @param map Map that contains token subsequences and how often they occur across comparisons.
+     * @param tokens Token list of the match.
+     */
+
+    private void addToMap(Map<String, List<String>> map, List<String> tokens) {
+        String key = String.join(" ", tokens);
+        map.computeIfAbsent(key, k -> new ArrayList<>());
```
tsaglam: Same here, do you add nothing to the list on purpose?

core/src/main/java/de/jplag/highlight_extraction/FrequencyDetermination.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * Applies the "create" and "check" methods of the frequency strategy.
+     * @param comparisons contains information of matches between two submissions.
+     * @return Map containing (sub-)matches and their frequency according to the strategy.
+     */
+    public Map<String, List<String>> runAnalysis(List<JPlagComparison> comparisons) {
+        frequencyBuilder(comparisons, freqencyStrategy::create);
+        frequencyBuilder(comparisons, freqencyStrategy::check);
```
tsaglam: I am not the biggest fan of the functional interface here. Either provide the strategy as an instance, or make this class abstract, merge the `FrequencyStrategy` into this class, and make the strategies subclasses (https://en.wikipedia.org/wiki/Strategy_pattern).

tsaglam (changes requested):
I briefly reviewed your code. The improvements are great! I found a few issues that we should address first. But you will also receive a full review at a later time. However, feel free to start with these early issues!

core/src/main/java/de/jplag/highlight_extraction/CompleteMatchesStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+     * @param tokens List of tokens representing the match.
+     * @param comparisonId Identifier for the comparison.
+     * @param map Map that associates token subsequences with how often they occur across comparisons.
+     * @param size The minimum sub length considered in other strategies.
+     */
+
+    @Override
+    public void check(List<String> tokens, String comparisonId, Map<String, List<String>> map, int size) {
+        String key = String.join(" ", tokens);
+        map.computeIfAbsent(key, k -> new java.util.ArrayList<>()).add(comparisonId);
```
tsaglam: Just another comment: If you count the comparisons, you cannot count if the same token sequence occurs multiple times in a single comparison. You should use an integer as the map value type instead of List<String>.

tsaglam:
> What is our convention for package names? the `hightlight_extraction` package is currently in snake_case. Should this be camelCase or all lower or kept as is?

lower case and no underscores, see https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html.
@heylissi also try to address the sonar issues and resolve the conversations in my review that you already fixed.

core/src/main/java/de/jplag/highlightextraction/SubSequenceUtil.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Calculates all possible sublists with min minSubSequenceSize length of the match matchTokenTypes.
+     * @param matchTokenTypes is a List of the match TokenTypes to create the SubSequences used for the SubSequenceStrategy
+     * and ContainedStrategy.
+     * @param minSubSequenceSize is the minimum considered size of the sub-Sequence-List.
+     * @return List of all as considered subSequences.
+     */
+    public static List<List<TokenType>> getSubSequences(List<TokenType> matchTokenTypes, int minSubSequenceSize) {
+        List<List<TokenType>> subSequences = new LinkedList<>();
```
TwoOfTwelve: To improve performance here replace this with an ArrayList and set the required length in the constructor. The length should be something like n*(n+1)/2 where n = matchTokenTypes.size() - minSubSequenceSize + 1. This avoid constant re-allocation of list members.

core/src/main/java/de/jplag/highlightextraction/ContainedMatchesStrategy.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public class ContainedMatchesStrategy implements FrequencyStrategy {
+    /**
+     * Adds all submatches with min size length of the matches to a map using the token sequence as the key.
+     * @param matchTokenTypes Token list of the match.
+     * @param frequencyMap Map that contains token subsequences and how often they occur across comparisons.
+     * @param strategyNumber Minimum length of the considered submatches.
+     */
+    @Override
+    public void addMatchToFrequencyMap(List<TokenType> matchTokenTypes, Map<Integer, Integer> frequencyMap, int strategyNumber) {
```
TwoOfTwelve: I think it would be easier to read this if you name strategyNumber something like minSequenceLength instead.

core/src/main/java/de/jplag/highlightextraction/ContainedMatchesStrategy.java
```diff
@@ -0,0 +1,32 @@
...
+        List<List<TokenType>> subSequences = getSubSequences(matchTokenTypes, strategyNumber);
+        for (List<TokenType> subSequence : subSequences) {
+            frequencyMap.putIfAbsent(subSequence.hashCode(), 0);
+        }
+        if (matchTokenTypes.size() >= strategyNumber) {
```
TwoOfTwelve: Am I right in that getSubSequences includes the original list as a sub-list? In that case this part would be redundant.

core/src/main/java/de/jplag/highlightextraction/SubSequenceUtil.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * Calculates all possible sublists with min minSubSequenceSize length of the match matchTokenTypes.
+     * @param matchTokenTypes is a List of the match TokenTypes to create the SubSequences used for the SubSequenceStrategy
+     * and ContainedStrategy.
+     * @param minSubSequenceSize is the minimum considered size of the sub-Sequence-List.
+     * @return List of all as considered subSequences.
+     */
+    public static List<List<TokenType>> getSubSequences(List<TokenType> matchTokenTypes, int minSubSequenceSize) {
```
TwoOfTwelve, line 25: JavaDoc should generally cover common edge-cases. In this case I think it's worth telling us if the original list is considered a sub-list or not.

core/src/main/java/de/jplag/highlightextraction/FrequencyDetermination.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+import java.util.Map;
+
+import de.jplag.*;
+
+/**
+ * Calculates frequencies of match subsequences across all comparisons according to different strategies.
+ */
+public class FrequencyDetermination {
+    private final Map<Integer, Integer> matchFrequencyMap = new HashMap<>();
```
TwoOfTwelve: I had some trouble with attributes initialized here with the android JVM. @tsaglam do we want to avoid this or is it fine for you?

core/src/main/java/de/jplag/highlightextraction/FrequencyDetermination.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Applies the "create" and "check" methods of the frequency strategy.
+     * @param comparisons contains information of matches between two submissions.
+     */
+    public void buildFrequencyMap(List<JPlagComparison> comparisons) {
+        for (JPlagComparison comparison : comparisons) {
+            Submission leftSubmission = comparison.firstSubmission();
+            List<Token> submissionTokens = leftSubmission.getTokenList();
+            List<TokenType> submissionTokenTypes = new ArrayList<>();
```
TwoOfTwelve: You could initialize this with submissionTokes.stream().map(Token::getType()).toList()

Pick whatever you think looks better

core/src/main/java/de/jplag/highlightextraction/FrequencyDetermination.java
```diff
@@ -0,0 +1,59 @@
...
+            for (Token token : submissionTokens) {
+                submissionTokenTypes.add(token.getType());
+            }
+
+            for (Match match : comparison.matches()) {
+                int startIndexOfMatch = match.startOfFirst();
+                int lengthOfMatch = match.lengthOfFirst();
+                if (startIndexOfMatch + lengthOfMatch > submissionTokenTypes.size()) {
+                    continue;
```
TwoOfTwelve: Is this something that happens often? In that case we might want to make this an issue. If not I'd suggest throwing an exception here.

core/src/main/java/de/jplag/highlightextraction/FrequencyStrategies.java
```diff
@@ -0,0 +1,24 @@
+package de.jplag.highlightextraction;
+
+/**
+ * Enum representing the different strategies for frequency calculation of token subsequences in comparisons.
+ */
+public enum FrequencyStrategies {
```
TwoOfTwelve, line 6: Have you considered solving this using a sealed interface instead? That would avoid having this additional enum.

core/src/main/java/de/jplag/highlightextraction/SubMatchesStrategy.java
```diff
@@ -0,0 +1,30 @@
...
+
+import static de.jplag.highlightextraction.SubSequenceUtil.getSubSequences;
+
+import java.util.List;
+import java.util.Map;
+
+import de.jplag.TokenType;
+
+/**
+ * Strategy that uses submatches of matches from the comparisons and calculates their frequency in the matches across
```
TwoOfTwelve: Maybe quickly describe here what makes this different from ContainedMatches. You could also name this TrueSubMatchesStrategy to make it clear that the original match isn't considered here (I assume that is what makes this different)

core/src/main/java/de/jplag/highlightextraction/SubMatchesStrategy.java
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Creates submatches to build the keys and adds their frequencies to the frequencyMap.
+     * @param frequencyMap Map that contains token subsequences and how often they occur across comparisons.
+     * @param matchTokenTypes List of matchTokenTypes representing the match.
+     * @param strategyNumber Minimum length of the considered submatches.
+     */
+    @Override
+    public void addMatchToFrequencyMap(List<TokenType> matchTokenTypes, Map<Integer, Integer> frequencyMap, int strategyNumber) {
+        List<List<TokenType>> subSequences = getSubSequences(matchTokenTypes, strategyNumber);
+        for (List<TokenType> subSequence : subSequences) {
```
TwoOfTwelve, line 25: Related to the issue above: If getSubSequences returns the original sequence this doesn't work

core/src/main/java/de/jplag/highlightextraction/WindowOfMatchesStrategy.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Calculates all possible Sublists with length of windowSize.
+     * @param matchTokenTypes tokens of the Match.
+     * @param windowSize considered size of the sublists.
+     * @return List of all as considered sublists.
+     */
+    public static List<List<TokenType>> getWindowSequences(List<TokenType> matchTokenTypes, int windowSize) {
+        List<List<TokenType>> windowSequences = new LinkedList<>();
+        if (matchTokenTypes.size() >= windowSize) {
```
TwoOfTwelve: This if is redundant, as the loop below will run 0 times if this condition is false.

core/src/test/java/de/jplag/highlightextraction/frequencydetermination/StrategyIntegrationTest.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Logger for outputting test information.
+     */
+    private static final Logger logger = LoggerFactory.getLogger(StrategyIntegrationTest.class);
+
+    /**
+     * Prepares the comparison result before each test.
+     * @throws ExitException if building the submissionSet or the comparisons fails.
+     */
+    @BeforeEach
```
TwoOfTwelve: Is the result different for each test or changed in the tests? If not you could make this BeforeAll instead to avoid running JPlag multiple times

core/src/test/java/de/jplag/highlightextraction/frequencydetermination/StrategyTest.java
```diff
@@ -0,0 +1,317 @@
...
+    @DisplayName("Test check() of window strategy")
+    void testWindowOfMatchesStrategy() {
+        FrequencyStrategy strategy = new WindowOfMatchesStrategy();
+        Map<Integer, Integer> windowMap = new HashMap<>();
+        int windowSize = 5;
+
+        List<TokenType> matchTokenTypes = getMatchTokenTypes(matchShort);
+        strategy.addMatchToFrequencyMap(matchTokenTypes, windowMap, windowSize);
+
+        // Checks if all expected windows are added.
```
TwoOfTwelve: I don't really like this type of comment. Have you considered making each of those checks it's own function?

core/src/main/java/de/jplag/highlightextraction/FrequencyDetermination.java
tsaglam: Yes, this should be in the constructor, among other reasons, for style and consistency.

core/src/main/java/de/jplag/highlightextraction/WindowOfMatchesStrategy.java
```diff
@@ -0,0 +1,46 @@
...
+     * Adds all submatches with window length of the matches to a map using the token sequence as the key.
+     * @param matchTokenTypes Token list of the match.
+     * @param frequencyMap Map that contains token subsequences and how often they occur across comparisons.
+     * @param strategyNumber The length of the considered token window.
+     */
+    @Override
+    public void addMatchToFrequencyMap(List<TokenType> matchTokenTypes, Map<Integer, Integer> frequencyMap, int strategyNumber) {
+        List<List<TokenType>> windowSequences = getWindowSequences(matchTokenTypes, strategyNumber);
+        for (List<TokenType> windowSequence : windowSequences) {
+            frequencyMap.put(windowSequence.hashCode(), frequencyMap.getOrDefault(windowSequence.hashCode(), 0) + 1);
```
robinmaisch: This line occurs in all strategies. To reduce duplicate code, I propose extracting this line to a method `addSequence`/`registerSequence` or similar inside FrequencyDetermination.
Then, you can pass a reference to that method (a `Consumer<List<TokenType>>`) as an argument instead of the frequencyMap, which will look much cleaner.
Lastly, I think the name `processMatch` or `processMatchTokenTypes` would better describe what happens here. `addMatchToFrequencyMap` is very specific and sounds like a one-step process, but in reality, the different strategies actually perform complex operations here.

core/src/main/java/de/jplag/highlightextraction/SubMatchesStrategy.java
heylissi: Thanks for the suggestion! The original match is considered in this strategy. I'll update the description accordingly.

core/src/main/java/de/jplag/highlightextraction/ContainedMatchesStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+     * @param matchTokenTypes Token list of the match.
+     * @param frequencyMap Map that contains token subsequences and how often they occur across comparisons.
+     * @param minSubSequenceLength Minimum length of the considered submatches.
+     */
+    @Override
+    public void processMatchTokenTypes(List<TokenType> matchTokenTypes, Map<Integer, Integer> frequencyMap, int minSubSequenceLength) {
+        Consumer<List<TokenType>> sequenceConsumer = seq -> addSequence(frequencyMap, seq);
+        List<List<TokenType>> subSequences = getSubSequences(matchTokenTypes, minSubSequenceLength);
+        for (List<TokenType> subSequence : subSequences) {
+            frequencyMap.putIfAbsent(subSequence.hashCode(), 0);
```
tsaglam: After a discussion with @TwoOfTwelve: We should probably use the list itself as key instead of the hashcode to reduce the chance of collisions. This means we just replace `subSequence.hashCode()` with `subSequence` and also change the hashmap to `Map<List<TokenType>, Integer>`.

core/src/main/java/de/jplag/highlightextraction/CompleteMatchesStrategy.java
```diff
@@ -0,0 +1,26 @@
+package de.jplag.highlightextraction;
+
+import static de.jplag.highlightextraction.SubSequenceUtil.addSequence;
```
tsaglam: Also, try to replace the static import with a non-static import.
Then call `SubSequenceUtil.addSequence()`.

tsaglam (approved):

core/src/main/java/de/jplag/highlightextraction/SubSequenceUtil.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public final class SubSequenceUtil {
+    private SubSequenceUtil() {
+        // private constructor to prevent instantiation
+    }
+
+    /**
+     * Calculates all possible sublists with min minSubSequenceSize length of the matchTokenTypes.
+     * @param matchTokenTypes is a List of the match, in TokenTypes, to create the SubSequences used for the
+     * SubSequenceStrategy and ContainedStrategy. The original list is also included. All contiguous (Sub-)lists, that can
```
TwoOfTwelve: I'd rather put this edge case in the general description instead of the parameter description.

Also, if you want to, you could make this method generic, so it works for all possible types.

core/src/main/java/de/jplag/highlightextraction/CompleteMatchesStrategy.java
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * Adds the given token sequence to the map and updates its frequency.
+     * @param matchTokenTypes List of tokensTypes representing the match.
+     * @param frequencyMap Map that associates token subsequences with the number of times they occur across comparisons.
+     * @param strategyNumber Ignored in this strategy. The minimum sub length considered in other strategies.
+     */
+    @Override
+    public void processMatchTokenTypes(List<TokenType> matchTokenTypes, Map<List<TokenType>, Integer> frequencyMap, int strategyNumber) {
+        Consumer<List<TokenType>> sequenceConsumer = seq -> SubSequenceUtil.addSequence(frequencyMap, seq);
+        sequenceConsumer.accept(matchTokenTypes);
```
TwoOfTwelve: Is there any reason for the consumer? Couldn't you just write:

```java
SubSequenceUtil.addSequence(frequencyMap, matchTokenTypes);
```

core/src/main/java/de/jplag/highlightextraction/ContainedMatchesStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+     * @param matchTokenTypes List of tokens representing the match.
+     * @param frequencyMap Map that contains token subsequences and how often they occur across comparisons.
+     * @param minSubSequenceLength Minimum length of the considered submatches.
+     */
+    @Override
+    public void processMatchTokenTypes(List<TokenType> matchTokenTypes, Map<List<TokenType>, Integer> frequencyMap, int minSubSequenceLength) {
+        Consumer<List<TokenType>> sequenceConsumer = seq -> SubSequenceUtil.addSequence(frequencyMap, seq);
+        List<List<TokenType>> subSequences = SubSequenceUtil.getSubSequences(matchTokenTypes, minSubSequenceLength);
+        for (List<TokenType> subSequence : subSequences) {
+            frequencyMap.putIfAbsent(subSequence, 0);
```
TwoOfTwelve: Shouldn't you increase the count if the match already exists or do you not want to count sub-matches? In that case maybe put that in the documentation.

core/src/main/java/de/jplag/highlightextraction/ContainedMatchesStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+        }
+        if (matchTokenTypes.size() >= minSubSequenceLength) {
+            sequenceConsumer.accept(matchTokenTypes);
```
TwoOfTwelve: Like above, does the consumer have any benefit?

core/src/main/java/de/jplag/highlightextraction/SubMatchesStrategy.java
```diff
@@ -0,0 +1,31 @@
...
+     * @param frequencyMap Map that contains token subsequences and how often they occur across comparisons.
+     * @param matchTokenTypes List of matchTokenTypes representing the match.
+     * @param minSubSequenceSize Minimum length of the considered submatches.
+     */
+    @Override
+    public void processMatchTokenTypes(List<TokenType> matchTokenTypes, Map<List<TokenType>, Integer> frequencyMap, int minSubSequenceSize) {
+        Consumer<List<TokenType>> sequenceConsumer = sequence -> SubSequenceUtil.addSequence(frequencyMap, sequence);
+        List<List<TokenType>> subSequences = SubSequenceUtil.getSubSequences(matchTokenTypes, minSubSequenceSize);
+        for (List<TokenType> subSequence : subSequences) {
+            sequenceConsumer.accept(subSequence);
```
TwoOfTwelve: Same as above with the consumer

core/src/main/java/de/jplag/highlightextraction/SubMatchesStrategy.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.function.Consumer;
+
+import de.jplag.TokenType;
+
+/**
+ * Strategy that uses submatches of matches from the comparisons and calculates the frequency of all submatches in the
+ * matches across all submissions. The full Match is also considered a submatch if it is longer or equals to
+ * minSubSequenceSize.
+ */
+public class SubMatchesStrategy implements FrequencyStrategy {
```
TwoOfTwelve, line 12: I'm still not sure what the difference to ContainedMatches is. Is there a difference for the strategy between calculating the frequency and counting the number of occurences? If not, maybe use the same phrasing in both strategies.

core/src/main/java/de/jplag/highlightextraction/WindowOfMatchesStrategy.java
```diff
@@ -0,0 +1,47 @@
...
+     * @param matchTokenTypes Token list of the match.
+     * @param frequencyMap Map that contains token subsequences and how often they occur across comparisons.
+     * @param strategyNumber The length of the considered token window.
+     */
+    @Override
+    public void processMatchTokenTypes(List<TokenType> matchTokenTypes, Map<List<TokenType>, Integer> frequencyMap, int strategyNumber) {
+        Consumer<List<TokenType>> sequenceConsumer = seq -> SubSequenceUtil.addSequence(frequencyMap, seq);
+        List<List<TokenType>> windowSequences = getWindowSequences(matchTokenTypes, strategyNumber);
+        for (List<TokenType> windowSequence : windowSequences) {
+            sequenceConsumer.accept(windowSequence);
```
TwoOfTwelve: Again, same as above

core/src/main/java/de/jplag/highlightextraction/CompleteMatchesStrategy.java
robinmaisch:
```suggestion
    public void processMatchTokenTypes(List<TokenType> matchTokenTypes, Consumer<List<TokenType>> sequenceConsumer, int strategyNumber) {
        sequenceConsumer.accept(matchTokenTypes);
```

There was a misunderstanding as to how the `Consumer` should be integrated here. This is what I had in mind, and this was supposed to prevent random access to the map in this method and reduce duplicate code. I will add a proposed change to the method call as well.

core/src/main/java/de/jplag/highlightextraction/SubMatchesStrategy.java
robinmaisch: @heylissi, please provide a brief, but precise description so that the difference becomes clearer.
A suitable template might be: "Strategy that counts all occurrences of X inside Y".

core/src/main/java/de/jplag/highlightextraction/SubSequenceUtil.java
```diff
@@ -0,0 +1,48 @@
...
+     * be taken from the matchTokenTypes list that have length greater or equal minSubSequenceSize.
+     * @param minSubSequenceSize is the minimum considered size of the sub-Sequence-List.
+     * @return List of all as considered subSequences.
+     */
+    public static List<List<TokenType>> getSubSequences(List<TokenType> matchTokenTypes, int minSubSequenceSize) {
+        int possibleSubSequenceStartPositions = matchTokenTypes.size() - minSubSequenceSize + 1;
+        int subSequencyCount = possibleSubSequenceStartPositions * (possibleSubSequenceStartPositions + 1) / 2;
+        List<List<TokenType>> subSequences = new ArrayList<>(subSequencyCount);
+        if (matchTokenTypes.size() >= minSubSequenceSize) {
```
robinmaisch: To decrease complexity, please check for the inverse condition before Line 26 and return `List.of()` immediately if that condition is true.

core/src/main/java/de/jplag/highlightextraction/SubSequenceUtil.java
```diff
@@ -0,0 +1,48 @@
...
+            for (int subSequenceSize = minSubSequenceSize; subSequenceSize <= matchTokenTypes.size(); subSequenceSize++) {
+                for (int windowStartIndex = 0; windowStartIndex <= matchTokenTypes.size() - subSequenceSize; windowStartIndex++) {
+                    List<TokenType> subSequence = new ArrayList<>(matchTokenTypes.subList(windowStartIndex, windowStartIndex + subSequenceSize));
```
robinmaisch: Possible optimization point here:
Under the assumption that the `matchTokenTypes` will not be changed for the duration of the frequency analysis, the lists returned by the `subList` would also be suitable (so, no new ArrayList needed). These lists access the base list directly and are thus extremely lightweight and low in space consumption.

core/src/main/java/de/jplag/highlightextraction/SubSequenceUtil.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /**
+     * Updates the frequency of the given sequence in the frequency map.
+     * @param frequencyMap Map of sequences to their frequency counts.
+     * @param sequence The token sequence whose frequency will be updated.
+     */
+    static void addSequence(Map<List<TokenType>, Integer> frequencyMap, List<TokenType> sequence) {
+        frequencyMap.put(sequence, frequencyMap.getOrDefault(sequence, 0) + 1);
+    }
```
robinmaisch, line 41: I would move this to `FrequencyDetermination`, as the frequencyMap is stored there, making calls there more concise.

core/src/main/java/de/jplag/highlightextraction/FrequencyDetermination.java
```diff
@@ -0,0 +1,57 @@
...
+            List<TokenType> submissionTokenTypes = submissionTokens.stream().map(Token::getType).toList();
+
+            for (Match match : comparison.matches()) {
+                int startIndexOfMatch = match.startOfFirst();
+                int lengthOfMatch = match.lengthOfFirst();
+                if (startIndexOfMatch + lengthOfMatch > submissionTokenTypes.size()) {
+                    throw new IllegalArgumentException("startIndexOfMatch + lengthOfMatch <= submissionTokenTypes.size()");
+                }
+                List<TokenType> matchTokenTypes = submissionTokenTypes.subList(startIndexOfMatch, startIndexOfMatch + lengthOfMatch);
+                frequencyStrategy.processMatchTokenTypes(matchTokenTypes, matchFrequencyMap, strategyNumber);
```
robinmaisch:
```suggestion
                frequencyStrategy.processMatchTokenTypes(matchTokenTypes, this::addSequence, strategyNumber);
```

robinmaisch (changes requested):
@heylissi, if possible, please handle these changes soon, so that we can merge and move on to the second PR :)

heylissi:
> @heylissi, what is the status here? Are all remaining feedback points addressed?

Hi, I have checked the Javadoc and corrected typos. As far as I can see, all comments have been considered.

robinmaisch (approved):
Looks good to me!

PR #2396, opened 2025-05-23
Introduce asymmetric matches

### LINKED ISSUE #2395: Assumptions regarding matches and match length
JPlag always had the assumption that a match (meaning two matching subsequences of tokens between two programs) has the same length on both sides. Note that this does not mean the code fragments that were matched have the same length, but the number of tokens that represent these code fragments is the same.

In the future, we want to move away from this assumption. Especially when using defence mechanisms like subsequence match merging, matches can have different lengths. We circumvented this issue by removing tokens in the past. In the future, we want to change how we represent matches by introducing two lengths for each side of a match (see #2268).

This has almost no effect on users. The main difference is a change in the API:

https://github.com/jplag/JPlag/blob/df45b33a2010265fdffca33ea17762536db42d9d/core/src/main/java/de/jplag/Match.java#L3-L10

This change would introduce two lengths per match.

tsaglam:
Introduces asymmetric matches, updates, and improves subsequence match merging. Closes #2395.

- Change how matches are modeled, introduce two length methods for both sides of the match.
- Introduce a minimal length method for cases where the totality of the match is measured.
- Adapt greedy string tiling (GST) and test cases.
    - Note that greedy string tiling still produces symmetric matches.
    - However, post-processing mechanisms may make them asymmetric.
- Adapt subsequence match merging (SMM):
    - No longer delete tokens between matches when merging them.
    - Instead, use asymmetric matches to include them in the merged match.
    - Update neighbors after merging instead of re-generating them.
    - The performance of the merging alone is thus strongly increased[^1].
    - The overhead of SMM on the greedy string tiling stays the same[^2].
- Backwards compatibility:
	- This PR adds a deprecated length method in the match class that is identical to the formerly generated access method.
	- It maps to the minimum length of both sides of the match.
	- This allows backwards compatibility, as currently, matches are still symmetrical, except when using SMM.

TODO:

- [x] Adapt report viewer and the code for creating the result files.
- [x] Discuss API-breaking state.
- [x] Remove hotfix for SMM + basecode

[^1]: Now takes 4 sec instead of 20 sec for 10K files containing 655K LOC divided into 434 submissions at ~1500 LOC each (measured on a M1 pro MacBook).
[^2]: SMM requires matches below the min token match, thus drastically increasing the runtime of the GST.


core/src/main/java/de/jplag/Match.java
```diff
@@ -50,4 +50,15 @@ public int endOfSecond() {
...
         return Math.min(lengthOfFirst, lengthOfSecond);
     }
+
+    /**
+     * @return length is the length of these similar sections (number of tokens).
+     * @deprecated matches are no longer required to be symmetrical. Thus, both sides can have different lengths. This
+     * method now returns the minimal length.
+     * @see Match#minimumLength()
+     */
+    @Deprecated(since = "4.6.2", forRemoval = true)
```
Kr0nox, line 60: I believe you meant to put `6.2.0` or something similar for the version
(should it not be 7.0.0?)

Kr0nox:
@tsaglam If you could take a quick look at the texts in the tooltips of `report-viewer/src/components/fileDisplaying/MatchList.vue` lines 84ff. that would be good. otherwise the report viewer side should be done

PR #2412, opened 2025-06-02
Improve the comparison algorithm performance for large workloads

tsaglam:
This PR introduces a speed increase of around 4-10% for the greedy string tiling algorithm[^1]. Multiple measures are taken to improve performance. Note that the majority of the speedup is attributed to virtual threads.

- Switch to virtual threads for the parallelization of the pairwise comparison in `LongestCommonSubsequenceSearch`.
- Optimize calculations for the subsequence hash lookup by replacing arithmetic operations with bitwise shifts in `SubsequenceHashLookupTable`[^2].
- Avoid excessive hash table lookups via more efficient methods such as `computeIfAbsent()` in `SubsequenceHashLookupTable`.

Moreover, this PR contains a minor optimization to the `DummyBar`, which is used when a progress bar is not supported.

[^1]: Dataset of 434 submissions at around 1500 LOC per submission, totalling 655 KLOC in total. Measured on an M1 Pro MacBook for runs with and without subsequence match merging (`--match-merging`) enabled.
[^2]: Multiplication by `2^(windowSize - 1)` is replaced with a left shift: `hashValue * 2^(windowSize - 1)` --> `hashValue << (windowSize - 1)`. Multiplying the rolling hash by 2 is replaced with a left shift: `hash * 2 → hash << 1`.


uuqjz (approved):

tsaglam:
@jplag/devteam @jplag/maintainer reminder to take a look at this.

PR #2423, opened 2025-06-05
Improved error messages for token position tests

### LINKED ISSUE #2305: Usability of the language module test framework
While our [framework](https://github.com/jplag/JPlag/tree/main/language-testutils) for language module tests is very advanced, creating tests can be painful as the test framework fails/crashes without intuitive errors when the testing input is incorrect. We should add JUnit Assumptions in the API so devs receive meaningful feedback for invalid input.

Examples found so far:
- [x] Token position tests fail weirdly when a file instead of a directory is passed (addressed in #2303)
- [ ] Token position tests crash when empty lines are in the input files. If a line does not start with the expected prefixes, a helpful error message should be given, and the test should fail via an `assume(...)`.
- [ ] Any file-based tests that are generated via our API should check assumptions.

TwoOfTwelve:
Fixes #2305

### DIFF
--- a/language-testutils/src/test/java/de/jplag/testutils/datacollector/TokenPositionTestData.java
+++ b/language-testutils/src/test/java/de/jplag/testutils/datacollector/TokenPositionTestData.java
@@ -19,7 +19,12 @@ import de.jplag.util.FileUtils;
  * information for tests. The sources can be used as regular test sources.
  */
 public class TokenPositionTestData implements TestData {
-    private final static String TOKEN_DEFINITION_LINE_REGEX = "\\$[\\s]*\\| [a-zA-Z0-9_]+ [0-9]+[\\s]*";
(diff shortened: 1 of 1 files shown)


language-testutils/src/test/java/de/jplag/testutils/datacollector/TokenPositionTestData.java
```diff
@@ -40,22 +44,46 @@ private void readFile(File testFile) throws IOException {
...
             }
         }
     }
 
+    private void extractTokenData(String line, int currentSourceLine) {
+        if (!line.matches(TOKEN_DEFINITION_LINE_REGEX)) {
+            Assertions.fail("Invalid line for token position test " + this.descriptor + " at line: " + line);
+        }
+
+        int column = line.indexOf('|');
```
tsaglam, line 84: We have quite a few chars with specific meaning in this PR (>, $, #, |). Can you introduce well-named constants for them?

tsaglam (changes requested):

PR #2426, opened 2025-06-10
Extracting comments from submissions

c0derMo:
This PR adds the option to extract comments from submitted source code files.
To achieve this, a new `CommentExtractor` class has been implemented, which parses over the textual content of source code files, identifying line and block comments by previously set character sequences. These character sequences can be defined for each language, and are currently only implemented for Java. In the future, the default implementation returning the extractor settings will be removed, and every language should have it's settings defined, or opt out of the comment processing.

The comment extractor serves as a foundation for a future PR, to enhance the plagiarism detection by analyzing and comparing the extracted comments.


core/src/main/java/de/jplag/Submission.java
```diff
@@ -263,11 +268,28 @@ private static File createErrorDirectory(String... subdirectoryNames) {
             return false;
         }
 
+        this.extractAndParseComments();
```
tsaglam: We should introduce an option to enable and disable this. This involves two changes:

- An additional CLI flag in the CLI module (I would group it under the argument group `advanced`).
- An additional option in the `JPlagOptions` class (default: false).

If we merge this right now, the next release would already extract comments even though nothing is done with them.

core/src/main/java/de/jplag/Submission.java
```diff
@@ -263,11 +268,28 @@ private static File createErrorDirectory(String... subdirectoryNames) {
...
+
         tokenList = Collections.unmodifiableList(tokenList);
         state = VALID;
         return true;
     }
 
+    private void extractAndParseComments() {
+        CommentExtractorSettings commentExtractorSettings = language.getCommentExtractorSettings();
+        if (commentExtractorSettings != null) {
```
tsaglam: To avoid null checks here, you could also provide a settings instance with empty lists in the default method of the language interface. This makes the code a bit cleaner.

language-api/src/main/java/de/jplag/Language.java
```diff
@@ -138,4 +139,12 @@ default boolean supportsMultiLanguage() {
     default boolean hasPriority() {
         return false;
     }
+
+    /**
+     * Return {@code null} if language does not support comment extraction.
+     * @return Settings for the comment extractor, or null if language does not support comment extraction.
+     */
+    default CommentExtractorSettings getCommentExtractorSettings() {
+        return null;
```
tsaglam: See comment above.

language-api/src/test/java/de/jplag/CommentExtractorTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import de.jplag.commentextraction.Comment;
+import de.jplag.commentextraction.CommentExtractor;
+import de.jplag.commentextraction.CommentExtractorSettings;
+import de.jplag.commentextraction.CommentType;
+import de.jplag.commentextraction.EnvironmentDelimiter;
+
+class CommentExtractorTest {
```
tsaglam, line 21: Add a short JDoc comment that explains what this class is testing (as in what feature).

language-api/src/test/java/de/jplag/CommentExtractorTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private static final Path TEST_FILE_LOCATION = Path.of("src", "test", "resources", "de", "jplag", "samples");
```
tsaglam, line 23: You may be able to reuse some functionality of the common test base class. Maybe it is just the base directory, but even there, you could consider reusing it.

language-api/src/test/java/de/jplag/CommentExtractorTest.java
```diff
@@ -0,0 +1,51 @@
...
+    private static final String TEST_FILE_NAME = "CommentExtractorTest.txt";
+
+    @Test
+    void testCommentExtractor() throws IOException {
```
tsaglam: use the `@Description` tag for unit tests to give more context on the specific test scenario.

tsaglam (changes requested):
Just a few minor issues to consider. Also, try to address the sonar issues.

language-api/src/test/java/de/jplag/CommentExtractorTest.java
c0derMo: The `language-api` module currently has no test base class to my knowledge, extending the one from the `core` would introduce another (test) dependency to the module

tsaglam: Ah, right, makes sense. I did not consider that the test is in the language API.

cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -107,6 +107,9 @@ public static class Advanced {
 
         @Option(names = "--log-level", description = "Set the log level for the cli.")
         public Level logLevel = Level.INFO;
+
+        @Option(names = "--include-comments", description = "Enables comment extraction.")
+        public boolean includeComments = false;
```
```suggestion
        @Option(names = "--comments", description = "Analyze similarity of comments. This never reduces program similarity.")
        public boolean includeComments = false;
```

Keeps the flag short, but extends the description a bit.

core/src/main/java/de/jplag/Submission.java
```diff
@@ -232,10 +237,11 @@ private static File createErrorDirectory(String... subdirectoryNames) {
      * @param debugParser specifies if the submission should be copied upon parsing errors.
      * @param normalize specifies if the token sequences should be normalized.
      * @param minimalTokens specifies the minimum number of tokens required of a valid submission.
+     * @param includeComments specifies if comments should be extracted.
      * @return Whether parsing was successful.
      * @throws LanguageException if the language parser is not able to parse at all.
      */
-    /* package-private */ boolean parse(boolean debugParser, boolean normalize, int minimalTokens) throws LanguageException {
+    /* package-private */ boolean parse(boolean debugParser, boolean normalize, int minimalTokens, boolean includeComments) throws LanguageException {
```
tsaglam: I think we should use the term `analyzeComments`, as it is a bit more precise.

cli/src/main/java/de/jplag/cli/options/CliOptions.java
tsaglam: Also rename the option to `analyzeComments`. See below.

tsaglam (changes requested):
Just a few minor comments regarding the naming of the option and CLI flag.

cli/src/main/java/de/jplag/cli/options/CliOptions.java
Kr0nox: Please also make sure to also extract the new option in the report Viewer.
For that add it to the interface in report-viewer/src/model/CliOptions.ts and add it to be displayed in report-viewer/src/views/InformationView.vue

README.md
```diff
@@ -122,6 +122,9 @@ Parameter descriptions:
                           but might lead to more false-positives.
 
 Advanced
+      --comments        Analyze similarity of comments. Increases the
+                          similarity of submissions if similar comments are
+                          found, but never decreases it.
```
TwoOfTwelve: The feature should be documented in a bit more detail in the wiki. That can be a different PR, though.

core/src/main/java/de/jplag/Submission.java
```diff
@@ -263,11 +269,30 @@ private static File createErrorDirectory(String... subdirectoryNames) {
...
     }
 
+    private void extractAndParseComments() {
+        CommentExtractorSettings commentExtractorSettings = language.getCommentExtractorSettings();
+        if (commentExtractorSettings.hasComments()) {
+            for (File file : files) {
+                try {
+                    CommentExtractor extractor = new CommentExtractor(file, commentExtractorSettings);
+                    comments.addAll(extractor.extract());
+                } catch (IOException e) {
```
TwoOfTwelve: I think it would be better to let the CommentExtractor print the warning and return an empty list.

core/src/main/java/de/jplag/Submission.java
```diff
@@ -263,11 +269,30 @@ private static File createErrorDirectory(String... subdirectoryNames) {
...
+        }
+
         tokenList = Collections.unmodifiableList(tokenList);
         state = VALID;
         return true;
     }
 
+    private void extractAndParseComments() {
+        CommentExtractorSettings commentExtractorSettings = language.getCommentExtractorSettings();
+        if (commentExtractorSettings.hasComments()) {
```
TwoOfTwelve: I think it would be more intuitive to make hasComments() a function of Language or call is doesLanguageSupportComments()

core/src/main/java/de/jplag/Submission.java
```diff
@@ -263,11 +269,30 @@ private static File createErrorDirectory(String... subdirectoryNames) {
...
+            for (File file : files) {
```
TwoOfTwelve, line 285: @tsaglam Would you prefer this.files?

language-api/src/main/java/de/jplag/commentextraction/Comment.java
```diff
@@ -0,0 +1,15 @@
+package de.jplag.commentextraction;
+
+import java.io.File;
+
+/**
+ * Stores a single comment of a submitted file.
+ * @param file File the comment originated from
+ * @param content Textual content of the comment
+ * @param line Line of the comment (1-indexed)
+ * @param column Column of the comment (1-indexed)
```
TwoOfTwelve, line 10: @tsaglam Do you think we should introduce a type Index that encapsulates the 1-based behavior and some common index operations? We seem to be using it in a lot of places

language-api/src/main/java/de/jplag/commentextraction/CommentExtractor.java
```diff
@@ -0,0 +1,172 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import de.jplag.util.FileUtils;
+
+/**
+ * Extracts comments from submitted files, by reading and parsing the file content manually.
+ * @author Moritz Rimpf
```
TwoOfTwelve: We are not using the author field

language-api/src/main/java/de/jplag/commentextraction/CommentExtractor.java
```diff
@@ -0,0 +1,172 @@
...
+        this.file = file;
+        this.currentCol = 1;
+        this.currentLine = 1;
+    }
+
+    private void match(String expected) {
+        if (remainingContent.startsWith(expected)) {
+            this.advance(expected.length());
+        } else {
+            throw new MatchException("Matched incorrectly, expected: " + expected + ", received: " + remainingContent.substring(0, expected.length()),
```
TwoOfTwelve: I'm not entirely sure about that exception type. We want to avoid RuntimeExceptions and this also seems to be an internally used exception type. Furthermore MatchException in the context of JPLag makes this sounds like it has something to do with Greedy String Tiling. Is there a reason not to use an ExitException or something similar from JPlag?

language-api/src/main/java/de/jplag/commentextraction/CommentExtractor.java
```diff
@@ -0,0 +1,172 @@
...
+                    null);
+        }
+    }
+
+    private String advance(int length) {
+        String advancedBy = remainingContent.substring(0, length);
+        remainingContent = remainingContent.substring(length);
+
+        // Checking for line breaks
```
TwoOfTwelve: Can you make a dedicated function for this instead of this comment? Also, if this just refers to the condition of the if, I think it's obvious you are checking if the string contains a line break

language-api/src/test/java/de/jplag/CommentExtractorTest.java
```diff
@@ -0,0 +1,56 @@
...
+        File input = new File(TEST_FILE_LOCATION.toFile(), TEST_FILE_NAME);
+
+        CommentExtractor extractor = new CommentExtractor(input, settings);
+
+        List<Comment> comments = extractor.extract();
+
+        assertEquals(2, comments.size());
+        assertEquals(new Comment(input, " This is a line comment.", 2, 3, CommentType.LINE), comments.get(0));
+
+        StringJoiner multilineComment = new StringJoiner(System.lineSeparator());
```
TwoOfTwelve: Is there any reason not to use this:
```java
String.join(System.lineSeparator(), " This is", "a", "multiline", "comment", "");
```
That feels more concise to me

README.md
tsaglam: If results are as promising as I hope they are, we can also add this feature to the FAQ. I would second a different PR for documentation when the complete feature is done.

core/src/main/java/de/jplag/Submission.java
tsaglam: No, I prefer using `this` only when needed, as modern IDEs highlight the difference between local variables and fields via color coding.

language-api/src/main/java/de/jplag/commentextraction/Comment.java
tsaglam: Currently, I do not see a need for this.

language-api/src/main/java/de/jplag/commentextraction/CommentExtractor.java
tsaglam: Another underlying question is whether a crash here should really end the JPlag run? Is there no way to fail gracefully? E.g. skip comment extraction and log an error?

languages/java/src/main/java/de/jplag/java/JavaLanguage.java
```diff
@@ -55,4 +57,13 @@ public boolean supportsNormalization() {
     public String toString() {
         return this.getIdentifier();
     }
+
+    @Override
+    public CommentExtractorSettings getCommentExtractorSettings() {
+        return new CommentExtractorSettings(
+                List.of(new EnvironmentDelimiter("\""), new EnvironmentDelimiter("'"), new EnvironmentDelimiter("\"\"\"")), // No comment environment
```
TwoOfTwelve: I think you have to put the multiline string before the normal string. Otherwise your code will just look at the first '"' and read it as a normal string. This will not work if quotes are present in a multiline string

language-api/src/test/java/de/jplag/CommentExtractorTest.java
robinmaisch: One argument might be that this matches the expected comment visually. Other than that, I would not mind too much.

core/src/main/java/de/jplag/Submission.java
c0derMo: The original idea behind this was to avoid adding two methods to the language interface, i'd wrap it into an `Optional` now, to maybe make this a bit more intuitive, without adding more methods than needed

Kr0nox:
Is the extraction and their results exported in the report or given to the user in some way? 
If not is this planned in this PR?
If not, would it make sense to move the CLI flag to that PR, as otherwise we would add a flag that does nothing for the user

tsaglam:
> What are your thoughts on that?

We do not need to remove the CLI code, but the options should be either disabled or hidden. Also, the readme must not include the feature.

report-viewer/src/views/InformationView.vue
```diff
@@ -27,6 +27,9 @@
           <TextInformation label="Exclusion File Name">{{
             options.exclusionFileName
           }}</TextInformation>
+          <TextInformation label="Analyze comments">{{
```
Kr0nox: I would also avoid displaying the option in the report viewer (but keep it in the AbstractOptions interface)

PR #2438, opened 2025-06-13
Add end position to Token

### LINKED ISSUE #2245: Correctly display tokens stretching multiple lines
The Token does not have information about the end line of itsself
This can break highlighting

Image
The Token for `System` was stretching till the closing braket in this version of Jplag

 --> The Java Tokens need an end position

Kr0nox:
When tokens stretch multiple lines, the token model could not represent that.
This PR adds an end position to the token class and makes the start more explicit (fixes #2245).

Tokens now explicitly store both start and end line and column information instead of relying on a single line/column and a length value. All language modules, parsers, scanners, and reporting utilities have been updated to utilize these new properties, enabling precise tracking of multi-line and multi-column tokens.

There are some exceptions:

- SCXML does not have multi-line tokens, but uses the new format implicitly.
- EMF-metamodel-dynamic and EMF-model do not have multi-line tokens, but use the new format implicitly.
- Rust will be updated in an upcoming PR (as part of the update to our format for ANTLR-based modules).
- The TokenPrinter will be updated in an upcoming PR.

Additionally, this PR contains some minor improvements:

- Disentangle the token and view creation for the SCXML module.
- Disentangle the token and view creation for the EMF modules.
- Improve token trace conversion safety in the Java module.


Kr0nox:
In the meeting we discussed:
- TokenPrinter will be subject to another PR
- Text will remain in its current state
- Rust will be left like this, as it will be adapted when moving to our own antlr framework
- scxml and emf will be looked at later in this PR

languages/rust/src/main/java/de/jplag/rust/RustParserAdapter.java
```diff
@@ -71,7 +71,8 @@ private void parseFile(File file) throws ParsingException {
      * @param length the length of the Token
      */
     /* package-private */ void addToken(RustTokenType type, int line, int start, int length) {
-        tokens.add(new Token(type, currentFile, line, start, length));
+        // This does not cover multiline tokens. That will be solved when moving this module to the antlr wrapper
+        tokens.add(new Token(type, currentFile, line, start, line, start + length, length));
```
tsaglam: @Kr0nox it might be better to use the deprecated constructor here, so it's clear that this change has not been addressed yet for Rust.

languages/rust/src/main/java/de/jplag/rust/RustParserAdapter.java
```diff
@@ -71,6 +71,7 @@ private void parseFile(File file) throws ParsingException {
      * @param length the length of the Token
      */
     /* package-private */ void addToken(RustTokenType type, int line, int start, int length) {
+        // This does not cover multiline tokens. That will be solved when moving this module to the antlr wrapper
         tokens.add(new Token(type, currentFile, line, start, length));
```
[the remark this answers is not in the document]
tsaglam: Intended.

language-antlr-utils/src/main/java/de/jplag/antlr/TokenCollector.java
```diff
@@ -47,22 +47,24 @@ <T> void addToken(TokenType jplagType, Function<T, CodeSemantics> semanticsSuppl
         }
         org.antlr.v4.runtime.Token antlrToken = extractStartToken.apply(entity);
         org.antlr.v4.runtime.Token antlrEndToken = extractEndToken.apply(entity);
-        int line = antlrToken.getLine();
-        int column = antlrToken.getCharPositionInLine() + 1;
+        int startLine = antlrToken.getLine();
+        int startColumn = antlrToken.getCharPositionInLine() + 1;
+        int endLine = antlrEndToken.getLine();
+        int endColumn = antlrEndToken.getCharPositionInLine() + 1;
```
robinmaisch, line 53: Does this account for the length of the "end token"?

language-testutils/src/test/java/de/jplag/testutils/LanguageModuleTest.java
```diff
@@ -95,7 +96,7 @@ final void testSourceCoverage(TestData data) throws ParsingException, IOExceptio
         this.configureIgnoredLines(ignoredLines);
         List<Integer> relevantLines = new ArrayList<>(ignoredLines.getRelevantLines());
 
-        tokens.stream().map(Token::getLine).forEach(relevantLines::remove);
+        tokens.stream().flatMap(t -> IntStream.rangeClosed(t.getStartLine(), t.getEndLine()).boxed()).forEach(relevantLines::remove);
```
robinmaisch, line 99: Is `rangeClosed` what we want here? I would think that `List.of(t.getStartLine(), t.getEndLine())` is more helpful, but I might well be wrong!

language-testutils/src/test/java/de/jplag/testutils/LanguageModuleTest.java
```diff
@@ -216,17 +217,17 @@ final void testTokenPositions(TokenPositionTestData testData) throws ParsingExce
             TokenType expectedType = this.languageTokens.stream().filter(type -> type.toString().equals(expectedToken.typeName())).findFirst()
                     .orElseThrow(() -> new IOException(String.format("The token type %s does not exist.", expectedToken.typeName())));
 
-            if (extractedTokens.stream().noneMatch(token -> token.getType() == expectedType && token.getLine() == expectedToken.lineNumber()
-                    && token.getColumn() == expectedToken.columnNumber() && token.getLength() == expectedToken.length())) {
+            if (extractedTokens.stream()
+                    .noneMatch(token -> token.getType() == expectedType && token.getStartLine() == expectedToken.startLine()
+                            && token.getStartColumn() == expectedToken.startColumn() && token.getEndLine() == expectedToken.endLine()
+                            && token.getEndColumn() == expectedToken.endColumn())) {
```
robinmaisch: Before this condition gets out of hand, we may just extract a variable ;)

language-testutils/src/test/java/de/jplag/testutils/LanguageModuleTest.java
Kr0nox: Yes, we want to test, that all lines have a token in them. 
Let's say we have a token that starts on line 1 and ends on line 3. Then the token covers line 1, 2 and 3. `rangeClosed` generates exactly this range. 
Taking `List.of(t.getStartLine(), t.getEndLine())` would leave out line 2 which is covered by a token

language-testutils/src/test/java/de/jplag/testutils/LanguageModuleTest.java
Kr0nox: Would you rather just have the expression extracted into a variable above the if or have it split into an external function for the comparison

language-antlr-utils/src/main/java/de/jplag/antlr/TokenCollector.java
Kr0nox: Im not sure about maybe somebody with more antlr knowledge knows the anwer. @tsaglam @TwoOfTwelve

tsaglam: In theory, it should; however, we might need to test this.

language-testutils/src/test/java/de/jplag/testutils/LanguageModuleTest.java
robinmaisch: A function would surely be even better! There doesn't seem to be a `Token::equals` method as of now, but this would be a real use case then.

Kr0nox: Equals method will not work here, as we are not comparing a token to a token, but a token to an datatype only present in testing

Issue #2320, opened 2025-04-23
Lack in consistency in specifying supported file endings for languages


tsaglam:
Currently, some languages specify lowercase and uppercase versions of the same file endings for the supported file types.
Ideally, we ensure that JPlag ignores cases for file ending checks. Then, we can remove all duplicate file endings.
Importantly, we should check where the supported file endings are used throughout the codebase.

- [x] check file extensions independent of capitalization
- [x] remove duplicate file extensions from language modules
- [ ] Remove inconsistencies regarding the terminology in the codebase

robinmaisch:
While we are on the topic of consistency, we may aim for consistent _naming_ in identifiers and documentation. Currently, language modules define "suffixes" and sometimes "file endings," and according to the Markdown documentation, the report files now also have a new "file ending."

It appears that "file ending" is a Germanism for "file extension" or "filename extension" (https://en.wikipedia.org/wiki/Filename_extension). If we restrict ourselves to proper _file extensions_ instead of the more general concept of a _suffix_, we can utilize this when checking the submission files for the specific file extensions we are looking for. Currently, all suffixes of supported language modules are actually proper file extensions.

However, if we want to/must allow suffixes other than a "dot + extension" for specific language modules, then we should call them "suffixes" consistently.

tsaglam:
Currently, we truly allow suffixes, meaning you can check if a file ends in any given string. Especially due to the `-p` and `--suffixes` options, this is also user-facing. Regarding terminology, I would say we can broadly move to file extension in our codebase. However, I would not change the CLI parameter itself. For the public API, we probably need to work with deprecation (JPlagOptions, Language interface, etc.)

PR #2439, opened 2025-06-16
Enforce consistent terminology regarding file extensions

tsaglam:
Address remaining points in #2320.

- TODO

### DIFF
--- a/language-api/src/main/java/de/jplag/Language.java
+++ b/language-api/src/main/java/de/jplag/Language.java
@@ -13,7 +13,7 @@ public interface Language {
 
     /**
      * File extensions for the files containing code of the language. All capitalization variants of extensions are matched.
-     * An empty array means all suffixes are valid.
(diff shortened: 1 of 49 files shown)


language-api/src/main/java/de/jplag/Language.java
```diff
@@ -13,9 +13,19 @@ public interface Language {
 
     /**
      * File extensions for the files containing code of the language. All capitalization variants of extensions are matched.
-     * An empty array means all suffixes are valid.
+     * An empty array means all extensions are valid.
+     * @deprecated see {@link Language#fileExtensions()}
      */
-    String[] suffixes();
+    @Deprecated(since = "7.0.0", forRemoval = true)
```
Kr0nox: we should be consistent with the deprecation versions. In other PRs we entered 6.2.0 here it is 7.0.0

PR #2452, opened 2025-06-23
Add Boxplot

Kr0nox:
Adds a boxplot as a secondary way of displaying the distribution of comparisons
Bildschirmfoto am 2025-06-23 um 19 24 57


tsaglam (changes requested):
Can you reduce the width of the box to roughly 1/3 of the current one?

Kr0nox:
> Can you reduce the width of the box to roughly 1/3 of the current one?

I now set it to a static smaller width. If we want the width to be dynamic based on the container size I can also do that

tsaglam:
> I now set it to a static smaller width. If we want the width to be dynamic based on the container size I can also do that

I think static is fine...

tsaglam (commented):
I tested the feature with a few datasets. Looks great. Just one minor thing: In the light theme, the lower whisker blends with the x-axis line. If it is easy to tweak, maybe slightly increase the line width of the whisker.

tsaglam:
> I can only increase it for the entire box

Looks fine, let us go with this!

PR #2461, opened 2025-06-26
Fix bug that causes crashes for min token match values above 25

tsaglam:
Fixes a bug introduced in #534 that leads to crashes when the minimum token match exceeds the value 25. The reason was that the class `SubsequenceHashLookupTable` incorrectly clamped the window size, using the unclamped value for the initialization of the array of subsequence hashes. This was caused by omitting the `this` keyword.

As a result, the code now reliably supports larger minimum token match values. The change also adds a test to verify correct behavior for edge cases and improves overall code stability when handling larger window sizes.


robinmaisch (approved):
LGTM

PR #2463, opened 2025-06-27
Adjust dev-demo path

sebinside:
We moved the dev demo from https://jplag.github.io/DevDemo/ to https://devdemo.jplag.de/.
This PR updates the vite config that used in the github action to reflect these changes.

### DIFF
--- a/report-viewer/vite.config.ts
+++ b/report-viewer/vite.config.ts
@@ -19,7 +19,7 @@ export default defineConfig((userConfig: UserConfig) => {
       base = '/Demo/'
       break
     case 'dev-demo':
-      base = '/DevDemo/'
+      base = '/'
       break
   }
   return {


report-viewer/vite.config.ts
```diff
@@ -19,7 +19,7 @@ export default defineConfig((userConfig: UserConfig) => {
       base = '/Demo/'
       break
     case 'dev-demo':
-      base = '/DevDemo/'
```
Kr0nox, line 22: this does not change the base variable. do we want this to be an explicit case?

sebinside: Good point, I should have made this more explicit: I guessed its better to have it explicit until we moved from using github.io. paths completely (in an ideal world, including the new website then). Until then, having more explicit information in this switch is probably no bad idea. Afterwards, we could ideally remove the switch completely (or at least to a large extent).

Kr0nox: Then we can probably set the sonar issue (its the same thing) as accepted issue and come back to it in the future

Kr0nox (approved):

PR #2469, opened 2025-07-01
Added cli option for specifying the submission encoding

### LINKED ISSUE #2398: please provide an option to specify file encoding, don't  detect file encoding
Could you please provide an option to read files using a specified encoding, rather than relying on guesswork? Guessing often leads to inconsistencies and confusion. for example, using an --encoding to specify the file encoding.
When using JPlag, these files are all encoded in UTF-8, it reports some files (not all) have encoding issues, but these files are also encoded in UTF-8. There are no garbled characters when opening them with UTF-8 in VS Code. I don't understand why it would guess that these files are encoded in Windows-1252.
```plain
t81.java:8: 错误: 编码 windows-1252 的不可映射字符 (0x8F)
        return "?–????????????‘?”?é”?è??: " + e.getMessage();
```

[t81.txt](https://github.com/user-attachments/files/20437441/t81.txt)

please rename to t81.java

TwoOfTwelve:
Resolves #2398 by adding a new cli option that manually specifies the charset used for submissions

### DIFF
--- a/cli/src/main/java/de/jplag/cli/options/CharsetCandidates.java
+++ b/cli/src/main/java/de/jplag/cli/options/CharsetCandidates.java
@@ -5,7 +5,13 @@ import java.util.List;
 
 import com.ibm.icu.text.CharsetDetector;
 
+/**
+ * Provides a list of charsets for picocli
+ */
 public class CharsetCandidates extends ArrayList<String> {
+    /**
+     * New instance
+     */
     public CharsetCandidates() {
         super(List.of(CharsetDetector.getAllDetectableCharsets()));
     }
--- a/cli/src/main/java/de/jplag/cli/options/CliOptions.java
+++ b/cli/src/main/java/de/jplag/cli/options/CliOptions.java
@@ -109,7 +109,7 @@ public class CliOptions implements Runnable {
         @Option(names = "--log-level", description = "Set the log level for the cli.")
         public Level logLevel = Level.INFO;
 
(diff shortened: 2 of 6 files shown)


cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -107,6 +108,9 @@ public static class Advanced {
 
         @Option(names = "--log-level", description = "Set the log level for the cli.")
         public Level logLevel = Level.INFO;
+
+        @Option(names = "--submission-charset", description = "Specifies a charset to be used for submissions. This disables the automatic charset detection for submissions.", completionCandidates = CharsetCandidates.class, converter = CharsetConverter.class)
```
```suggestion
        @Option(names = "--encoding", description = "Specifies the charset of the submissions. This disables the automatic charset detection", completionCandidates = CharsetCandidates.class, converter = CharsetConverter.class)
```

I would be in favor of a shorter CLI parameter, e.g. `encoding` or similar.

cli/src/main/java/de/jplag/cli/options/CharsetCandidates.java
```diff
@@ -0,0 +1,12 @@
+package de.jplag.cli.options;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.ibm.icu.text.CharsetDetector;
+
+public class CharsetCandidates extends ArrayList<String> {
```
tsaglam, line 11: Remember to add JDoc.

tsaglam (commented):
Just a few minor things.

PR #2478, opened 2025-07-08
Fix merging matches across files

tsaglam:
This PR fixes the logic in the subsequence match merging mechanisms that prevents merging across multiple files. Previously, the merging process did not correctly handle cases where the last token between two neighboring matches is an end-of-file (EOF) token. This bug is most likely in the codebase since match merging was introduced, but only arises rarely as it is an off-by-one error.

### DIFF
--- a/core/src/test/java/de/jplag/merging/MatchMergingTest.java
+++ b/core/src/test/java/de/jplag/merging/MatchMergingTest.java
@@ -3,7 +3,9 @@ package de.jplag.merging;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertIterableEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
+import java.io.File;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


core/src/test/java/de/jplag/merging/MatchMergingTest.java
```diff
@@ -187,6 +190,54 @@ void testMinimalRequiredMerges() throws ExitException {
...
+        JPlagOptions options = getDefaultOptions("crossFile").withMergingOptions(mergingOptions);
+        SubmissionSetBuilder builder = new SubmissionSetBuilder(options);
+        SubmissionSet submissionSet = builder.buildSubmissionSet();
+        LongestCommonSubsequenceSearch comparisonStrategy = new LongestCommonSubsequenceSearch(options);
+        JPlagResult result = comparisonStrategy.compareSubmissions(submissionSet);
+
+        assumeTrue(2 == result.getNumberOfSubmissions());
+        assumeTrue(1 == result.getAllComparisons().size());
+        JPlagComparison comparison = result.getAllComparisons().getFirst();
+        assumeTrue(2 == comparison.matches().size());
```
uuqjz: Should be assertEquals

uuqjz (changes requested):
Change the three assertTrue to assertEquals, then it is good to go.

PR #2486, opened 2025-07-09
Enable avoidance of wildcard imports for spotless

tsaglam:
Spotless will now clean up wildcard imports, which is a [recent feature](https://github.com/diffplug/spotless/pull/2517).
This PR also addresses the new spotless issues introduced by this feature.

**Important**: This does not automatically refactor imports, it just makes sure that:

- _check_ fails if wildcard imports are used.
- _apply_ removes wildcard imports.


dfuchss:
Cool feature .. I need to adapt some templates of mine :D

tsaglam:
At first glance, I see no option of only having `removeWildcardImports` only for `spotless:check` but not for `spotless:apply`. @dfuchss do you know it that is possible without too much of a hassle?

dfuchss:
> At first glance, I see no option of only having `removeWildcardImports` only for `spotless:check` but not for `spotless:apply`. @dfuchss do you know it that is possible without too much of a hassle?

You probably need to define it in the profiles (i.e., you run spotless:check and spotless:apply with different profiles)

PR #2498, opened 2025-07-18
Fix some code quality issues in the clustering module

tsaglam:
This PR fixes quality issues in some of the worst offenders in the clustering module (excluding the BFGS class):

-

### DIFF
--- a/core/src/main/java/de/jplag/clustering/algorithm/AgglomerativeClustering.java
+++ b/core/src/main/java/de/jplag/clustering/algorithm/AgglomerativeClustering.java
@@ -27,7 +27,7 @@ public class AgglomerativeClustering implements GenericClusteringAlgorithm {
     public Collection<Collection<Integer>> cluster(RealMatrix similarityMatrix) {
         int size = similarityMatrix.getRowDimension();
         // all clusters that do not have a parent yet
-        Set<Cluster> clusters = new HashSet<>(size);
+        Set<Cluster> clusters = HashSet.newHashSet(size);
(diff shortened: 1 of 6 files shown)


core/src/main/java/de/jplag/clustering/algorithm/AgglomerativeClustering.java
```diff
@@ -27,7 +27,7 @@ public AgglomerativeClustering(ClusteringOptions options) {
     public Collection<Collection<Integer>> cluster(RealMatrix similarityMatrix) {
         int size = similarityMatrix.getRowDimension();
         // all clusters that do not have a parent yet
-        Set<Cluster> clusters = new HashSet<>(size);
+        Set<Cluster> clusters = HashSet.newHashSet(size);
```
uuqjz, line 30: Could you explain the semantic difference here to me?

tsaglam: `new HashSet<>(size)` does not guarantee an initial capacity of `size`, rather it creates the set with a size of `size * loadFactor` (which is 0.75 by default). Therefore, resizing is necessary to accommodate all initial elements. Increasing the load factor to guarantee the initial sizes sacrifices a useful load factor if more elements are added. `HashSet.newHashSet(size)` guarantees that the initial size is `size` without sacrificing a useful load factor.

uuqjz: Did not know that!

PR #2502, opened 2025-07-21
Added an assign token to java variable declarations if an initializer is present

TwoOfTwelve:
Adds an assign token to variable declarations with initializers.

For example:

```java
int x = 1;
| VARDEF
      | ASSIGN
```

This should be better since it is now equivalent to:

```java
int x;
| VARDEF
x = 1;
  | ASSIGN
```

To make this compatible with the new pattern syntax, an assignment is also added there:

```java
public class PatternMatchingManual {
    private static final record Test(int x) {
    }

    public void test() {
        Object a = new Test(1);
        if (a instanceof Test testA) {
                         | VARDEF
                                  | ASSIGN
        }
    }
}
```

Required for #1962

### DIFF
--- a/languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java
+++ b/languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java
@@ -13,6 +13,7 @@ import de.jplag.semantics.VariableScope;
 import com.sun.source.tree.AnnotationTree;
 import com.sun.source.tree.AssertTree;
 import com.sun.source.tree.AssignmentTree;
+import com.sun.source.tree.BindingPatternTree;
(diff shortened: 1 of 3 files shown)


tsaglam:
For PRs like this one that fundamentally change tokenization, we should ensure there is a detailed PR description including examples of how the tokenization changes.
Moreover, we must ensure that this change has no adverse side effects. I will have to do some testing with larger datasets. Perhaps you could include the summary metric results of the E2E tests regarding the shift for plagiarism pairs and unrelated pairs in the PR description.

tsaglam (changes requested):
Reminder for @jplag/maintainer: Do not merge before internal impact testing with real-world data.

Issue #2464, opened 2025-06-27
Code highlighting per file


Kr0nox:
Due to the multi language module each file could be a different language and thus the report viewer should hightlight each file accordingly. 
For that the submissionFileIndex needs to be expanded by an option to specify the language of a file. 
The Parsing result for each file needs to give the found language

PR #2511, opened 2025-07-24
Dissolve the shared parser superclass

tsaglam:
Historically, all concrete parsers of the individual language module all had a shared superclass `AbstractParser`.
Over time, more and more functionality was moved away from that superclass, making it superfluous.
Its only functionality was to hold a shared logger instance, which is not required.

Thus, this PR dissolves the shared class. As a consequence, parsers do not share a common superclass, but there is also no need for it.

```mermaid

classDiagram
    class AbstractParser {
        + logger: slf4j.Logger
    }

    AbstractParser <|-- JavaParser
    AbstractParser <|-- CppParser
    AbstractParser <|-- RustParser
	AbstractParser <|-- OtherParser

    %% After removal
    class JavaParser 
    class CppParser
    class RustParser
    class OtherParser["..."]

```


Kr0nox:
We should keep in mind that in the future we might want a shared method in the parsers due to #2464 needing the language for each file from the parser. So we should look at possible implementations for that and check whether a shared parser super class might be beneficial for that.

PR #2514, opened 2025-07-24
Rewrite of the scala module in java

TwoOfTwelve:
This PR contains a complete rewrite of the Scala module. It is still based on the same library, so there should be no changes in the behavior.

To call the Scala method from java a few lines are a bit convoluted. Working with the Scala data-structures in Java is also a bit clunky, but I think it's worth it to not be dependent on a Scala compiler anymore.

The tests have also been updated to the new framework.


languages/scala/src/main/scala/de/jplag/scala/TR.java
```diff
@@ -0,0 +1,82 @@
...
+import scala.meta.Tree;
+
+/**
+ * A token record, that contains information about the tokens to be added for a specific ast node. The name is kept
+ * short to make the code more readable. Also contains information about how to traverse the node
+ * @param before Token before the node
+ * @param after Token after the ode
+ * @param traverse The traverse function
+ */
+public record TR(Optional<ScalaTokenType> before, Optional<ScalaTokenType> after, BiConsumer<Tree, ScalaParser> traverse) {
```
tsaglam: As discussed, we probably want to rename this.

report-viewer/package-lock.json
```diff
@@ -7454,7 +7448,7 @@
       "version": "5.8.3",
       "resolved": "https://registry.npmjs.org/typescript/-/typescript-5.8.3.tgz",
       "integrity": "sha512-p1diW6TqL9L07nNxvRMM7hMMw4c5XOo/1ibL4aAIGmSAt9slTE1Xgw5KWuof2uTOvCg9BY7ZRi+GaF+7sfgPeQ==",
-      "devOptional": true,
+      "dev": true,
       "license": "Apache-2.0",
       "bin": {
```
tsaglam: Should probably not be part of this PR.

Issue #2467, opened 2025-06-30
Sonarcloud GitHub action is deprecated


tsaglam:
The [sonarcloud-github-action](https://github.com/SonarSource/sonarcloud-github-action) we use for the report viewer is deprecated.
We should use the [sonarqube-scan-action](https://github.com/marketplace/actions/official-sonarqube-scan) instead.

At first glance, this only affects:

https://github.com/jplag/JPlag/blob/794633dd5e10723a5022179974bcf938bec587d6/.github/workflows/report-viewer-sonarcloud.yml#L21-L35

PR #2518, opened 2025-07-29
Switch to new sonarcloud workflow implementation

TwoOfTwelve:
This PR updates the SonarCloud workflow to use a new action implementation by switching from the deprecated SonarCloud GitHub action to the newer SonarQube scan action with a specific version.

See also: #2467


robinmaisch (approved):
Should be good to go 👍

PR #2536, opened 2025-07-31
Bump vite to v7

Kr0nox:
replaces #2523, #2525


Kr0nox:
This PR will work after #2537 since vite needs a newer node.js version

PR #2542, opened 2025-08-05
Comparing comments from submissions

c0derMo:
This PR is a follow up to #2426. This PR adds the option to use the previously extracted comments and compare them between submissions.
This is done by first running all the comments through the builtin `text` language module, to tokenize the comments, and then using the existing `GreedyStringTiling` algorithm to find common subsequences of tokens between the comments of submissions.
The original `GreedyStringTiling` and `TokenSequenceMapper` classes had to be slightly modified in order to work with different token lists than just the "code token list" of a submission.
To ensure that the comment comparison can only ever increase the similarity of two submissions, the comment tokens are only considered in the similarity calculation if they are part of a match.

### DIFF
--- a/core/src/main/java/de/jplag/JPlag.java
+++ b/core/src/main/java/de/jplag/JPlag.java
@@ -99,13 +99,13 @@ public class JPlag {
         if (options.mergingOptions().enabled()) {
             result = new MatchMerging(options).mergeMatchesOf(result);
         }
-      
+
         // Compare comments
         if (options.analyzeComments()) {
             CommentComparer commentComparer = new CommentComparer(options);
             result = commentComparer.compareCommentsAndMergeMatches(result);
         }
-      
+
         if (options.frequencyAnalysisOptions().enabled()) {
             MatchWeighting matchWeighter = new MatchWeighting(options.frequencyAnalysisOptions());
             List<JPlagComparison> frequencyWeightedComparisons = matchWeighter.useMatchFrequencyToInfluenceSimilarity(result);


core/src/main/java/de/jplag/commenthandling/CommentComparer.java
```diff
@@ -0,0 +1,158 @@
...
+                JPlagComparison newBaseCodeComparison = new JPlagComparison(baseCodeCommentComparison.firstSubmission(),
+                        baseCodeCommentComparison.secondSubmission(), Collections.emptyList(), Collections.emptyList(),
+                        baseCodeCommentComparison.matches());
+                currentSubmission.setBaseCodeComparison(newBaseCodeComparison);
+            }
+        }
+    }
+
+    private List<Match> flipMatches(List<Match> matches) {
+        return matches.stream().map(match -> new Match(match.startOfSecond(), match.startOfFirst(), match.lengthOfFirst(), match.lengthOfSecond()))
```
Kr0nox: since #2438 we do not identify tokens by their start and length, but by their start and end. So you probably need to merge develop into your branch.

In addition you only flipped the start and not the length

report-viewer/src/model/Match.ts
```diff
@@ -24,6 +24,7 @@ export interface ReportFormatMatch {
   endInSecond: CodePosition
   lengthOfFirst: number
   lengthOfSecond: number
+  isComment?: boolean
```
Kr0nox: just use a proper boolean and not a posibly undefined

report-viewer/src/model/factories/ComparisonFactory.ts
```diff
@@ -45,6 +45,11 @@ export class ComparisonFactory extends BaseFactory {
 
       fileOfFirst.matchedTokenCount += match.lengthOfFirst
       fileOfSecond.matchedTokenCount += match.lengthOfSecond
+
+      if (match.isComment) {
+        fileOfFirst.tokenCount += match.lengthOfFirst
+        fileOfSecond.tokenCount += match.lengthOfSecond
```
Kr0nox, line 82: similar to my general comment, I feel it is wrong to only add the length of matched comments. This leads to the same submission having different total token counts, which should not be the case and might confuse the user

Kr0nox (commented):
In general i do not feel its right to only include the tokens of matched comments in the similarity calculation. The base should include the tokens from all comments. Otherwise the similarity can get scewed towards the comments.
If a person activates the comment option, I think they can reasonably expect all comments to be included in the comparison. 

Also if the tokens are included in the total, they should also be included in the similarity metric. I think you did not change that

Kr0nox:
> The issue with that is that plagiarisms could be hidden more easily, simply by adding a lot of unnecessary, non-matching comments

This would be similar to adding a lot of unnecessary code, like Mossad does. But this is the reason we have multiple similarity metrics like the Maximum Similarity or Maximum Code Length, which would catch such submissions.
So I think this should not be a problem, and we should include all tokens, to properly reflect what was parsed and compared

core/pom.xml
```diff
@@ -44,6 +44,11 @@
             <version>47</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
TwoOfTwelve, line 36: @jplag/maintainer do we want to look for a way to avoid this dependency long term?

tsaglam: No, this is not possible, as one of the fundamental concepts of this PR is re-using the text module. Other NLP measures for text similarity did not perform well enough due to their large number of false positives.

core/src/main/java/de/jplag/reporting/reportobject/model/Match.java
```diff
@@ -10,7 +10,32 @@
...
+    }
+
+    /**
+     * Transforms this match into a comment match.
+     * @return Copy of the match representing a comment match.
+     */
+    public Match asComment() {
+        return new Match(this.firstFileName, this.secondFileName, this.startInFirst, this.endInFirst, this.startInSecond, this.endInSecond,
+                this.lengthOfFirst, this.lengthOfSecond, true);
+    }
```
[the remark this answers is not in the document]
c0derMo: False positive

core/src/main/java/de/jplag/commenthandling/CommentComparer.java
```diff
@@ -0,0 +1,158 @@
...
+                commentComparisons.put(future.get().getKey(), future.get().getValue());
+            }
+        } catch (InterruptedException | ExecutionException e) {
+            Thread.currentThread().interrupt();
+            throw new ComparisonException("Error during comment comparison algorithm.", e);
+        } finally {
+            progressBar.dispose();
+        }
+
+        List<JPlagComparison> fixedComparisons = new ArrayList<>();
```
robinmaisch: This section looks like a `map`ping operation to me; maybe we can make that more explicit.
The existing comparisons are _extended_ through the commentComparisons rather than `fixed`, right?
Please also include a brief comment on what will happen next.

core/src/main/java/de/jplag/commenthandling/CommentPreprocessor.java
```diff
@@ -0,0 +1,78 @@
...
+        this.comments = comments;
+        this.comments.sort(Comparator.comparing(Comment::file));
+        this.textParser = new ParserAdapter();
+    }
+
+    /**
+     * Processes all input comments into a list of tokens.
+     * @return List of tokens containing all comments
+     */
+    public List<Token> processToToken() {
```
robinmaisch: This method looks like it should be `static`, getting the `Comment`s as an argument instead of the instance state. This way, the same `CommentPreprocessor` instance can be reused across all `Submission`s.
It appears that the `ParserAdapter` instance can also be reused, despite having a state.

languages/text/src/main/java/de/jplag/text/ParserAdapter.java
```diff
@@ -57,17 +57,33 @@ public List<Token> parse(Set<File> files) throws ParsingException {
...
+    public List<Token> parseStrings(Set<String> fileContents) {
+        tokens = new ArrayList<>();
+        for (String fileContent : fileContents) {
+            logger.trace("Parsing next string file content");
+            this.currentFile = null;
+            parseFile(fileContent);
+            tokens.add(Token.fileEnd(null));
+        }
+        return tokens;
+    }
```
robinmaisch, line 81: I think it is fair to assume that the `String`s given here as an argument do not represent one file each. Do you need the fileEnd token here? Currently, these are filtered out later in `CommentPreprocessor::fixTokenPositions`.

core/src/main/java/de/jplag/commenthandling/CommentPreprocessor.java
```diff
@@ -0,0 +1,78 @@
...
+            int startLine = comment.line() + token.getStartLine() - 1;
+            int endLine = comment.line() + token.getEndLine() - 1;
+            int startColumn = token.getStartColumn();
+            int endColumn = token.getEndColumn();
+            if (token.getStartLine() == 1) {
+                startColumn += comment.column() - 1;
+            }
+            if (token.getEndLine() == 1) {
+                endColumn += comment.column() - 1;
+            }
```
robinmaisch, line 64: Is there a test case that covers this edge case?

core/src/main/java/de/jplag/Submission.java
```diff
@@ -284,11 +292,14 @@ private static File createErrorDirectory(String... subdirectoryNames) {
     private void extractAndParseComments() {
         Optional<CommentExtractorSettings> commentExtractorSettings = language.getCommentExtractorSettings();
         if (commentExtractorSettings.isPresent()) {
+            List<Comment> rawComments = new ArrayList<>();
             for (File file : files) {
                 CommentExtractor extractor = new CommentExtractor(file, commentExtractorSettings.get());
-                comments.addAll(extractor.extract());
+                rawComments.addAll(extractor.extract());
```
robinmaisch, line 300: Also here, the `CommentExtractor` is a throwaway object. Can we initialize it once with the settings and use it multiple times?

core/src/main/java/de/jplag/commenthandling/CommentPreprocessor.java
c0derMo: Yes, as this is triggered for each single-line comment and the first line of multi-line comments

core/src/main/java/de/jplag/Submission.java
c0derMo: The comment extractor does have quite a bit of internal state, which would need to be transformed in some way to make it reusable, which is why I would prefer to keep this as as throwaway object

PR #2549, opened 2025-08-09
CLI option to skip version check

b0ink:
We use JPlag in an automated docker environment and our institution has a strict firewall that is blocking outgoing API calls to GitHub, causing JPlag to hang for about 2 minutes when attempting to fetch the latest version on every run

We've been able to get around this by setting `network_mode: none` on our JPlag containers which removes internet access entirely, but there's still a brief delay before it times out and proceeds with the analysis


cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -114,6 +114,9 @@ public static class Advanced {
 
         @Option(names = "--encoding", description = "Specifies the charset of the submissions. This disables the automatic charset detection", completionCandidates = CharsetCandidates.class, converter = CharsetConverter.class)
         public Charset submissionCharsetOverride;
+
+        @Option(names = "--skip-version-check", description = "Skip fetching latest version information from the API.")
```
Kr0nox: Please implement the flag as [hidden](https://picocli.info/#_hidden_options_and_parameters), so it does not clutter the help text

robinmaisch:
@b0ink Please resolve the indicated conflict, we do not have sufficient rights for that. The merging should be easy.
After that, we should finally be ready to incorporate this PR.

Kr0nox (approved):

report-viewer/model/src/CliOptions.ts
```diff
@@ -21,6 +21,7 @@ interface AbstractOptions<L> {
   mergingOptions: CliMergingOptions
   normalize: boolean
   analyzeComments: boolean
+  skipVersionCheck: boolean
```
dfuchss: Do we need to include this option in the report viewer ? I'd argue it does not change the behavior regarding plagiarism checks . @jplag/studdev do we need to have all options in the report viewer . Iff not, I'd not include it here :)

core/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -49,15 +49,16 @@
...
 @RecordBuilder()
 public record JPlagOptions(@JsonSerialize(using = LanguageSerializer.class) Language language, Integer minimumTokenMatch,
         @JsonSerialize(contentUsing = FileSerializer.class) Set<File> submissionDirectories,
         @JsonSerialize(contentUsing = FileSerializer.class) Set<File> oldSubmissionDirectories,
         @JsonSerialize(using = FileSerializer.class) File baseCodeSubmissionDirectory, String subdirectoryName, List<String> fileSuffixes,
         String exclusionFileName, SimilarityMetric similarityMetric, double similarityThreshold, int maximumNumberOfComparisons,
-        ClusteringOptions clusteringOptions, boolean debugParser, MergingOptions mergingOptions, boolean normalize, boolean analyzeComments)
-        implements JPlagOptionsBuilder.With {
+        ClusteringOptions clusteringOptions, boolean debugParser, MergingOptions mergingOptions, boolean normalize, boolean analyzeComments,
+        boolean skipVersionCheck) implements JPlagOptionsBuilder.With {
```
dfuchss: I'm unsure whether I would add this to the JPlag options , I'd argue it would be enough in the cli parser as it does not affect the core but only the cli.

@tsaglam @robinmaisch what do you think ?

dfuchss (commented):
Thank you for your contribution! From my side, I have one minor point that should be discussed by the maintainers before merging :) otherwise the PR looks good to me.

report-viewer/model/src/CliOptions.ts
Kr0nox: Yes, all options are in the report viewer
But relating to your other other comment: If the field is removed there, it will also need to be removed here, as we are writing the JPlagOptions

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -53,22 +53,32 @@ public CLI(String[] args) {
...
-                case RUN_AND_VIEW -> runAndView();
-                case AUTO -> selectModeAutomatically();
-            }
+        // check version regardless of parsing result
+        if (!this.inputHandler.getCliOptions().advanced.skipVersionCheck) {
+            JPlagVersionChecker.printVersionNotification();
+        }
+
+        if (shouldAbortRunNow) {
+            // help text has been printed, do nothing else
```
TwoOfTwelve, line 65: I'd say both of these comments are redundant, as they don't provide a lot more insight. The comment about help would belong in the JavaDoc of inputHandler.parse in my opinion. @jplag/maintainer what do you think?

report-viewer/model/src/CliOptions.ts
TwoOfTwelve: I tend to agree with @dfuchss. This parameter doesn't affect JPlag it only affects the cli. This is how we handle port and logLevel as well.

cli/src/main/java/de/jplag/cli/CLI.java
Kr0nox: we decided to add that in the last meeting, as we found the parse method returning a boolen and it being true when we should abort, to be quite counterintuitive

core/src/main/java/de/jplag/options/JPlagOptions.java
tsaglam: I have not had time to look at the PR, but if the version check is executed in the CLI only (as I remember it to be), then it does not make sense in the options.

robinmaisch (approved):
All remarks should be addressed now.

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -53,22 +53,32 @@ public CLI(String[] args) {
...
+            return;
         }
+
```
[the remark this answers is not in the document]
dfuchss: on purpose :)

tsaglam:
@TwoOfTwelve all good on your side?

PR #2559, opened 2025-08-20
Support file uploads and opening comparisons via postMessage

b0ink:
Allows the JPlag report viewer to run in an iframe or popup in another app and automatically load reports or open comparisons by submission IDs.

We currently bundle the JPlag report viewer within our Angular app and embed it in an iframe. Since JPlag comparisons are saved in our database and their submission IDs are known, this would allow the app to directly 'View a comparison' by displaying the iframe, sending the report via postMessage, and then opening the specific comparison.

Example HTML:
```html
<iframe
  id="jplag-iframe" 
  src="/JPlag/"
  frameborder="0"
  class="w-full h-screen overflow-hidden"
  [scrolling]="false"
></iframe>
```

Example of loading `.jplag` report files:
```ts
const iframe = document.querySelector<HTMLIFrameElement>('#jplag-iframe')
iframe?.contentWindow?.postMessage(
  {
    type: 'upload-jplag-file',
    file: response.body, // Blob file of the .jplag report
    name: 'report.jplag',
  }
)
```

Example of opening a comparison:
```ts
const iframe = document.querySelector<HTMLIFrameElement>('#jplag-iframe')
iframe?.contentWindow?.postMessage(
  {
    type: 'open-comparison',
    firstSubmissionId: 'student_1',
    secondSubmissionId: 'student_2',
  }
)
```

### DIFF
--- a/report-viewer/src/views/FileUploadView.vue
+++ b/report-viewer/src/views/FileUploadView.vue
@@ -203,8 +203,8 @@ onMounted(() => {
   window.addEventListener('message', async (event) => {
     const { type, file, name } = event.data ?? {}
     if (type === 'upload-jplag-report' && file) {
-      const zipFile = new File([file], name ?? 'report.jplag')
-      handleFile(zipFile)
+      const jplagReport = new File([file], name ?? 'report.jplag')
+      handleFile(jplagReport)
     }
   })
 })


b0ink:
@Kr0nox 
The `?file=` param seems to work with our setup which is great!

Regarding viewing comparisons, I'm unable to change Vue routes externally, meaning it's not possible to load the `overview/{submissionId1}/{submissionId2}` route in its current state, but if the comparisons view route supported both the `file` param, as well as params for `submissionId1` and `submissionId2` like you suggest, then we would be able to get away without POST requests.

Let me know if you'd like me to make those changes and open a new PR.
Otherwise, you can close this one :)

Kr0nox:
I have implemented it here #2598

PR #2563, opened 2025-08-27
Frequency analysis

heylissi:
This introduces frequency analysis, which builds upon the FrequencyDetermination. It utilizes frequency information to influence similarity calculations through different weighting functions, based on the frequency of a match in all comparisons.


cli/src/main/java/de/jplag/cli/JPlagOptionsBuilder.java
```diff
@@ -64,7 +64,9 @@ private JPlagOptions initializeJPlagOptions(Set<File> submissionDirectories, Set
                 oldSubmissionDirectories, null, this.cliOptions.advanced.subdirectory, suffixes, this.cliOptions.advanced.exclusionFileName,
                 JPlagOptions.DEFAULT_SIMILARITY_METRIC, this.cliOptions.advanced.similarityThreshold, this.cliOptions.shownComparisons,
                 clusteringOptions, this.cliOptions.advanced.debug, mergingOptions, this.cliOptions.normalize,
-                this.cliOptions.advanced.analyzeComments);
+                this.cliOptions.advanced.analyzeComments, this.cliOptions.frequencyAnalysis.frequency,
```
Kr0nox: Move all options regarding the frequency analysis into their own class similar to the ClusteringOptions

cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -250,4 +256,31 @@ public static class Merging {
     /** Absolute threshold used in preprocessing to cut off low-similarity pairs (hidden). */
     @Option(names = {"--cluster-pp-threshold"}, hidden = true)
     public double clusterPreprocessingThreshold;
+
+    /** Highlight extraction options. */
+    public static class FrequencyAnalysis {
+        /** Skip Highlight extraction options. */
+        @Option(names = {"--frequency-analysis"}, description = "if frequency calculation is used")
```
Kr0nox: Start your descriptions with a capital letter. This goes for all the ones below

core/src/main/java/de/jplag/JPlagComparison.java
```diff
@@ -27,6 +27,31 @@ public JPlagComparison(Submission firstSubmission, Submission secondSubmission,
         this.ignoredMatches = Collections.unmodifiableList(ignoredMatches);
     }
 
+    /**
+     * Similarity score of the comparison calculated in frequency analysis depending on the frequency of a match among all
+     * comparisons.
+     */
+    private static double frequencyWeightedScore = -1;
```
Kr0nox: this is static, so it will be the same for all comparisons. Thus all comparisons would have the same similarity. is that wanted?

core/src/main/java/de/jplag/Match.java
```diff
@@ -1,17 +1,81 @@
...
 /**
  * Represents two code fragments in two submissions that are structurally similar. These sections are usually identical
  * token subsequences, but can vary slightly when employing post-processing mechanisms, for example subsequence match
  * merging.
- * @param startOfFirst is the index of the first token of the match in the first submission.
- * @param startOfSecond is the index of the first token of the match in the second submission.
- * @param lengthOfFirst is the length of these similar sections (number of tokens) in the first submission.
- * @param lengthOfSecond is the length of these similar sections (number of tokens) in the second submission.
  */
-public record Match(int startOfFirst, int startOfSecond, int lengthOfFirst, int lengthOfSecond) {
```
Kr0nox, line 12: Please revert this back to a record. Also keep the old names without the `get` prefix

core/src/main/java/de/jplag/Match.java
```diff
@@ -61,4 +125,25 @@ public int minimumLength() {
     public int length() {
         return minimumLength();
     }
+
+    /**
+     * @param o Match Object to compare
+     * @return if there equal
+     */
+    @Override
+    public boolean equals(Object o) {
```
Kr0nox: are the equals method and hashCode needed? if not please remove them.
In addition, I am not sure, but they may be equal to a records default equal. @robinmaisch do you know that?

cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -170,6 +177,33 @@ public static class ClusteringEnabled {
...
 
+    /** Highlight extraction options. */
+    public static class FrequencyAnalysis {
+        /** Skip Highlight extraction options. */
+        @Option(names = {"--frequency-analysis"}, description = "If frequency calculation is used")
+        public boolean frequency = new FrequencyAnalysisOptions().frequency();
+
+        /** Frequency Determination strategy. */
+        @Option(names = {
+                "--frequency-strategy"}, description = "Strategy for frequency Analysis, Options: completeMatches, containedMatches, subMatches, windowOfMatches")
```
```suggestion
                "--frequency-strategy"}, description = "Strategy for frequency Analysis, one of: ${COMPLETION-CANDIDATES} (default: ${DEFAULT_VALUE}).")
```

cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -170,6 +177,33 @@ public static class ClusteringEnabled {
...
+        public FrequencyStrategies frequencyStrategy = new FrequencyAnalysisOptions().frequencyStrategy();
+
+        /** Min value for considered subsequence length in Frequency Determination strategy. */
+        @Option(names = {
+                "--frequency-min-value"}, description = "Max of min match length that will be compared and this value, is min size of considered submatches")
+        public int frequencyStrategyMinValue = new FrequencyAnalysisOptions().frequencyStrategyMinValue();
+
+        /** Weighting function to combine with frequency Determination strategy. */
+        @Option(names = {"--weighting-strategy"}, description = "Strategy for frequency Weighting, Options: PROPORTIONAL, LINEAR, QUADRATIC, SIGMOID")
```
```suggestion
        @Option(names = {"--weighting-strategy"}, description = "Strategy for frequency Weighting, one of: ${COMPLETION-CANDIDATES} (default: ${DEFAULT_VALUE}).)
```

cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -170,6 +177,33 @@ public static class ClusteringEnabled {
...
+        public WeightingStrategies weightingStrategy = new FrequencyAnalysisOptions().weightingStrategy();
+
+        /** How strong the weighting maximal influences a match length with up to double the length. */
+        @Option(names = {
+                "--weighting-factor"}, description = "Factor on how strong the weighting will be considered, scale factor for max stretch of a token sequence")
+        public double weightingStrategyWeightingFactor = new FrequencyAnalysisOptions().weightingFactor();
```
tsaglam: I would make `--frequency-min-value` and `--weighting-factor` hidden options. See other hidden options in this class for how it works.

core/src/main/java/de/jplag/JPlag.java
```diff
@@ -92,6 +97,20 @@ public static JPlagResult run(JPlagOptions options) throws ExitException {
...
+        frequencyDetermination.buildFrequencyMap(result.getAllComparisons());
+        MatchWeighting matchWeighting = new MatchWeighting(strategy, frequencyDetermination.getMatchFrequencyMap());
+        matchWeighting.weightAllComparisons(result.getAllComparisons());
+        SimilarityStrategy similarityStrategy = options.weightingStrategy().getStrategy();
+        FrequencySimilarity similarity = new FrequencySimilarity(result.getAllComparisons(), similarityStrategy);
+        for (JPlagComparison comparison : result.getAllComparisons()) {
+            double score = similarity.weightedComparisonSimilarity(comparison, options.weightingStrategyWeightingFactor());
+            JPlagComparison.setFrequencyWeightedScore(score);
+            JPlagComparison.setFrequency(options.frequency());
+        }
```
tsaglam: This should be moved into a dedicated class, see MatchMerging or TokenSequenceNormalizer.

core/src/main/java/de/jplag/JPlagComparison.java
```diff
@@ -27,6 +27,31 @@ public JPlagComparison(Submission firstSubmission, Submission secondSubmission,
...
+        JPlagComparison.frequency = frequency;
+    }
+
+    /**
+     * @param score Similarity score of the comparison calculated in frequency analysis depending on the frequency of a
+     * match among all comparisons.
+     */
+    public static void setFrequencyWeightedScore(double score) {
+        frequencyWeightedScore = score;
+    }
```
tsaglam: I would prefer to use a copy constructor here, which takes a comparison and the frequencies and frequencyWeightedScore.

core/src/main/java/de/jplag/JPlagComparison.java
```diff
@@ -27,6 +27,31 @@ public JPlagComparison(Submission firstSubmission, Submission secondSubmission,
         this.ignoredMatches = Collections.unmodifiableList(ignoredMatches);
     }
 
+    /**
+     * Similarity score of the comparison calculated in frequency analysis depending on the frequency of a match among all
+     * comparisons.
+     */
+    public static double frequencyWeightedScore = -1;
```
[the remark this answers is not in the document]
tsaglam: Also, set the field value in the constructor and make the field final.

core/src/main/java/de/jplag/Match.java
```diff
@@ -1,17 +1,81 @@
...
+     * @param startOfSecond is the index of the first token of the match in the second submission.
+     * @param lengthOfFirst is the length of these similar sections (number of tokens) in the first submission.
+     * @param lengthOfSecond is the length of these similar sections (number of tokens) in the second submission.
+     */
+    public Match(int startOfFirst, int startOfSecond, int lengthOfFirst, int lengthOfSecond) {
+        this.startOfFirst = startOfFirst;
+        this.startOfSecond = startOfSecond;
+        this.lengthOfFirst = lengthOfFirst;
+        this.lengthOfSecond = lengthOfSecond;
+    }
```
tsaglam: We need to discuss whether converting the record to a class is the best solution. Also, We need to make sure the changes are API-preserving.

core/src/main/java/de/jplag/highlightextraction/CompleteMatchesStrategy.java
```diff
@@ -23,4 +25,17 @@ public void processMatchTokenTypes(List<TokenType> matchTokenTypes, Consumer<Lis
...
+     * @param match Considered match
+     * @param frequencyMap Frequency map build with processMatchTokenTypes method
+     * @param matchToken tokenType sequence of the match
+     * @return a weight for the match
+     */
+    @Override
+    public double calculateWeight(Match match, Map<List<TokenType>, Integer> frequencyMap, List<TokenType> matchToken) {
+        Integer values = frequencyMap.get(matchToken);
+        return values != null ? values : 0.0;
+    }
```
tsaglam: Is it really the weight? Or is it the frequency? or rarity? We should avoid calling multiple concepts weights...

core/src/main/java/de/jplag/highlightextraction/ContainedMatchesStrategy.java
```diff
@@ -1,15 +1,25 @@
...
 
 /**
  * Strategy that counts all occurrences of complete matches inside all complete matches and contiguous submatches from
  * the comparisons, if the submatches are longer than minSubSequenceLength.
  */
 public class ContainedMatchesStrategy implements FrequencyStrategy {
+    /**
+     * Minimum considered subsequence length.
+     */
+    int minSubSequenceLength;
```
tsaglam: Why is this not private?

core/src/main/java/de/jplag/highlightextraction/FrequencyAnalysisOptions.java
```diff
@@ -0,0 +1,63 @@
...
+ * @param frequencyStrategy the strategy used to determine the frequency of a Match
+ * @param frequencyStrategyMinValue the minimum considered size of Subsequences from matches in the frequencyStrategy
+ * @param weightingStrategy strategy used to influence the similarity based on Match frequency
+ * @param weightingFactor how strong the impact of the weightingStrategy is
+ */
+public record FrequencyAnalysisOptions(boolean frequency, FrequencyStrategies frequencyStrategy, int frequencyStrategyMinValue,
+        WeightingStrategies weightingStrategy, double weightingFactor) {
+    /**
+     * Options for Frequency Analysis.
+     * @param frequency if frequency analysis is used
```
tsaglam: the field should be called `enableFrequencyAnalysis` or `enabled` or similar.

core/src/main/java/de/jplag/highlightextraction/FrequencyAnalysisOptions.java
```diff
@@ -0,0 +1,63 @@
...
+     * Default options for frequency Analysis.
+     */
+    public FrequencyAnalysisOptions() {
+        this(false, FrequencyStrategies.COMPLETE_MATCHES, 1, WeightingStrategies.SIGMOID, 0.25);
+    }
+
+    /**
+     * @return if the frequency Analysis is used.
+     */
+    public boolean frequency() {
```
tsaglam: and this method should be called `isFrequencyAnalysisEnabled` or `isEnabled`.

core/src/main/java/de/jplag/highlightextraction/FrequencyAnalysisOptions.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public FrequencyAnalysisOptions withFrequency(boolean frequency) {
+        return new FrequencyAnalysisOptions(frequency, frequencyStrategy, frequencyStrategyMinValue, weightingStrategy, weightingFactor);
+    }
+
+    public FrequencyAnalysisOptions withFrequencyStrategy(FrequencyStrategies strategy) {
+        return new FrequencyAnalysisOptions(frequency, strategy, frequencyStrategyMinValue, weightingStrategy, weightingFactor);
+    }
+
+    public FrequencyAnalysisOptions withFrequencyStrategyMinValue(int minValue) {
```
tsaglam: Also, find a better name for minValue. There are several "minValues" in JPlag, so each concept should be named so it is clear what it actually is.

core/src/main/java/de/jplag/highlightextraction/FrequencySimilarity.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.Objects;
+
+import de.jplag.JPlagComparison;
+import de.jplag.Match;
+
+/**
+ * Calculates the frequency dependent similarity for the comparisons according to the frequency similarity weighting
+ * strategy.
+ */
+public class FrequencySimilarity {
```
tsaglam: Again, naming.

core/src/main/java/de/jplag/highlightextraction/FrequencySimilarity.java
```diff
@@ -0,0 +1,131 @@
...
+     * Sorts the comparisons, according to the frequency.
+     * @param comparisons considered comparisons to calculate the similarity score for
+     * @param weight weighting factor, is factor for the (max) influence of the frequency
+     * @return the comparisons sorted with similarity score
+     */
+
+    public List<JPlagComparison> calculateFrequencySimilarity(List<JPlagComparison> comparisons, double weight) {
+        final double frequencyWeight = weight;
+        this.comparisons = comparisons.stream()
+                .sorted(Comparator.comparingDouble((JPlagComparison c) -> frequencySimilarity(c, frequencyWeight)).reversed()).toList();
```
tsaglam: Avoid abbreviations besides i, j, x, y, it...

core/src/main/java/de/jplag/highlightextraction/FrequencySimilarity.java
```diff
@@ -0,0 +1,131 @@
...
+            minWeight = 0.01;
+            maxWeight = 2.0;
+        } else {
+            minWeight = 1.0;
+            maxWeight = 2.0;
+        }
+
+        double finalMaxFrequency = maxFrequency;
+        double weightedSum = comparison.matches().stream().mapToDouble(match -> {
+            double freq = match.getFrequencyWeight();
```
tsaglam: No abbreviations!

core/src/main/java/de/jplag/highlightextraction/FrequencySimilarity.java
```diff
@@ -0,0 +1,131 @@
...
+                } else {
+                    return match.lengthOfSecond();
+                }
+            }
+            if (Double.isNaN(freq) || freq < 0.0)
+                freq = 0.0;
+            double rarity = 1.0 - Math.min(freq / finalMaxFrequency, 1.0);
+            double rarityWeight = strategy.computeWeight(minWeight, maxWeight, rarity);
+            double myWeight = (1 - weight) + weight * rarityWeight;
+
```
tsaglam: Please improve variable naming!

core/src/main/java/de/jplag/highlightextraction/MatchWeighting.java
```diff
@@ -0,0 +1,71 @@
...
+
+import de.jplag.JPlagComparison;
+import de.jplag.Match;
+import de.jplag.Token;
+import de.jplag.TokenType;
+
+/**
+ * Calculates frequency values and writes them into the matches.
+ */
+public class MatchWeighting {
```
tsaglam: Should be named `MatchWeightCalculator` or similarly.

core/src/main/java/de/jplag/highlightextraction/SimilarityStrategy.java
```diff
@@ -0,0 +1,15 @@
+package de.jplag.highlightextraction;
+
+/**
+ * Strategy for calculating the weight of a match.
+ */
+public interface SimilarityStrategy {
```
tsaglam: Why is this called a similarity strategy? They are not about similarities, right? Rather about weighting? Also, the name is too generic in the context of JPlag. Maybe `MatchWeightingFunction` or similar?

core/src/main/java/de/jplag/highlightextraction/FrequencyStrategies.java
```diff
@@ -1,7 +1,7 @@
 package de.jplag.highlightextraction;
 
 /**
- * Enum representing the different strategies for frequency calculation of token subsequences in comparisons.
+ * Enum representing the different strategies for frequency similarity calculation.
  */
 public enum FrequencyStrategies {
```
tsaglam: Enum names should not be in plural. Maybe name it `FrequencyAnalysisStrategy`.

core/src/main/java/de/jplag/highlightextraction/SubMatchesStrategy.java
```diff
@@ -21,10 +28,32 @@ public class SubMatchesStrategy implements FrequencyStrategy {
...
+     * @param matchToken tokenType sequence of the match
+     * @return a weight for the match
+     */
+    @Override
+    public double calculateWeight(Match match, Map<List<TokenType>, Integer> frequencyMap, List<TokenType> matchToken) {
+        List<List<TokenType>> keys = SubSequenceUtil.getSubSequences(matchToken, minSubSequenceLength);
+        SubSequenceUtil.getSubSequences(matchToken, minSubSequenceLength);
+        List<Integer> frequencies = new ArrayList<>();
+        for (List<TokenType> key : keys) {
+            int freq = frequencyMap.getOrDefault(key, 0);
```
tsaglam: Naming.

core/src/main/java/de/jplag/highlightextraction/WeightingStrategies.java
```diff
@@ -0,0 +1,28 @@
+package de.jplag.highlightextraction;
+
+/**
+ * This class contains the possible weighting functions for a match, in the frequency analysis.
+ */
+public enum WeightingStrategies {
```
tsaglam: Merge the enum and the subclasses passed to the enums, please. No reason to have this duplication/separation. Still consider my naming proposal though.

core/src/test/java/de/jplag/highlightextraction/frequencysimilarity/FrequencySimilarityTest.java
```diff
@@ -0,0 +1,186 @@
...
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import de.jplag.*;
+import de.jplag.comparison.LongestCommonSubsequenceSearch;
+import de.jplag.exceptions.ExitException;
+import de.jplag.highlightextraction.*;
+import de.jplag.options.JPlagOptions;
+
+public class FrequencySimilarityTest extends TestBase {
```
tsaglam: Since a recent PR we also require JDoc in test code.

core/src/test/java/de/jplag/highlightextraction/frequencysimilarity/FrequencySimilarityTest.java
```diff
@@ -0,0 +1,186 @@
...
+        // List<JPlagComparison> result8 = similaritySigmoid.calculateFrequencySimilarity(comparisons, 0.5);
+        // assertEquals(0.0, result5.get(1).similarity(), 0.01, "proportional frequencySimilarity should have following weight
+        // at 0.5: 1");
+        // assertEquals(0.0, result6.get(1).similarity(), 0.01, "linear frequencySimilarity should have following weight at 0.5:
+        // 1");
+        // assertEquals(0.0, result7.get(1).similarity(), 0.01, "quadratic frequencySimilarity should have following weight at
+        // 0.5: 1");
+        // assertEquals(0.0, result8.get(1).similarity(), 0.01, "sigmoid frequencySimilarity should have following weight at
+        // 0.5: 1");
+
```
tsaglam: Remove out-commented code, please. There's no reason to commit something like this.

docs/1.-How-to-Use-JPlag.md
```diff
@@ -106,6 +106,7 @@ Subsequence Match Merging
       --required-merges=<minimumRequiredMerges>
                         Minimal required merges for the merging to be applied
                           (between 1 and 50, default: 6).
+
```
tsaglam, line 109: This is probably an accidental change, try to revert it.

tsaglam (changes requested):
This is only an initial review. There are probably more issues to discuss. Try to address the open comments by all reviewers as soon as possible, and then we can conduct further reviews on the changed code.

tsaglam:
@heylissi, please try to address the feedback as soon as possible and mark anything as resolved that has been fixed.

core/src/main/java/de/jplag/JPlagComparison.java
```diff
@@ -10,8 +10,12 @@
  * @param matches is the unmodifiable list of all subsequence matches between the two submissions.
  * @param ignoredMatches is the unmodifiable list of ignored matches whose length is below the minimum token match
  * threshold.
+ * @param frequencyWeightedSimilarity the similarity influenced by the isFrequencyAnalysisEnabled of a match
+ * @param isFrequencyAnalysisEnabled if the frequencyWeightedSimilarity shall be used or the similarity value without
+ * considering the match isFrequencyAnalysisEnabled
  */
-public record JPlagComparison(Submission firstSubmission, Submission secondSubmission, List<Match> matches, List<Match> ignoredMatches) {
+public record JPlagComparison(Submission firstSubmission, Submission secondSubmission, List<Match> matches, List<Match> ignoredMatches,
+        double frequencyWeightedSimilarity, boolean isFrequencyAnalysisEnabled) {
```
tsaglam, line 18: Isn't `isFrequencyAnalysisEnabled` and `frequencyWeightedSimilarity` a bit redundant? Wouldn't it be possible to remove `isFrequencyAnalysisEnabled` and add a method with that name that just checks if `frequencyWeightedSimilarity` is -1?

core/src/main/java/de/jplag/JPlagComparison.java
```diff
@@ -21,10 +25,18 @@ public record JPlagComparison(Submission firstSubmission, Submission secondSubmi
      * @param matches is the list of all matches between the two submissions.
      */
     public JPlagComparison(Submission firstSubmission, Submission secondSubmission, List<Match> matches, List<Match> ignoredMatches) {
-        this.firstSubmission = firstSubmission;
-        this.secondSubmission = secondSubmission;
-        this.matches = Collections.unmodifiableList(matches);
-        this.ignoredMatches = Collections.unmodifiableList(ignoredMatches);
+        this(firstSubmission, secondSubmission, Collections.unmodifiableList(matches), Collections.unmodifiableList(ignoredMatches), -1, false);
```
tsaglam, line 28: Note how this constructor wraps the lists as unmodifiable ones? The new implicit constructor with more parameters no longer does this (even though it might not be used).

core/src/main/java/de/jplag/Submission.java
```diff
@@ -136,7 +136,7 @@ public File getRoot() {
     /**
      * @return Similarity divisor for the submission.
      */
-    int getSimilarityDivisor() {
+    public int getSimilarityDivisor() {
```
tsaglam, line 141: Does this have to be public? Where is it used?

core/src/main/java/de/jplag/highlightextraction/ContainedMatchesStrategy.java
```diff
@@ -30,4 +41,21 @@ public void processMatchTokenTypes(List<TokenType> matchTokenTypes, Consumer<Lis
...
+     * @return a weight for the match
+     */
+    @Override
+    public double calculateMatchFrequency(Match match, Map<List<TokenType>, Integer> frequencyMap, List<TokenType> matchToken) {
+        List<List<TokenType>> keys = getSubSequences(matchToken, minSubSequenceLength);
+        List<Integer> frequencies = new ArrayList<>();
+        for (List<TokenType> key : keys) {
+            frequencies.add(frequencyMap.getOrDefault(key, 0));
+        }
+        return frequencies.stream().filter(freq -> freq > 0).mapToInt(Integer::intValue).average().orElse(0.0);
```
tsaglam: Avoid abbreviations. Use `it` or `frequency`.

core/src/main/java/de/jplag/highlightextraction/FrequencyAnalysisStrategy.java
```diff
@@ -1,22 +1,25 @@
 package de.jplag.highlightextraction;
 
 /**
- * Enum representing the different strategies for frequency calculation of token subsequences in comparisons.
+ * Enum representing the different strategies for isFrequencyAnalysisEnabled similarity calculation.
  */
-public enum FrequencyStrategies {
+public enum FrequencyAnalysisStrategy {
```
tsaglam, line 6: I still think we can get rid of this enum here (to avoid duplication with `FrequencyStrategy`), but we can also do this after merging. So just ignore this comment and leave it unresolved.

core/src/main/java/de/jplag/highlightextraction/FrequencyDetermination.java
```diff
@@ -20,7 +20,7 @@ public class FrequencyDetermination {
 
     /**
      * Constructor.
-     * @param frequencyStrategy The chosen strategy for frequency calculation.
+     * @param frequencyStrategy The chosen strategy for isFrequencyAnalysisEnabled calculation.
```
tsaglam: This seems like an odd renaming bug, right?

core/src/main/java/de/jplag/highlightextraction/FrequencyDetermination.java
```diff
@@ -41,35 +42,30 @@ public void buildFrequencyMap(List<JPlagComparison> comparisons) {
...
+                List<TokenType> matchTokenTypes = FrequencyUtil.matchesToMatchTokenTypes(match, submissionTokenTypes);
                 frequencyStrategy.processMatchTokenTypes(matchTokenTypes, this::addSequenceKey, this::addSequence, strategyNumber);
             }
         }
     }
 
     /**
      * Adds the Sequence to the Frequency map.
-     * @param sequence The token sequence whose frequency will be updated.
+     * @param sequence The token sequence whose isFrequencyAnalysisEnabled will be updated.
```
tsaglam: same here

core/src/main/java/de/jplag/highlightextraction/FrequencyDetermination.java
```diff
@@ -41,35 +42,30 @@ public void buildFrequencyMap(List<JPlagComparison> comparisons) {
...
      */
     private void addSequenceKey(List<TokenType> sequence) {
         matchFrequencyMap.putIfAbsent(sequence, 0);
     }
 
     /**
-     * @return Map containing (sub-)matches and their frequency according to the strategy.
+     * @return Map containing (sub-)matches and their isFrequencyAnalysisEnabled according to the strategy.
```
tsaglam: same here

core/src/main/java/de/jplag/highlightextraction/FrequencyDetermination.java
```diff
@@ -41,35 +42,30 @@ public void buildFrequencyMap(List<JPlagComparison> comparisons) {
...
      */
     public Map<List<TokenType>, Integer> getMatchFrequencyMap() {
         return matchFrequencyMap;
     }
 
     /**
-     * Updates the frequency of the given sequence in the frequency map.
-     * @param sequence The token sequence whose frequency will be updated.
+     * Updates the isFrequencyAnalysisEnabled of the given sequence in the isFrequencyAnalysisEnabled map.
+     * @param sequence The token sequence whose isFrequencyAnalysisEnabled will be updated.
```
tsaglam: same here

core/src/main/java/de/jplag/highlightextraction/FrequencyStrategy.java
```diff
@@ -1,22 +1,35 @@
...
 import java.util.function.Consumer;
 
+import de.jplag.Match;
 import de.jplag.TokenType;
 
 /**
- * Interface for different frequency calculation strategies. Implementations define how submatches are considered in the
- * frequency calculation of matches.
+ * Interface for different isFrequencyAnalysisEnabled calculation strategies. Implementations define how submatches are
+ * considered in the isFrequencyAnalysisEnabled calculation of matches.
```
tsaglam: Also here...

core/src/main/java/de/jplag/highlightextraction/FrequencyStrategy.java
```diff
@@ -1,22 +1,35 @@
...
 public interface FrequencyStrategy {
     /**
-     * Updates the frequency map with token subsequences and their frequencies according to the implemented strategy.
+     * Updates the isFrequencyAnalysisEnabled map with token subsequences and their frequencies according to the implemented
+     * strategy.
      * @param matchTokenTypes List of match token types representing the match.
-     * @param addSequenceKey Consumer that adds the sequence to the list, without counting the frequency.
-     * @param addSequence Consumer that adds the sequence to the list, and updates the frequency.
+     * @param addSequenceKey Consumer that adds the sequence to the list, without counting the isFrequencyAnalysisEnabled.
+     * @param addSequence Consumer that adds the sequence to the list, and updates the isFrequencyAnalysisEnabled.
```
tsaglam: Please check for more instances of this issue; I will not mark all of them.

core/src/main/java/de/jplag/highlightextraction/FrequencyUtil.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.util.List;
+
+import de.jplag.Match;
+import de.jplag.TokenType;
+
+/**
+ * Contains methods that are used in multiple highlight extraction classes.
+ */
+public class FrequencyUtil {
```
tsaglam: should be final

core/src/test/java/de/jplag/TestBase.java
```diff
@@ -37,7 +37,7 @@ public abstract class TestBase {
      * @param subdirectories list of directories that form a path relative to the base path.
      * @return the base path where the test samples reside concatenated with any number of subdirectories.
      */
-    protected String getBasePath(String... subdirectories) {
+    protected static String getBasePath(String... subdirectories) {
```
tsaglam: Why is this static now?

core/src/test/java/de/jplag/highlightextraction/frequencydetermination/MatchFrequencyWeightingTest.java
```diff
@@ -0,0 +1,244 @@
...
+     */
+    Match testMatch;
+    /**
+     * Match that is contained in the test match.
+     */
+    Match matchContained;
+    /**
+     * Submission used for the tests.
+     */
+    Submission testSubmission;
```
tsaglam: Don't make them public, then you do not need the verbose JDoc comments.

core/src/test/java/de/jplag/highlightextraction/frequencydetermination/StrategyIntegrationTest.java
```diff
@@ -46,10 +56,10 @@ void prepareMatchResult() throws ExitException {
     }
 
     /**
-     * Tests frequency determination using the CompleteMatchesStrategy.
+     * Tests isFrequencyAnalysisEnabled determination using the CompleteMatchesStrategy.
      */
     @Test
-    @DisplayName("Test token frequency completeMatches")
+    @DisplayName("Test token isFrequencyAnalysisEnabled completeMatches")
     void testFrequencyAnalysisStrategiesCompleteMatches() {
```
tsaglam: Even here the renaming issue caused problems.

core/src/main/java/de/jplag/Submission.java
heylissi: It is used in MatchFrequencyWeighting to calculate the new Match frequency based Similarity.

cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -170,6 +177,34 @@ public static class ClusteringEnabled {
         }
     }
 
+    /** Highlight extraction options. */
+    public static class FrequencyAnalysis {
+        /** Skip Highlight extraction options. */
+        @Option(names = {"--isFrequencyAnalysisEnabled-analysis"}, description = "If isFrequencyAnalysisEnabled calculation is used")
```
dfuchss: this option name seems odd :) Consider renaming it (see below)

core/src/main/java/de/jplag/highlightextraction/CompleteMatchesStrategy.java
```diff
@@ -1,26 +1,41 @@
...
 import java.util.function.Consumer;
 
+import de.jplag.Match;
 import de.jplag.TokenType;
 
 /**
  * Strategy that calculates the frequencies of matches across all submissions. For each match, the complete token
- * sequence is added to the frequency map without modification. So the Strategy counts all occurrences of complete
- * matches inside all the complete matches of the comparisons.
+ * sequence is added to the isFrequencyAnalysisEnabled map without modification. So the Strategy counts all occurrences
```
dfuchss: please fix javadoc

core/src/main/java/de/jplag/highlightextraction/CompleteMatchesStrategy.java
```diff
@@ -1,26 +1,41 @@
...
+    /**
+     * Calculates the isFrequencyAnalysisEnabled of a match. According to the strategy.
+     * @param match Considered match
+     * @param frequencyMap Frequency map build with processMatchTokenTypes method
+     * @param matchToken tokenType sequence of the match
+     * @return a weight for the match
+     */
+    @Override
+    public double calculateMatchFrequency(Match match, Map<List<TokenType>, Integer> frequencyMap, List<TokenType> matchToken) {
+        Integer values = frequencyMap.get(matchToken);
```
dfuchss: maybe simple use getOrDefault(..)

core/src/main/java/de/jplag/highlightextraction/ContainedMatchesStrategy.java
```diff
@@ -1,27 +1,38 @@
...
-     * @param addSequence Consumer that adds the sequence to the list, and updates the frequency
+     * @param addSequenceKey Consumer that adds the sequence to the list, without counting the isFrequencyAnalysisEnabled
+     * @param addSequence Consumer that adds the sequence to the list, and updates the isFrequencyAnalysisEnabled
      * @param minSubSequenceLength Minimum length of the considered submatches.
      */
     @Override
     public void processMatchTokenTypes(List<TokenType> matchTokenTypes, Consumer<List<TokenType>> addSequenceKey,
             Consumer<List<TokenType>> addSequence, int minSubSequenceLength) {
-        List<List<TokenType>> subSequences = SubSequenceUtil.getSubSequences(matchTokenTypes, minSubSequenceLength);
+        this.minSubSequenceLength = minSubSequenceLength;
```
dfuchss: Why is this attribute (minSubSequenceLength) dynamic? Could that lead to unwanted side-effects.

core/src/main/java/de/jplag/highlightextraction/ContainedMatchesStrategy.java
dfuchss: First you add "0" if not present. Then you filter them out. This can be refactored.

core/src/main/java/de/jplag/highlightextraction/FrequencyAnalysisOptions.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Options for Frequency Analysis.
+ * @param isFrequencyAnalysisEnabled if isFrequencyAnalysisEnabled analysis is used
+ * @param frequencyStrategy the strategy used to determine the isFrequencyAnalysisEnabled of a Match
+ * @param frequencyStrategyMinValue the minimum considered size of Subsequences from matches in the frequencyStrategy
+ * @param weightingStrategy strategy used to influence the similarity based on Match isFrequencyAnalysisEnabled
+ * @param weightingFactor how strong the impact of the weightingStrategy is
+ */
+public record FrequencyAnalysisOptions(boolean isFrequencyAnalysisEnabled, FrequencyAnalysisStrategy frequencyStrategy, int frequencyStrategyMinValue,
```
dfuchss: Not for this PR. But maybe for a follow-up. You can create builders like here: https://github.com/ardoco/lissa-ratlr/blob/f2db92ec2b079a9001782e9cd02cfcbf86bc0eac/src/main/java/edu/kit/kastel/sdq/lissa/ratlr/configuration/Configuration.java#L98

core/src/main/java/de/jplag/highlightextraction/FrequencyUtil.java
```diff
@@ -0,0 +1,28 @@
...
+        // private constructor to prevent instantiation
+    }
+
+    /**
+     * The Token Type representation of a match.
+     * @param match that's representation is created
+     * @param submissionTokenTypes the Token Type list of the complete submission
+     * @return the Token Type list of the match
+     */
+    public static List<TokenType> matchesToMatchTokenTypes(Match match, List<TokenType> submissionTokenTypes) {
```
dfuchss, line 22: How often is this used? One class for a 3-line method could be too much :D .. but if it's used at multiple places, it's fine :)

core/src/main/java/de/jplag/highlightextraction/MatchFrequency.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.Map;
+
+import de.jplag.TokenType;
+
+/**
+ * Contains the map that maps each Match to its isFrequencyAnalysisEnabled Calculated according to the chosen
+ * FrequencyStrategy.
+ * @param matchFrequencyMap maps each Match to its isFrequencyAnalysisEnabled.
+ */
+public record MatchFrequency(Map<List<TokenType>, Double> matchFrequencyMap) {
```
dfuchss, line 13: Why don't you simply use the map directly?

core/src/main/java/de/jplag/highlightextraction/MatchFrequencyWeighting.java
```diff
@@ -0,0 +1,168 @@
...
+ */
+public class MatchFrequencyWeighting {
+    /**
+     * All comparisons to calculate the similarity for.
+     */
+    List<JPlagComparison> comparisons;
+    /**
+     * Chosen weighting function.
+     */
+    MatchWeightingFunction strategy;
```
dfuchss: why not private?

core/src/main/java/de/jplag/highlightextraction/MatchFrequencyWeighting.java
```diff
@@ -0,0 +1,168 @@
...
+        double maximumFoundFrequency = 0.0;
+        maximumFoundFrequency = getMaximumFoundFrequency(maximumFoundFrequency);
+
+        if (weightingFunction == MatchFrequencyWeightingFunction.PROPORTIONAL) {
+            minimumWeight = 0.01;
+        } else {
+            minimumWeight = 1.0;
+        }
+        maximumWeight = 2.0;
+
```
dfuchss: Please move magic numbers to constants.

core/src/main/java/de/jplag/highlightextraction/SubMatchesStrategy.java
```diff
@@ -1,30 +1,58 @@
...
 
 /**
  * Strategy that counts all occurrences of complete matches and contiguous submatches from the comparisons, if the
  * matches and submatches are longer than minSubSequenceLength.
  */
 public class SubMatchesStrategy implements FrequencyStrategy {
+    /**
+     * Minimum considered subsequence length.
+     */
+    int minimumSubSequenceLength;
```
dfuchss: why not private?

core/src/main/java/de/jplag/highlightextraction/SubSequenceUtil.java
```diff
@@ -39,4 +39,5 @@ public static List<List<TokenType>> getSubSequences(List<TokenType> matchTokenTy
 
         return subSequences;
     }
+
```
dfuchss: please revert this change

core/src/main/java/de/jplag/highlightextraction/WindowOfMatchesStrategy.java
```diff
@@ -13,18 +15,23 @@
  * contiguous windows of the matches from the comparisons.
  */
 public class WindowOfMatchesStrategy implements FrequencyStrategy {
+    /**
+     * The window size for the considered window sequences.
+     */
+    int minSubSequenceLength;
```
dfuchss: why not private?

core/src/main/java/de/jplag/options/JPlagOptions.java
```diff
@@ -49,15 +51,22 @@
...
 public record JPlagOptions(@JsonSerialize(using = LanguageSerializer.class) Language language, Integer minimumTokenMatch,
         @JsonSerialize(contentUsing = FileSerializer.class) Set<File> submissionDirectories,
         @JsonSerialize(contentUsing = FileSerializer.class) Set<File> oldSubmissionDirectories,
         @JsonSerialize(using = FileSerializer.class) File baseCodeSubmissionDirectory, String subdirectoryName, List<String> fileSuffixes,
         String exclusionFileName, SimilarityMetric similarityMetric, double similarityThreshold, int maximumNumberOfComparisons,
-        ClusteringOptions clusteringOptions, boolean debugParser, MergingOptions mergingOptions, boolean normalize, boolean analyzeComments)
-        implements JPlagOptionsBuilder.With {
+        ClusteringOptions clusteringOptions, boolean debugParser, MergingOptions mergingOptions, boolean normalize, boolean analyzeComments,
+        boolean isFrequencyAnalysisEnabled, FrequencyAnalysisStrategy frequencyAnalysisStrategy, int frequencyStrategyMinValue,
+        MatchFrequencyWeightingFunction weightingStrategy, double weightingStrategyWeightingFactor) implements JPlagOptionsBuilder.With {
```
dfuchss: Correction for comment above (use .With also for your withX implementation in the other record :)

core/src/main/java/de/jplag/JPlagComparison.java
```diff
@@ -63,7 +75,10 @@ public final double minimalSimilarity() {
      * @return Average similarity in interval [0, 1]. 0 means zero percent structural similarity, 1 means 100 percent
      * structural similarity.
      */
-    public final double similarity() {
+    public double similarity() {
+        if (isFrequencyAnalysisEnabled && frequencyWeightedSimilarity >= 0) {
```
dfuchss, line 80: Maybe reflect this change also in the JavaDoc

core/src/main/java/de/jplag/Submission.java
dfuchss, line 141: why now public? Shall that be part of the API of JPlag?

core/src/test/java/de/jplag/highlightextraction/frequencydetermination/MatchFrequencyWeightingTest.java
```diff
@@ -0,0 +1,220 @@
...
+import de.jplag.options.JPlagOptions;
+
+/**
+ * Checks if the isFrequencyAnalysisEnabled value is calculated and written into the matches correctly.
+ */
+public class MatchFrequencyWeightingTest extends TestBase {
+    private static final FrequencyStrategy completeMatchesStrategy = new CompleteMatchesStrategy();
+    private static final FrequencyStrategy subMatchStrategy = new SubMatchesStrategy();
+    private static final FrequencyStrategy containedMatchesStrategy = new ContainedMatchesStrategy();
+    private static final FrequencyStrategy windowOfMatchesStrategy = new WindowOfMatchesStrategy();
```
dfuchss: Constants should be CAPS_CASED :)

docs/1.-How-to-Use-JPlag.md
```diff
@@ -106,6 +106,19 @@ Subsequence Match Merging
...
+Frequency Analysis
+      --isFrequencyAnalysisEnabled-analysis
```
dfuchss: see coment above. Name should probably be more like --is-frequency-analysis-enabled

dfuchss (commented):
Just some comments from my side :)

core/src/main/java/de/jplag/highlightextraction/FrequencyUtil.java
heylissi: Hi, it has multiple usages in different classes, so I decided to make it a reusable method.

core/src/main/java/de/jplag/highlightextraction/MatchFrequency.java
heylissi: Hi, the Map contains the results from the first - frequency determination - step of the frequency analysis, which should not be modified afterwards.

core/src/main/java/de/jplag/Submission.java
heylissi: I use this method when calculating a frequency-based similarity score in the MatchFrequencyWeighting class. To avoid duplicating this code, I made the method public.

dfuchss: All right, but then the JavaDoc might be a little bit over simplistic (for a public method) :) What is a similarity divisor?

core/src/main/java/de/jplag/highlightextraction/MatchFrequency.java
dfuchss: But the map can be modified even if it's in the record :)

core/src/main/java/de/jplag/highlightextraction/FrequencyAnalysisOptions.java
tsaglam: Already noted in #2595.

core/src/main/java/de/jplag/Submission.java
heylissi: The similarity divides a value for the Token found in matches of both submissions, with the divisor, which is a value representing the Token length of the submissions . I will add a short description :)

core/src/main/java/de/jplag/highlightextraction/MatchFrequency.java
heylissi: I mainly wanted to make it clear that this is the output of the first of the two parts of the frequency analysis. Using a record should also distinguish it more from the Map that is used to build it. But it probabely depends on preference, I can change that if you think just using a Map is better? :D

PR #2598, opened 2025-09-14
Allow loading files per query on all views

Kr0nox:
Allows loading reports on all views, with the known syntax: `report_viewer_url?file=<REPORT_URL>`


tsaglam (commented):
Besides the Copilot comments, nothing from my side. Could you check them, then we can merge?

PR #2600, opened 2025-09-16
Support for Java 25 (LTS)

tsaglam:
This PR adds support for Java 25. This also means the whole project is built with Java 25, as we use JavaC as part of our Java language module.

- [x] Update pom.xml
- [x] Update CI workflows
- [x] Update documentation
- [ ] Update Java language module
- [ ] Update Java language tests

### DIFF
--- a/.github/workflows/complete-e2e.yml
+++ b/.github/workflows/complete-e2e.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v5
         with:
-          java-version: 21
+          java-version: 25
           distribution: 'temurin'
 
       - uses: actions/setup-node@v5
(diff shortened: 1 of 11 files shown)


tsaglam:
The Temurin release is expected to be in September 2025.

core/pom.xml
```diff
@@ -62,6 +62,20 @@
                     <sourcepath>src/main/java;target/generated-sources/annotations</sourcepath>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.14.0</version>
```
dfuchss: Versions should be managed in the parent pom :)

tsaglam: Yes, this seems to be all over the place across the modules. I decided to fix that in a separate PR for change readability, as the Java version PRs are likely to be referenced in future version updates.

tsaglam: But for the record, builder, I also fixed it in here, I just realized. I can probably fix `maven-compiler-plugin` in here as well.

PR #2694, opened 2025-11-11
mvn dependency management

### LINKED ISSUE #2606: Version numbers of Maven dependencies should be managed in parent pom
As discussed in #2600, currently, many modules have messy dependencies.

- [ ] Version numbers of Maven dependencies should be managed in the parent pom.xml file
- [ ] Dependencies that are inherited from the pom parent should not be redeclared
- [ ] The order of version properties in the pom parent should not be random,

TwoOfTwelve:
This PR moves all dependency and plugin versions in maven into the management sections in the root pom.xml. Redundant and obsolete dependencies have been removed and the plugin configurations have mostly been moved to the pluginManagement section and simplified.

All versions are now defined using 'version.\*' variables in maven
All plugin versions are now defined using 'version.plugin.\*' in maven

Fixes #2606

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -15,131 +15,108 @@
         <dependency>
             <groupId>de.jplag</groupId>
             <artifactId>jplag</artifactId>
-            <version>${revision}</version>
         </dependency>
 
         <!-- Languages -->
         <dependency>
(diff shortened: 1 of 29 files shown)


robinmaisch (approved):
Looking good!

Issue #2683, opened 2025-11-03
Rename core e2e tests


Kr0nox:
The ["e2e tests in the core"](https://github.com/jplag/JPlag/tree/develop/endtoend-testing) should be renamed to avoid confusion and name collisions with test e2e test, that run through the entire jplag process (inclusing jar and report viewer)

This is not a breaking change, as we do not publish that package.

This was first discussed in  #2620

PR #2713, opened 2025-11-19
Rename core end-to-end tests to regression tests

TwoOfTwelve:
Renamed the end to end tests as discussed in: #2683

We still have to discuss how to adapt the documentation


regression-tests/src/main/java/de/jplag/regression_test/constants/TestDirectoryConstants.java
```diff
@@ -1,4 +1,4 @@
-package de.jplag.endtoend.constants;
+package de.jplag.regression_test.constants;
```
tsaglam: Underscores are typically not used in package names according to convention, with the exception of [escaping domains that produce invalid package names](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html). [See also SO](https://stackoverflow.com/questions/3179216/what-is-the-convention-for-word-separator-in-java-package-names). As far as I know, we also do not use underscores in JPlag yet (correct me if I am wrong).

Kr0nox:
@TwoOfTwelve the sonar issues are old ones that just reapeared due to the rename correct?

endtoend-testing/src/test/java/de/jplag/endtoend/architecture/PomVersionTest.java
Kr0nox, line 1: is there a reason this test was removed?

PR #2766, opened 2025-12-24
Add a graph-based Java module with syntax normalization and dead code detection.

LundiNord:
This PR builds upon #2767.

This PR adds abstract interpretation on Java source code to the java-cpg language module.
This is used to detect dead code that is hard/impossible to detect using static analysis. The dead code is removed from the tokens to prevent attacks where dead code is inserted to mask plagiarism.


tsaglam (changes requested):
Both the title and the description should be written in a way that anyone can understand this PR! ;)
See also: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/helping-others-review-your-changes

robinmaisch:
We aim to merge this into the feature/java-cpg branch. @LundiNord, please merge the current branch state into your branch, and also, please remove the "leon" directory.

PR #2779, opened 2026-01-08
Fix vulnerabilites in Sonar CI workflows

tsaglam:
This PR addresses some vulnerabilities in the GitHub Actions workflows related to Sonar scans.
Note that these issues are also raised by Sonar and thus cause the Sonar check to fail for the `develop` branch.

From sonar:

_Using `github.actor` or equivalent properties to check if the actor is a trusted principal on events like `pull_request_target` could be a security issue, because they do not always refer to the actual creator of the commit or the pull request._

_The value represents the entity who triggered the workflow event, which may differ from the actual author of the commit or pull request. If a threat actor could force a trusted actor (such as a bot) into making a change that triggers the workflow, they can bypass the check._

See also: https://rules.sonarsource.com/githubactions/RSPEC-8232/

### DIFF
--- a/.github/workflows/check-sonarcloud-pr.yml
+++ b/.github/workflows/check-sonarcloud-pr.yml
@@ -16,7 +16,7 @@ jobs:
   get-info:
     name: "Get information about the source run"
     runs-on: ubuntu-latest
-    if: ${{ github.actor != 'dependabot[bot]' && github.event.workflow_run.conclusion == 'success' }}
+    if: ${{ github.event.pull_request.user.login != 'dependabot[bot]' && github.event.workflow_run.conclusion == 'success' }}
     outputs:
       sourceHeadRepo: ${{ steps.source-run-info.outputs.sourceHeadRepo }}
       sourceHeadBranch: ${{ steps.source-run-info.outputs.sourceHeadBranch }}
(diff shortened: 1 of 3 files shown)


.github/workflows/check-sonarcloud-pr.yml
```diff
@@ -38,14 +38,16 @@ jobs:
...
     needs: get-info
 
     steps:
       - uses: actions/checkout@v6
+        env:
+          HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
         with:
           repository: ${{ github.event.workflow_run.head_repository.full_name }}
-          ref: ${{ github.event.workflow_run.head_branch }}
+          ref: ${{ env.HEAD_BRANCH }}
```
uuqjz: This change is not related to the linked vulnerability. Why the change here?

tsaglam: See https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&branch=develop&id=jplag_JPlag&open=AZt5mwprAhySEHxlvzKQ.

uuqjz:
```
The recommended approach is to assign untrusted data to environment variables using the env key, then reference these variables using standard shell variable syntax ($VARIABLE_NAME or ${VARIABLE_NAME}) rather than GitHub’s expression syntax (${{ …​ }}).
```

```
While storing untrusted data in environment variables is the correct approach, it is important to reference them properly. Simply putting the data in an environment variable but then accessing it through GitHub’s expression syntax (${{ env.VARIABLE_NAME }}) will still result in command injection vulnerabilities.
```
If i understand the article correctly it should be `$env.HEAD_BRANCH` rather than the Github expression syntax

tsaglam: Good point, I am not sure. I based my change on the listed compliant solution, but it does not match with the text you referenced, so it is quite confusing:

_[Compliant solution:](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&branch=develop&id=jplag_JPlag&open=AZt5mwprAhySEHxlvzKQ&tab=how_to_fix)_
```yaml
name: Example

on:
  pull_request:
    branches: [ main ]

jobs:
  main:
    runs-on: ubuntu-latest

    steps:
      - name: Example action call
        uses: ./helloaction
        with:
          message: "PR title: $PR_TITLE"
        env:
          PR_TITLE: ${{ github.event.pull_request.title }}
```
I have to check in detail what would be the right one.

tsaglam: I avoided the problem entirely by using the commit hash instead of the PR title. Thus injection is less of a problem.

uuqjz (approved):

Kr0nox:
Is it possible that the `event.pull_request.user.login` can cause issues, as the workflows do not get executed on a PR event and thus the workflow can run without having a PR attached. @tsaglam do you know something about that?

tsaglam:
> Is it possible that the `event.pull_request.user.login` can cause issues, as the workflows do not get executed on a PR event and thus the workflow can run without having a PR attached. @tsaglam do you know something about that?

As far as I know, it should not lead to issues, as on non-PR executions the value of `event.pull_request.user.login` should be not set.

See also https://github.com/tsaglam/actions-test/actions/runs/21037970195/job/60491243614 (internal).

PR #2845, opened 2026-03-01
Use language-specific option for metamodel loading when using emf-model language

tsaglam:
This PR replaces the clunky order-based mechanism to ensure metamodel files are loaded before parsing model files. This is done by introducing a language-specific option, which allows specifying one or more paths to `.ecore` files. As a consequence, the `emf-model` language no longer requires a submission order. As this is the only language that uses this feature, we can remove it. The corresponding methods in the `Language` interface are marked as deprecated.

While the prototypical `DynamicEmfLanguage` was never available in the CLI, it was available via the API. With this PR it is removed and only a deprecation placeholder remains. The `EmfModelLanguage` now depends on `EMFLanguage` instead of the `DynamicEmfLanguage`. All functionality that is still required is moved to the `EmfModelLanguage`. 

**Partly solves #576.**

Work in progress, tasks left to do:

- [x] Add more unit tests, especially regarding the metamodel path option.
- [x] Adapt the readme to show how to specify metamodel files.
- [x] Test if the language-specific option works with the jar.
- [ ] Decide if the EMF dynamic maven module and `Language` interface should be removed now or in a later release.

For release notes:
- EMF model language: Uses now the `--metamodel` CLI option to specify metamodel path(s) when loading the metamodel for EMF-based models.
- Breaking change: Deprecate the submission order functionality of JPlag languages.
- Deprecate the experimental dynamic EMF language, which was never available in the CLI.


tsaglam:
@robinmaisch, this PR deprecates a few methods of the public API. Is the next release set to be a major one? I could either remove these methods directly or just deprecate them for now. Any preferences?

robinmaisch (approved):
Nice! Feels good to get rid of very edgy edge cases.

PR #2846, opened 2026-03-04
Add coverage data for report viewer

Kr0nox:
This PR adds a test coverage report for the report-viewer.

The tests get executed for each package and then a script bundles them into a single report.

The action then posts a comment containing:
- The overview for the entire report viewer with the 4 available coverage metrics
- A collapsable section with the metrics for each package
- A collapse section with the metrics for each testable file

An HTML version of the report where a per line analysis can be seen is attached to the action run

If the line coverage falls below 20% the action fails

We should discuss whether the line coverage is what we want for this or whether we should use a different metric or even a composed one like sonarcloud does.
We should discuss if we want a simple message like the one below or put in the work to create a more styled one with colors for different values

### DIFF
--- a/.github/workflows/report-viewer-test-unit.yml
+++ b/.github/workflows/report-viewer-test-unit.yml
@@ -8,6 +8,10 @@ on:
       - ".github/workflows/report-viewer-test-unit.yml"
       - "report-viewer/**"
 
+permissions:
+  issues: write
+  pull-requests: write 
+
 jobs:
   pre_job:
     runs-on: ubuntu-latest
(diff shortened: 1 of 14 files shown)


Kr0nox:
Goal: Coverage on new Code

tsaglam:
Is this not something that could be done via Sonar? Or did you decide specifically against sonar-based coverage?

Kr0nox:
I think it is easily possible to do this with sonar. The only disadvantage would be that sonarcloud and the unit tests would either need to be in the same action or sonarcloud needs to depend on the unit tests and thus only get executed when tests are successful

Kr0nox:
replaced by https://github.com/jplag/JPlag/pull/2988

PR #2847, opened 2026-03-04
Added checkstyle to pre-commit hooks

TwoOfTwelve:
This PR add checkstyle to the pre-commit hook

### DIFF
--- a/gitHooks/scripts/javaPreCommit
+++ b/gitHooks/scripts/javaPreCommit
@@ -8,6 +8,7 @@ fi
 
 #prevents the shell from aborting if maven returns a non zero exit code
 set +e
+
 echo Checking spotless
 mvn spotless:check &> /dev/null
 exitCode=$?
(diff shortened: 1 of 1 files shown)


TwoOfTwelve:
We could also consider running all unit tests in the pre-commit hook, though I'm not sure if that would be worth the additional waiting time

Kr0nox:
I think it would be better if it still ran both and not fail directly after the spottless fail

> We could also consider running all unit tests in the pre-commit hook, though I'm not sure if that would be worth the additional waiting time

I think the unit tests are too long to be a reasonable pre commit hook

TwoOfTwelve:
I ran a quick benchmark on the unit tests:

```
Benchmark 1: mvn verify -B -U
  Time (mean ± σ):     59.783 s ± 14.999 s    [User: 250.361 s, System: 9.935 s]
  Range (min … max):   53.668 s … 102.423 s    10 runs
```

That is probably too long to run it on every commit

Kr0nox (approved):

PR #2900, opened 2026-04-09
Fix Tokens ending before they start

robinmaisch:
* parameters are parsed as length 0 by JavaC, then FixedSourcePositions previously turned it into length -1 
 * enum constants are parsed as the correct length, but the "initializer length" was subtracted from it (which is not present in the code). We use the (also artificially inserted) comment `/*enum*/` to detect this case. 
 - Similarly, for the implicit or explicit constructor calls for enum constants, no identifier is present in the code, so its length should not be subtracted.
 * For Java 25 compact classes, there was an off-by-one error in checking the start position.

### DIFF
--- a/languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java
+++ b/languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java
@@ -492,11 +492,13 @@ final class TokenGeneratingTreeScanner extends TreeScanner<Void, Void> {
             long start = positions.getStartPosition(ast, node);
             long end = positions.getEndPosition(ast, node);
             if (Objects.isNull(node.getInitializer())) {
-                // VARDEF token should end before the start of the assigned value.
+                // VARDEF token should end before semicolon
                 end -= 1;
-            } else if (!node.toString().contains(ENUM_MARKER)) { // enum initializers are implicit - no need to subtract.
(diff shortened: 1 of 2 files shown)


languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java
```diff
@@ -483,8 +490,13 @@ public Void visitAssert(AssertTree node, Void unused) {
     public Void visitVariable(VariableTree node, Void unused) {
         if (!node.getName().contentEquals(ANONYMOUS_VARIABLE_NAME)) {
             long start = positions.getStartPosition(ast, node);
-            long end = positions.getEndPosition(ast, node) - 1;
-            end -= node.getInitializer() == null ? 0 : node.getInitializer().toString().length();
+            long end = positions.getEndPosition(ast, node);
+            if (Objects.isNull(node.getInitializer())) {
+                // VARDEF token should end before the start of the assigned value.
+                end -= 1;
+            } else if (!node.toString().contains(ENUM_MARKER)) { // enum initializers are implicit - no need to subtract.
```
Kr0nox: Maybe rephrase this comment to say `-1` does not need to be subtracted, as you do subtract something and this caused confusion for me at first

robinmaisch: Agreed.

Kr0nox (approved):

PR #2901, opened 2026-04-14
feat(language): add bash

p-fruck:
This PR aims to add the bash language to jplag. I hoped to find a public antlr grammer for bash, but I didn't. Hence, Claude Opus has been used to generate the Grammer (AI assisted).

I reviewed and tested the grammer locally and it seems to work. Kindly let me know if you are interested in the upstream contribution of this language.

Personally, I require the feature of checking for plagiarism in shell scripts. However, I find it difficult to manage because of the differences between bash/ash/sh and the POSIX standard - This implementation cannot handle all of them, but I satisfied with the results for now.


tsaglam:
@jplag/maintainer @jplag/devteam, I fear we need to have a discussion regarding large blobs of AI-generated code. The effort for reviewing this and ensuring it works somewhat correctly is _so_ much higher than generating this code, that I don't think that it makes sense to accept it. Especially for grammars, which should be carefully tried and tested. This seems fully AI-generated, even the readme reads like it. I think JPlag may need a policy for this. This is just a short comment; I may expand on this when I find more time for it.

p-fruck:
No worries, this is completely fine for me. Feel free to close the PR if you want, I don't want to put additional load on the team, just thought of an easy way to add/prototype the new language.
Anyways, thanks for designing this cool tool!

TwoOfTwelve:
I took a look at your grammar and while it looks good in general I managed to break it with this script:
```bash
test=(1 2 3 4 5)
count=${#test}
```
The grammar does not support cases like the length of arrays. Grammars breaking in specific cases is not entirely unusual in JPlag, but we should still be aware of it.

Additionally, this language module is written in the same style the Java module is, however, there are specific APIs in JPlag for antlr language modules. It would be preferable to use that. You can take a look at modules like kotlin or typescript for examples.

tsaglam:
Here are some more thoughts in no particular order:

- The README files are inconsistent. The grammar readme states: "[...] covers the most commonly used Bash constructs." while the main readme states "[...] covers commonly used Bash constructs [...] It is not a complete implementation". Which one is it? This is an important distinction, especially for code that was AI-generated and has seen little to no verification.
- This PR includes next to no tests at all. Especially, no proper tests for the grammar itself. This means there was no effort put into verifying that the generated code and grammar are not just a blob of AI slop. Testing important in normal SWE, and even more important for newly created grammars. But when the entire thing is AI-generated, this becomes even more important.
- For JPlag, we try to stick to tried and tested grammar. Even a grammar manually created from scratch would require careful verification regarding correctness and usefulness. This is also true for existing grammars from existing repos. However, an AI-generated grammar is truly just a red flag.
- JPlag already suffers from the fact that we cannot include the official ANTLR grammars via dependencies, thus increasing the maintenance effort by requiring us to regularly update the grammars and the language module code. By directly submitting the fully generated grammar to us, you are shifting even more maintenance effort onto us. The _correct_ way would have been to submit the grammar to the ANTLR grammar repo first, and then, after it has been accepted, submit your PR here (however, I am not sure if they are interested in AI-generated grammars as PR submissions). Obviously, this point is something you could **not** have been aware of.
- The reasoning for the tokenization decisions is somewhat weak at best. Yes, legacy language modules are not much better in that regard for historical reasons (due to the age of JPlag, I think it is fair to use that word here), but for new languages, we should try to have a higher standard. Here, stating that the tokenization seems to work for you without providing any in-depth tests or data is not the way to go, and very telling of the way many people seem to use generative AI today.
- From my perspective, this PR is missing a clearer baseline for what would make it acceptable. To move forward with this PR, I would expect a clearly defined scope of the grammar, a meaningful set of tests (especially for the grammar), and some form of validation of correctness. Additionally, the tokenization decisions should be documented so that their rationale is understandable.
- I want to acknowledge that I have some bias against the _wrong kind of_ use of generative AI. It can be a useful tool when applied carefully and appropriately. However, software engineering does not consist of generating code and assuming it behaves correctly. In particular, AI does not address core concerns such as defining precise requirements, ensuring correctness, and designing for maintainability.
- At a minimum, I think AI-generated output should be treated as a black box that requires thorough verification. This includes carefully checking the code for correctness, usefulness, and long-term maintainability before submitting it for human review.
- This is not meant to be personal to the PR author; this is just the first time we at JPlag have to deal with a fully AI-based submission, so I hope my answer is taken in the spirit of constructiveness and not misunderstood. I DO NOT want to discourage first-time contributors. Also, this is my personal opinion, no concluding consensus of the @jplag/maintainer team.

p-fruck:
Thanks for all the feedback! This first version was inspired by the rust language implementation, I moved over the same API that Typescript uses - looks much cleaner!
I am currently working on integrating against the GNU bash tests to see if I can parse them without any errors - not sure if I'll finish adding all the bashisms this week...

Is there any way to implement language supersets? E.g. adding POSIX and then bash on top of it? Might have to take another look at your ts/js implementation, but I'd guess I need a separate lexer and parser?

tsaglam:
> Is there any way to implement language supersets? E.g. adding POSIX and then bash on top of it? Might have to take another look at your ts/js implementation, but I'd guess I need a separate lexer and parser?

Currently, we do not have such a mechanism. The JavaScript module is just deferring to the TypeScript module, thus both modules share the same parser, tokens, and grammar. We have some language modules that share functionalities, namely the EMF modules, but they are highly specialized modules that do not really fit your use case. I would suggest focusing on your bash module for now, and adding a POSIX variant later on when this PR is merged.

p-fruck:
@tsaglam thanks for the response! Honestly I am not sure if it makes sense to add an additional POSIX parser if no grammer can be shared since the bash module should be able to parse pure POSIX scripts as well? I migrated to the new Java API and added support for some more common bash specific features like command substition, heredoc etc. I think the full bash specification is out of scope, but this implementation should be capable of handling the common syntax features used in most scripts

robinmaisch:
Dear @p-fruck, thank you very much for your contribution! I agree with @tsaglam: a PR with predominantly AI-generated code puts us in a new situation here. A few points on behalf of the maintainers, even if I repeat part of what has already been said.

- As mentioned above, one core issue is that we usually rely on community-trusted grammar specifications from antlr/grammars-v4 for our ANTLR-based language modules. As we are not exactly experienced in grammar engineering, we do not want to accept the responsibility for verifying – and, in the long run, maintaining – the grammar that you supply. We therefore suggest that you put up the grammar to antlr/grammar-v4, get it approved over there, and only then will we consider including it.
- Also, the set of _common syntax features used in most scripts_ is hardly well-defined. Users will come running and yelling at us that their scripts could not be parsed. We need a clear specification of the feature set in order to manage expectations.
- Carefully considered token selection is a core feature that sets JPlag apart from similar tools, and we have regularly spent days trying to get it right for previously new language modules. So, please provide an explanation/rationale of which syntactic elements are featured as tokens, and which are omitted.
- Apart from the coverage and completeness of tokenization, please also include some example pairs of corresponding bash files and include those in the test process. As of now, we do not have a clear sense of what a plagiarized bash script might look like if it is not an exact copy, and we really would like to learn what kinds of plagiarism you encounter in your use case, even if it is just basic examples!
- PRs against the main branch are, of course, a no-no, but that should be easy to fix. Please merge the current _develop_ branch into your forked branch.

TwoOfTwelve:
I took another look at the code. Usually we prefer using our test framework for language tests (used in most language modules, for example: java). That should offer a good variety of tests and a decent coverage.

The token position tests are a lot of work to create, but give fine control over token positions.

That's probably something to look into once you figured out if you can add the grammar to the official antlr repository.

Kr0nox:
Closed as stale, if there is any update, feel free to reopen

PR #2908, opened 2026-04-16
Rewrite of the rust language module using the new language module api

TwoOfTwelve:
This PR contains a rewrite of the rust language module. The token rules are, as far as I can tell, identical to the original ones. Instead of the stack based approach of the old implementation this looks up the context of a node on demand without holding a state.


robinmaisch (commented):
This was a real headache to do back then. Looks much better now, thank you!

PR #2917, opened 2026-04-24
Java token position tests - Continue to Loop end

TwoOfTwelve:
Contains tests for token positions for the java token types continue to loop_end.

This PR also adjusts some token lengths. Since it does not change the token list, I think it's fine to do this without a separate PR


languages/java/src/main/java/de/jplag/java/TokenGeneratingTreeScanner.java
```diff
@@ -424,8 +424,8 @@ public Void visitNewClass(NewClassTree node, Void unused) {
     @Override
     public Void visitTypeParameter(TypeParameterTree node, Void unused) {
         long start = positions.getStartPosition(ast, node);
-        // This is odd, but also done like this in Java 1.7
-        addToken(JavaTokenType.J_GENERIC, start, 1, new CodeSemantics());
+        // This is odd, but also done like this in Javagg 1.7
```
robinmaisch: I'm sorry, what is Javagg?

robinmaisch (approved):

PR #2933, opened 2026-05-04
Fix overflow at top of comparison table

Kr0nox:
Elements at the top of the comparison table could easily overflow, especially when zoomed in.

This fixes that, by defining more concrete wrapping behaviors


tsaglam (approved):
Looks good on my side (macOS, Firefox, 100% Zoom): 

image

robinmaisch:
Perfect, also looks good on my part.

Issue #2934, opened 2026-05-06
CLI subcommand argument -v


TwoOfTwelve:
When using subcommands, the -v argument should provide the supported language version, e.g. Java 21 for the Java language module.

PR #2948, opened 2026-05-07
Add version information to language modules

TwoOfTwelve:
Adds version information to the language modules. It is now contained in the READMEs and the language modules. -V is now supported for JPlag and all language modules. Calling it on jplag directly prints the JPlag version, calling it on a language module prints the supported language version.

Solves #2934


TwoOfTwelve:
The duplications are due to the language classes. I think we should ignore them again.

robinmaisch:
Can JPlag now also display the supported language version of the language modules, or only for JPlag itself?

TwoOfTwelve:
Yes, it does. I expanded the PR description

robinmaisch:
Ah, I see, it works for
`java -V`
but not 
`-l java -V`
`-V -l java`
`-V java`

TwoOfTwelve:
I could support the other cases, but that would be unusual for CLIs.

Kr0nox:
When drafting the release notes we should remember that this is an api change and we thus should highlight it there

TwoOfTwelve:
We could also add a default implementation that returns something like "No version information provided", if we don't want to have an API breaking changes

robinmaisch:
> I could support the other cases, but that would be unusual for CLIs.

The expected behavior for me was "-V + any kind of language option => supported language version", but then again, I am not the right person to ask about conventions of CLI parameter combinations. At any rate, I am missing a proper guide to accessing the language version in the help text.

robinmaisch:
While we are at it, what does "See subcommands below." refer to? I would think there are some language-specific commands listed somewhere, but none appear "below" :)

languages/emf-metamodel/src/main/java/de/jplag/emf/EmfLanguage.java
```diff
@@ -53,6 +53,11 @@ public int minimumTokenMatch() {
         return 6;
     }
 
+    @Override
+    public String getVersionFlagInformation() {
+        return "emv 1.1.0";
```
tsaglam: Small typo:
```suggestion
        return "EMF 1.1.0";
```

tsaglam: Also, probably not 1.1.0? Should be 2.xx. But I think we also have an inconsistent version information in the readme, as the deps currently are:

https://github.com/jplag/JPlag/blob/508e92605827d8ff3319fa9480c964a188c7f5b8/pom.xml#L85-L86

(on main, on dev it is 2.45)

TwoOfTwelve:
> Ah, I see, it works for `java -V` but not `-l java -V` `-V -l java` `-V java`

It now works for all these cases. To make things consistent, `-h` is handled the same way

PR #2950, opened 2026-05-07
Add curve to distribution diagram

### LINKED ISSUE #2657: Add curve to distribution diagram
### Feature Description

There should be an option to overlay a curve (more correctly a curve with values from the binomial distribution) calculated from the mean of the distribution

### Use Case

This would make it clearer where outliers may be

Kr0nox:
Fixes #2657 

Adds an option to overlays a curve onto the distribution diagram.
The curve gets calculated from the binomial distribution using the average bucket value.

To fix the previous log issues, we just shift everything up by 1. That way we avoid 0 values. To the user this shift is never visible.

grafik

grafik


robinmaisch:
The idea was that the curve would highlight outliers, assuming that an "authentic" dataset often follows a binomial distribution, and potential plagiarisms stand out from the curve, right? I'm not convinced this visualization serves this purpose well. Maybe we need something along the lines of:
"If the distribution can be _much_ better described as the sum of _two_ binomial distributions than just one, then we show both curves, one for the bulk and one for the potential outliers."

This would really emphasize the potential plagiarism cases. In the screenshots, it seems the curve ignores the outliers, downplaying their importance. Of course, we need to develop a robust mathematical model for this (the two means should not be too close together, etc.). @jplag/devteam, what do you think?

Kr0nox:
@tsaglam What are your thoughts on this? do you remember a bit more about our original intend with this feature?

tsaglam:
> @tsaglam What are your thoughts on this? do you remember a bit more about our original intend with this feature?

Ah, it seems I overlooked this mention. Sorry. Our original idea was to somehow give another potential indicator that the statistical outliers at the top are conspicuous. However, early on, we had many concerns due to the "one curve to fit them all" nature of the approach. 

I still think it is quite hard to build such a feature, so it is unlikely to be misleading in some cases. Thus, I would definitely disable it by default, and clearly mark it as an experimental feature (e.g., toggle text could have prefix `(experimental)` and mouseover text could be `experimental feature: [...] (do not interpret as authoritative)`.

In its current form, I think it is not ideal, but this is more of an educated guess. I would definitely test it on many historic datasets, especially different types and sizes of assignments. However, I would also be careful with two binomial distributions or any other hybrid approach, as the duality of such a visualization might suggest a binary classification, meaning plagiarism and not plagiarism, which is exactly what jplag tries to avoid.

I know this answer is _not_ really that helpful. I think the best advice I have is to proceed with caution.

PR #2966, opened 2026-05-19
Show failed names in groups in info tab

### LINKED ISSUE #2790: Group failed submissions by reason
### Feature Description

I suggest grouping the list of failed submissions by reason. For each reason string, I would like the number of corresponding failed submissions and a collapsed list of submission names (preferably one submission name per line; maybe in a scrollable container; maybe with a preview text when collaped such as "_Submission753, Submission007, and 85 more_" like on the overview view).

### Use Case

When examining the failed submissions in the Report Viewer, the different reasons for failure may indicate that the user needs to take action to fix the failure, especially for "could not parse", which may be a syntax error (on the author's part) or an encoding issue (for instructors to fix).
Currently, the list of failed submissions makes it hard to get an overview of the reasons for failures. I would argue that this information is even slightly more relevant than the submission names—at least regarding what users would do about it.

Kr0nox:
Closes #2790 

I omitted showing anonymized names in the due to (1) them not being of any use to the user and (2) the user having to explicitly expand the section to see the names, thus hiding them by default

grafik
grafik
grafik
grafik

<details>
<summary>Entire Info Screen</summary>

grafik

</details>

### DIFF
--- a/report-viewer/report-viewer/src/views/InformationView.vue
+++ b/report-viewer/report-viewer/src/views/InformationView.vue
@@ -126,13 +126,13 @@
         <TextInformation label="Missing Comparisons" class="pb-1">{{
           missingComparisons
         }}</TextInformation>
-        <TextInformation label="Failed Submissions" class="pb-1">{{
-          runInformation.failedSubmissions.length > 0
-            ? runInformation.failedSubmissions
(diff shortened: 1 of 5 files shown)


robinmaisch (approved):
Looks good and is even backwards-compatible. Nice!

Issue #2978, opened 2026-05-27
Update Antlr Grammars


TwoOfTwelve:
This is a general issue to collect what ANTLR grammars need an update

The following ANTLR grammars should be updated:

- [x] C#
- [ ] C++ (Might not be possible, because the current grammar does not parse one of our test files)
- [x] Go
- [x] Kotlin
- [ ] Python
- [ ] Rlang
- [x] Rust
- [ ] Swift
- [ ] Typescript

Issue #2562, opened 2025-08-22
v6.2.0 skipping most files for being "too small"


b0ink:
### Bug Description

I've updated from v6.1.0 to v6.2.0 and the same set of files are now showing up as failed submissions for being "too small"

Is there a CLI option that controls this behaviour? It was working well previously and would like those files to be included

### JPlag Version

6.2.0

TwoOfTwelve:
This might be fixed by updating the grammar (#2978). If that does not work, we should create an issue in the antlr-grammars repository

TwoOfTwelve:
Tracked in #2979

Issue #1055, opened 2023-05-02
Crash using event accessors with lambda expressions in C# 6


olivernybroe:
Given the following code
```csharp
public class MyEventProvider
{
    internal event EventHandler MyEvent
    {
        add => test.progressUpdated += value;
        remove => test.progressUpdated -= value;
    }
}
```

it results in a complete crash
<details><summary>Stacktrace</summary>
<p>

```
2023-05-02-03:45:16_789 [main] [INFO] LanguageLoader - Available languages: '[C/C++ Scanner [basic markup], C# 6 Parser, EMF metamodel, Go Parser, Javac based AST plugin, Kotlin Parser, Python3 Parser, R Parser, Rust Language Module, Scala parser, SchemeR4RS Parser [basic markup], Swift Parser, Text Parser (naive)]'
line 5:12 mismatched input '=' expecting '{'
Exception in thread "main" java.lang.ClassCastException: class de.jplag.csharp.grammar.CSharpParser$Namespace_or_type_nameContext cannot be cast to class de.jplag.csharp.grammar.CSharpParser$Local_variable_declarationContext (de.jplag.csharp.grammar.CSharpParser$Namespace_or_type_nameContext and de.jplag.csharp.grammar.CSharpParser$Local_variable_declarationContext are in unnamed module of loader 'app')
	at de.jplag.csharp.grammar.CSharpParserBase.isLocalVariableDeclaration(CSharpParserBase.java:19)
	at de.jplag.csharp.grammar.CSharpParser.local_variable_declaration_sempred(CSharpParser.java:19486)
	at de.jplag.csharp.grammar.CSharpParser.sempred(CSharpParser.java:19473)
	at org.antlr.v4.runtime.atn.SemanticContext$Predicate.eval(SemanticContext.java:101)
	at org.antlr.v4.runtime.atn.ParserATNSimulator.evalSemanticContext(ParserATNSimulator.java:1433)
	at org.antlr.v4.runtime.atn.ParserATNSimulator.splitAccordingToSemanticValidity(ParserATNSimulator.java:1349)
	at org.antlr.v4.runtime.atn.ParserATNSimulator.getSynValidOrSemInvalidAltThatFinishedDecisionEntryRule(ParserATNSimulator.java:1304)
	at org.antlr.v4.runtime.atn.ParserATNSimulator.execATN(ParserATNSimulator.java:472)
[...]

PR #2986, opened 2026-05-28
CSharp grammar update (C# 8)

TwoOfTwelve:
Updates the C# grammar to the current version. This updates the supported C# version from 6 to 8.

Some features of version 6 do not work anymore in version 8, therefore the test file had to be adjusted. The token order with annotations is a little strange now, but this matches the way it is done in Java.

As far as I can tell the tokens are still equivalent to the old language module.

Relates to #2978 

The new grammar can still not parse the cases mentioned in #2562 and #1055


languages/csharp/README.md
```diff
@@ -1,6 +1,6 @@
-# CSharp langauge module
+# C# language module
 
-Extracts token from CSharp source using the ANTRL Grammar from the official [ANTLR repository](https://github.com/antlr/grammars-v4/tree/master/csharp)
+Extracts token from C# source using the ANTRL Grammar from the official [ANTLR repository](https://github.com/antlr/grammars-v4/tree/master/csharp)
```
robinmaisch:
```suggestion
Extracts tokens from C# source using the ANTLR Grammar from the official [ANTLR repository](https://github.com/antlr/grammars-v4/tree/master/csharp).
```

PR #2987, opened 2026-05-29
Update cpp grammar (Still C++ 14)

TwoOfTwelve:
Updates the C++ grammar to the current version from the ANTLR repository. Currently the grammar does not work for one of our test files. A PR for the ANTLR repository that fixes this issue has been created [here](https://github.com/antlr/grammars-v4/pull/4874). Once that is merges, I'll update the grammar here and we can merge it.

This also adds the possibility to disable checkstyle for specific files by adding a comment, so we can prevent it from failing on the ANTLR base code.

Relates to #2978


robinmaisch (approved):
There seem to be minimal adaptations from our side for this update, so nothing much to worry about.

PR #3007, opened 2026-06-09
Updated Rust grammar (only reformatting)

TwoOfTwelve:
Updates the Rust grammar

#2978


robinmaisch:
Is this really a grammar update or just reformatting?

TwoOfTwelve:
Looks like all the changes are just formatting and code-style. I still think we should update it to be consistent with the ANTLR repository

robinmaisch:
If there is any indication, please add the current Rust grammar version, also in the PR title.

PR #3008, opened 2026-06-09
Updated swift grammar (Still Swift 5.4)

TwoOfTwelve:
Updated the swift grammar.

#2978


robinmaisch (approved):
Mostly ANTLR code with only one naming adaptation.

Issue #3000, opened 2026-06-04
Add support for configurable bind address for Report Viewer server


Kuznetsov-Artyom:
### Feature Description

The Report Viewer server is hardcoded to bind to `127.0.0.1` in `ReportViewer.java:73`, which prevents Docker port forwarding from working.

I would like to configure the bind address via environment variable or CLI flag:

```bash
export JPLAG_BIND_ADDRESS=0.0.0.0
# Or CLI flag
java -jar jplag.jar --bind-address 0.0.0.0 ...
```
## Suggested implementation:
```java
String bindAddress = System.getenv().getOrDefault("JPLAG_BIND_ADDRESS", "127.0.0.1");
server = HttpServer.create(new InetSocketAddress(InetAddress.getByName(bindAddress), currentPort), 0);
```
Default should remain 127.0.0.1 for security.

## Running JPlag in Docker with port forwarding:

```bash
docker run --rm -it -p 1996:1996 -v $PWD:/home/user/data jplag:6.3.0
```
## Current problem:
Server binds to 127.0.0.1 inside container, making it inaccessible from host browser.

## Current workarounds:
- --network host — Linux only, not available on macOS/Windows
- sed patch — unstable, breaks with JPlag updates
Desired: Cross-platform Docker support with standard port forwarding, useful for CI/CD pipelines and containerized deployments.

TwoOfTwelve:
This behavior is intended to prevent accidentally leaking sensitive information over the network. It might be fine to open it in the network if a cli flag is explicitly specified. However, that decision is up to @jplag/maintainer.

I'd also opt for only using the cli flag and not the environment variable, as people might forget they had set it.

tsaglam:
@TwoOfTwelve I would agree; a CLI flag is the better option. I see no issue with opening it up as long as the user chooses to do so explicitly.

Kuznetsov-Artyom:
@tsaglam @TwoOfTwelve Thank you for your answers! 
Then I'll post a patch.

TwoOfTwelve:
Let me know if you have any trouble. In that case I can make the change

Kuznetsov-Artyom:
@TwoOfTwelve Hi! Okay, I was planning on doing this in the next 2-3 days.

tsaglam:
Closed by #3017.

PR #3017, opened 2026-06-10
Add --host flag for configurable report viewer bind address

Kuznetsov-Artyom:
Adds --host/-H CLI option to configure the Report Viewer server bind address. The server was previously hardcoded to 127.0.0.1, making it inaccessible via Docker port forwarding or remote network access.

### Changes:
- CliOptions.java: Added --host option (default: 127.0.0.1)
- ReportViewer.java: Added InetAddress bindAddress parameter, Objects.requireNonNull check, deprecated old constructor
- JPlagRunner.java: Added InetAddress parameter with deprecated overload, IPv6 URL formatting, 0.0.0.0 warning, localhost display for wildcard addresses
- CLI.java: Added fail-fast validateHostOption() in executeCli()
- HostOptionTest.java: 6 new tests (default, custom, localhost, IPv6, invalid, error message)
- CliArgument.java: Added HOST test constant
- README.md, docs/1.-How-to-Use-JPlag.md: Updated CLI help text

### Closes issue #3000

### DIFF
--- a/cli/src/main/java/de/jplag/cli/CLI.java
+++ b/cli/src/main/java/de/jplag/cli/CLI.java
@@ -3,8 +3,6 @@ package de.jplag.cli;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.net.InetAddress;
-import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -67,8 +65,6 @@ public final class CLI {
             // help text has been printed, do nothing else
             return;
         }
-
-        validateHostOption();
 
         CollectedLogger.setLogLevel(this.inputHandler.getCliOptions().advanced.logLevel);
         ProgressBarLogger.setProgressBarProvider(new CliProgressBarProvider());
@@ -143,17 +139,7 @@ public final class CLI {
      */
     public void runViewer(File resultFile) throws IOException {
         finalizeLogger(); // Prints the errors. The later finalizeLogger will print any errors logged after this point.
-        InetAddress bindAddress = InetAddress.getByName(this.inputHandler.getCliOptions().advanced.host);
-        JPlagRunner.runInternalServer(resultFile, this.inputHandler.getCliOptions().advanced.port, bindAddress);
-    }
-
-    private void validateHostOption() throws CliException {
-        String host = this.inputHandler.getCliOptions().advanced.host;
-        try {
-            InetAddress.getByName(host);
-        } catch (UnknownHostException e) {
-            throw new CliException("Invalid bind address: " + host);
-        }
+        JPlagRunner.runInternalServer(resultFile, this.inputHandler.getCliOptions().advanced.port, this.inputHandler.getCliOptions().advanced.host);
     }
 
     private void selectModeAutomatically() throws IOException, ExitException {
--- a/cli/src/main/java/de/jplag/cli/CLI.java
+++ b/cli/src/main/java/de/jplag/cli/CLI.java
@@ -3,6 +3,8 @@ package de.jplag.cli;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -65,6 +67,8 @@ public final class CLI {
             // help text has been printed, do nothing else
             return;
         }
+
+        validateHostOption();
 
         CollectedLogger.setLogLevel(this.inputHandler.getCliOptions().advanced.logLevel);
         ProgressBarLogger.setProgressBarProvider(new CliProgressBarProvider());
@@ -139,7 +143,17 @@ public final class CLI {
      */
     public void runViewer(File resultFile) throws IOException {
         finalizeLogger(); // Prints the errors. The later finalizeLogger will print any errors logged after this point.
-        JPlagRunner.runInternalServer(resultFile, this.inputHandler.getCliOptions().advanced.port);
+        InetAddress bindAddress = InetAddress.getByName(this.inputHandler.getCliOptions().advanced.host);
+        JPlagRunner.runInternalServer(resultFile, this.inputHandler.getCliOptions().advanced.port, bindAddress);
+    }
+
+    private void validateHostOption() throws CliException {
+        String host = this.inputHandler.getCliOptions().advanced.host;
+        try {
+            InetAddress.getByName(host);
+        } catch (UnknownHostException e) {
+            throw new CliException("Invalid bind address: " + host);
+        }
     }
 
     private void selectModeAutomatically() throws IOException, ExitException {
--- a/cli/src/main/java/de/jplag/cli/JPlagRunner.java
+++ b/cli/src/main/java/de/jplag/cli/JPlagRunner.java
@@ -39,18 +39,6 @@ public final class JPlagRunner {
      * Runs the internal server. Blocks until the server has stopped.
      * @param resultFile is the result file to pass to the server. May be null.
      * @param port is the port to open the server on.
-     * @throws IOException if the internal server throws an exception
-     * @deprecated Use {@link #runInternalServer(File, int, InetAddress)} instead
-     */
-    @Deprecated(since = "7.0.0", forRemoval = true)
-    public static void runInternalServer(File resultFile, int port) throws IOException {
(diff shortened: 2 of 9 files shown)


cli/src/main/java/de/jplag/cli/server/ReportViewer.java
```diff
@@ -40,8 +42,23 @@ public class ReportViewer implements HttpHandler {
      * @param resultFile The result file to use for the report viewer
      * @param port The port to use for the server. You can use 0 to use any free port.
      * @throws IOException If the result file cannot be read
+     * @deprecated Use {@link #ReportViewer(File, int, InetAddress)} instead
      */
+    @Deprecated
```
tsaglam: `@Deprecated` is missing `since` and `forRemoval` arguments, which SonarCloud already flagged. Should be:
```suggestion
    @Deprecated(since = "7.0.0", forRemoval = true)
```

cli/src/main/java/de/jplag/cli/JPlagRunner.java
```diff
@@ -38,23 +40,61 @@ public static JPlagResult runJPlag(JPlagOptions options) throws ExitException {
      * @param resultFile is the result file to pass to the server. May be null.
      * @param port is the port to open the server on.
      * @throws IOException if the internal server throws an exception
+     * @deprecated Use {@link #runInternalServer(File, int, InetAddress)} instead
      */
+    @Deprecated
```
tsaglam: Same here...

```suggestion
    @Deprecated(since = "7.0.0", forRemoval = true)
```

cli/src/main/java/de/jplag/cli/CLI.java
```diff
@@ -139,7 +143,17 @@ public void runAndView() throws IOException, ExitException {
...
+        InetAddress bindAddress = InetAddress.getByName(this.inputHandler.getCliOptions().advanced.host);
+        JPlagRunner.runInternalServer(resultFile, this.inputHandler.getCliOptions().advanced.port, bindAddress);
+    }
+
+    private void validateHostOption() throws CliException {
+        String host = this.inputHandler.getCliOptions().advanced.host;
+        try {
+            InetAddress.getByName(host);
+        } catch (UnknownHostException e) {
+            throw new CliException("Invalid bind address: " + host);
```
tsaglam:
```suggestion
            throw new CliException("Invalid bind address: " + host, e);
```

cli/src/main/java/de/jplag/cli/options/CliOptions.java
```diff
@@ -131,6 +131,10 @@ public static class Advanced {
         @Option(names = {"-P", "--port"}, description = "The port used for the internal report viewer (default: ${DEFAULT-VALUE}).")
         public int port = 1996;
 
+        /** Bind address for internal report viewer. */
+        @Option(names = {"-H", "--host"}, description = "The bind address for the internal report viewer (default: ${DEFAULT-VALUE}).")
+        public String host = "127.0.0.1";
```
tsaglam: `host` is stored as a raw `String` and resolved to `InetAddress` twice (once in `validateHostOption()`, once in `runViewer()`). A cleaner approach is to use a picocli type converter and store it as `InetAddress` directly:

```java
@Option(names = {"-H", "--host"}, description = "...", converter = InetAddressConverter.class)
public InetAddress host = InetAddress.getLoopbackAddress();
```

This eliminates the duplicate resolution, removes the need for `validateHostOption()` entirely, and makes the validated type available everywhere downstream.

cli/src/test/java/de/jplag/cli/HostOptionTest.java
```diff
@@ -0,0 +1,60 @@
...
+import java.io.IOException;
+
+import org.junit.jupiter.api.Test;
+
+import de.jplag.cli.picocli.CliInputHandler;
+import de.jplag.cli.test.CliArgument;
+import de.jplag.cli.test.CliTest;
+import de.jplag.exceptions.ExitException;
+
+class HostOptionTest extends CliTest {
```
tsaglam, line 17: Consider deduplicating repeatedly used values in this class via well-named constants.

For example:
```java
private static final String DEFAULT_HOST = "127.0.0.1";
private static final String ANY_HOST = "0.0.0.0";
private static final String LOCALHOST = "localhost";
private static final String IPV6_LOOPBACK = "::1";
private static final String INVALID_HOST = "invalid.host";
private static final String INVALID_HOST_WITH_DOMAIN = "invalid.host.address.that.does.not.exist";
```

tsaglam (changes requested):
Thanks for the contribution. Overall, the PR looks nice. However, there are some smaller issues we should address.

tsaglam (commented):
Overall, this looks good to me and appears to address #3000 as discussed.

I recommend documenting the security implications more explicitly. The default remains loopback-only, which is great, but `--host 0.0.0.0` intentionally exposes the report viewer on all network interfaces.

It might be worth adding a note to the CLI help text and/or documentation along the lines of:

> The report viewer has no authentication and is intended for local use. When binding to `0.0.0.0` or another non-loopback address, ensure it is only exposed to trusted networks.

This would help users understand that the flag changes the viewer's network exposure and reduces the risk of accidental exposure in containerized or shared environments. Otherwise, the implementation looks clean, and the default behavior remains secure.

But regarding this decision, maybe @jplag/maintainer or @jplag/devteam can weigh in.

cli/src/main/java/de/jplag/cli/server/ReportViewer.java
```diff
@@ -40,8 +42,23 @@ public class ReportViewer implements HttpHandler {
      * @param resultFile The result file to use for the report viewer
      * @param port The port to use for the server. You can use 0 to use any free port.
      * @throws IOException If the result file cannot be read
+     * @deprecated Use {@link #ReportViewer(File, int, InetAddress)} instead
      */
+    @Deprecated(since = "7.0.0", forRemoval = true)
```
TwoOfTwelve: Since this is internal in the CLI module, we can just remove this method. As far as I know we don't consider this part of JPlags public API.

Unless there are still a lot of tests that use this method, of course

cli/src/main/java/de/jplag/cli/JPlagRunner.java
```diff
@@ -38,23 +40,61 @@ public static JPlagResult runJPlag(JPlagOptions options) throws ExitException {
      * @param resultFile is the result file to pass to the server. May be null.
      * @param port is the port to open the server on.
      * @throws IOException if the internal server throws an exception
+     * @deprecated Use {@link #runInternalServer(File, int, InetAddress)} instead
      */
+    @Deprecated(since = "7.0.0", forRemoval = true)
```
TwoOfTwelve: This could also be deleted

cli/src/main/java/de/jplag/cli/JPlagRunner.java
```diff
@@ -38,23 +40,61 @@ public static JPlagResult runJPlag(JPlagOptions options) throws ExitException {
...
+     * @throws IOException if the internal server throws an exception
+     */
+    public static void runInternalServer(File resultFile, int port, InetAddress bindAddress) throws IOException {
         if (!ReportViewer.hasCompiledViewer()) {
             logger.warn("The report viewer is not available. Check whether you compiled JPlag with the report viewer.");
             return;
         }
-        ReportViewer reportViewer = new ReportViewer(resultFile, port);
+
+        if (bindAddress.isAnyLocalAddress()) {
```
TwoOfTwelve: I'd prefer a check for isLoopbackAddress() here. The user could also input their public IP instead of a wildcard and the message would not be triggered. The messaged could be changed to may be accessible from any network.

TwoOfTwelve (changes requested):
Looks good to me. Just a few minor suggestions.

cli/src/main/java/de/jplag/cli/JPlagRunner.java
Kuznetsov-Artyom: Good comment, thx!

cli/src/main/java/de/jplag/cli/server/ReportViewer.java
Kuznetsov-Artyom: If these changes are really just part of the internal API, then yes, it doesn't make much sense

TwoOfTwelve (approved):

tsaglam (approved):

PR #3020, opened 2026-06-18
Temporarily deactivate windows e2e tests

Kr0nox:
Till we figure out the issue, we deactivate windows e2e tests


robinmaisch (approved):
Thanks! It is a pity that the windows process does not terminate.

PR #3039, opened 2026-07-01
Add Java Babylon language module

JFronny:
Another approach for structural normalization in the Java language module.
Relies on the in-development code model API from [Project Babylon](https://openjdk.org/projects/babylon/).
Draft until that API has been stabilized, at which point this PR, after being updated to align with expected API changes, should offer a more lightweight alternative to other normalization approaches.
(For now, this needs a custom JDK build available [here](https://github.com/JFronny/jdk-babylon/releases/tag/artifacts))


language-api/src/main/java/de/jplag/LanguageLoader.java
```diff
@@ -52,6 +62,15 @@ public static synchronized Map<String, Language> getAllAvailableLanguages() {
         return cachedLanguageInstances;
     }
 
+    private static boolean tryHasNext(Iterator<Language> iterator) {
```
TwoOfTwelve, line 65: Is this method unused?

language-testutils/src/test/java/de/jplag/testutils/LanguageModuleTest.java
```diff
@@ -43,7 +43,10 @@
  */
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 public abstract class LanguageModuleTest {
-    private static final Path DEFAULT_TEST_CODE_PATH_BASE = Path.of("src", "test", "resources", "de", "jplag");
+    /**
+     * Default base directory for test code.
+     */
+    public static final Path DEFAULT_TEST_CODE_PATH_BASE = Path.of("src", "test", "resources", "de", "jplag");
```
TwoOfTwelve: I don't think making this public is good in terms of code style. If you need this in multiple places, it would be better to put it somewhere it can be referred to from both places.

languages/golang/src/main/antlr4/de/jplag/golang/grammar/GoLexer.g4
```diff
@@ -34,190 +34,488 @@
  * https://golang.org/ref/spec
```
TwoOfTwelve, line 34: Don't reformat files that you didn't change. ANTLR files in particular should not be reformatted, since we just copy them from the ANTLR repository

languages/golang/src/main/java/de/jplag/golang/grammar/GoParserBase.java
```diff
@@ -1,206 +1,26 @@
-// CHECKSTYLE:OFF
 package de.jplag.golang.grammar;
```
TwoOfTwelve, line 2: Again. Do not commit files you didn't change (I assume you didn't). In this case it would have led to you overriding changes made to this file since you started

languages/golang/src/main/java/de/jplag/golang/GoListener.java
```diff
@@ -253,8 +252,7 @@ private void objectCreation() {
 
         visitCompositeLitChild(TypeNameContext.class).map(NAMED_TYPE_CONSTRUCTOR);
         visitCompositeLitDelegate(TypeNameContext.class).map(NAMED_TYPE_BODY_BEGIN, NAMED_TYPE_BODY_END);
-        visit(TypeNameContext.class, context -> hasAncestor(context, InterfaceTypeContext.class, GoParser.TypeDefContext.class,
-                MethodSpecContext.class, FieldDeclContext.class)).mapRange(TYPE_CONSTRAINT);
+        visit(TypeNameContext.class, context -> context.parent instanceof InterfaceTypeContext).mapRange(TYPE_CONSTRAINT);
```
TwoOfTwelve: Did you commit this by accident?

languages/java-babylon/src/main/java/de/jplag/java/babylon/extractor/CodeModelExtractorImpl.java
```diff
@@ -0,0 +1,33 @@
...
+     * Create a new instance.
+     * @param task the current compilation task
+     */
+    public CodeModelExtractorImpl(JavaCompiler.CompilationTask task) {
+        this.task = task;
+    }
+
+    @Override
+    public Optional<CoreOp.FuncOp> toOp(MethodTree methodTree) {
+        // if no Op can be constructed, that is an error that should not be construed with deliberate omission
```
TwoOfTwelve: I'd avoid comments like this unless they are absolutely necessary to understand the code. In this case I don't really understand what you are trying to say. If you cannot construct the operation, I assume that refers to the orElseThrow part, you don't want to omit the method in the exception. Is the method ever omitted in the exception?

languages/java-babylon/src/main/java/de/jplag/java/babylon/extractor/ExtractionFailedException.java
```diff
@@ -0,0 +1,32 @@
+package de.jplag.java.babylon.extractor;
+
+import com.sun.source.tree.MethodTree;
+
+/**
+ * {@link RuntimeException} that signals that a {@link jdk.incubator.code.dialect.core.CoreOp.FuncOp} could not be
+ * extracted.
+ */
+public class ExtractionFailedException extends RuntimeException {
```
TwoOfTwelve: JPlag normally uses checked exceptions, so it would be better to stay consistent with that unless there is a good reason not to

languages/java-babylon/src/main/java/de/jplag/java/babylon/extractor/ExtractionFailedException.java
```diff
@@ -0,0 +1,32 @@
...
+    public ExtractionFailedException(MethodTree source) {
+        super("Can't resolve body of method: " + source.getName());
+        this.source = source;
+    }
+
+    /**
+     * Create a new instance.
+     * @param source the exception that caused this
+     */
+    public ExtractionFailedException(ExtractionFailedException source) {
```
TwoOfTwelve: In java the parent exception is usually called "cause". Not particularly important, but it's good to adhere to common phrasings

languages/java-babylon/src/main/java/de/jplag/java/babylon/pipeline/LazyPrepassExecutor.java
```diff
@@ -0,0 +1,23 @@
...
+
+import de.jplag.java.babylon.extractor.CodeModelExtractor;
+import de.jplag.java.babylon.transformer.TransformationStep;
+
+import com.sun.source.tree.CompilationUnitTree;
+
+/**
+ * A lazy prepass executor that always applies each previous step when an extracted model is requested.
+ */
+public final class LazyPrepassExecutor implements PrepassExecutor {
```
TwoOfTwelve, line 13: Is there any benefit for having three separate classes for this? Couldn't you just use the hybrid one and pass the interval as a parameter?

The way I understand it, 0 as the interval would make it lazy and 1 would make it eager

languages/java-babylon/src/main/java/de/jplag/java/babylon/pipeline/TransformationPipeline.java
```diff
@@ -0,0 +1,86 @@
...
+     * Creates a new pipeline wrapping some steps.
+     * @param steps the steps to wrap
+     * @param prepassExecutor the prepass executor for this pipeline
+     * @throws IllegalArgumentException if the sequence of steps is not well-formed
+     */
+    public TransformationPipeline(List<? extends TransformationStep<?>> steps, PrepassExecutor prepassExecutor) {
+        this.steps = List.copyOf(steps);
+        this.prepassExecutor = prepassExecutor;
+
+        // ensure all dependencies are fulfilled
```
TwoOfTwelve: You could remove this comment and move the code to a method called something like "ensureDependenciesFulfilled". Comments like this often end up being wrong after some time, since developers tend to not touch them when changing the code

languages/java-babylon/src/main/java/de/jplag/java/babylon/tokenizer/impl/FullBabylonTokenizer.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * {@link BabylonTokenizer.Provider} implementation that fully outputs all {@link Op}s as tokens without further
+ * interpretation.
+ */
+public class FullBabylonTokenizer extends AbstractBabylonTokenizer {
+    /**
+     * Identifier of this tokenizer.
+     */
+    public static final String IDENTIFIER = "full";
```
TwoOfTwelve, line 22: I don't really like these IDENTIFIER variables, it feels a little bit like a hack. I'm not sure if there is an easy way to remove them, though.

If you need them for logging, couldn't you just override toString()? And as a unique identifier for each one, you could just use the class-instance or the class-name. I think that would make the code a bit cleaner.

languages/java-babylon/src/main/java/de/jplag/java/babylon/transformer/impl/CopyPropagationTransformer.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public String getIdentifier() {
+        return IDENTIFIER;
+    }
+
+    @Override
+    public Block.Builder acceptOp(Block.Builder builder, Op op) {
+        switch (op) {
```
TwoOfTwelve, line 52: This code is pretty hard to read. I get that you don't have control over the API you are using, but all the abbreviations make it almost impossible to tell what's going on at a glance. Maybe you could split this up into separate methods with helpful names?

languages/java-babylon/src/main/java/de/jplag/java/babylon/transformer/impl/StreamFusionTransformer.java
```diff
@@ -0,0 +1,423 @@
...
+
+    private static final ClassType LIST = ClassType.J_U_LIST;
+    private static final MethodRef LIST_NEW = MethodRef.constructor(ArrayList.class);
+    private static final MethodRef LIST_ADD = method(List.class, "add", boolean.class, Object.class);
+
+    private Block.Builder addAll(Block.Builder builder, Collect pipeline) {
+        Op.Location location = pipeline.source().location();
+        Value resultVariable = switch (pipeline) {
+            case Collect.ToList toList -> {
+                Value newHolder = place(builder, location, new_(LIST_NEW));
```
TwoOfTwelve, line 258: This entire class is very complex. I think for a transformation step, that is fine, since it's pretty clear what the code is supposed to do, but maybe put the case-blocks into separate methods, if possible, so it's easier to see the structure of your code.

You could also put some javadoc to private methods to explain what they do.

languages/java-babylon/src/main/java/de/jplag/java/babylon/transformer/TransformationStepLoader.java
```diff
@@ -0,0 +1,110 @@
...
+     */
+    public static Map<String, TransformationStep<?>> getAllAvailableTransformationSteps() {
+        if (cachedStepInstances == null) {
+            synchronized (TransformationStepLoader.class) {
+                if (cachedStepInstances == null) {
+                    Map<String, TransformationStep<?>> steps = new TreeMap<>();
+                    Set<String> skipped = new HashSet<>();
+
+                    if (stepLoader == null)
+                        // noinspection unchecked,rawtypes - unfortunately, this seems to be needed to use ServiceLoader with a generic class
```
TwoOfTwelve, line 47: Yeah, that is sadly a downside of the java type-system. The alternative would be to use the un-typed class and cast it whenever you load the actual instances. I think your version is preferable, though.

languages/java-babylon/src/main/java/de/jplag/java/babylon/BabylonDSL.java
```diff
@@ -0,0 +1,331 @@
...
+import jdk.incubator.code.dialect.core.Inliner;
+import jdk.incubator.code.dialect.java.JavaOp;
+import jdk.incubator.code.dialect.java.JavaType;
+import jdk.incubator.code.extern.OpWriter;
+
+/**
+ * A set of utility methods for working with Babylon code models. Designed to be used by "implementing" this interface
+ * (inspired by ASM).
+ */
+public interface BabylonDSL {
```
TwoOfTwelve: I think it's more common to do have utility methods as public static methods. This should work to, though

languages/java-babylon/src/main/java/de/jplag/java/babylon/JavaBabylonLanguage.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public BabylonOptions getOptions() {
+        return options;
+    }
+
+    @Override
+    public boolean supportsNormalization() {
+        return false; // for now
```
TwoOfTwelve: This comment does not really provide any benefit

languages/java-babylon/src/main/java/de/jplag/java/babylon/JavaBabylonLanguage.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public String getIdentifier() {
+        return super.getIdentifier() + "-babylon";
+    }
+
+    @Override
+    public int minimumTokenMatch() {
+        return super.minimumTokenMatch();
```
TwoOfTwelve: Since your language module behaves very different from the "normal" java module, you might need a different default value here. Since you de-obfuscate a lot I could imagine a higher value works better, but you'd have to make some experiments to determine that.

languages/java/src/main/java/de/jplag/java/JavacAdapter.java
```diff
@@ -99,4 +96,22 @@ private List<ParsingException> processErrors(DiagnosticCollector<Object> listene
         }).toList();
     }
 
+    protected boolean shouldAnalyze(final CompilationTask task) {
+        return false;
+    }
+
+    protected void handle(Iterable<? extends CompilationUnitTree> trees, Parser parser, SourcePositions positions,
```
TwoOfTwelve: I don't really like the methods called "handle". Can you change those to something more expressive? In this case maybe "scanCompilationUnits"

TwoOfTwelve (changes requested):
You code looks generally good. I made some suggestions, mostly just simple code-style things.

I only skimmed through the test code. For that you could consider putting all the long strings into files, but it's also fine the way it is.

Concerning your approach, my biggest concern would be the running time of the module. Also there might be an increase in false-positives, since some of your transformations feel pretty aggressive to me. I assume both of those are covered by your evaluation.

language-api/src/main/java/de/jplag/LanguageLoader.java
JFronny: `getAllAvailableLanguages` is modified to use this method to handle the case where a language implementation (=`java-babylon`) requires a newer JDK version.\
Since this PR shouldn't get merged until Babylon is upstream and probably won't get merged until it is in an LTS release, which, based on the track record, this project would likely switch to, I can also remove this change if you prefer.\
For now, the benefit is that builds with this PR still work on normal JDKs.

language-testutils/src/test/java/de/jplag/testutils/LanguageModuleTest.java
JFronny: Done

languages/java-babylon/src/main/java/de/jplag/java/babylon/extractor/CodeModelExtractorImpl.java
JFronny: This method returns an `Optional` and I don't think the reason I use what amounts to`Optional.of(optional.orElseThrow())` is self-evident.\
The method returns `Optional` so that `TransformingCodeModelExtractor` can use the nullable result of `transformation.apply` without surprising users of the API.
I have written a new comment that hopefully explains this more clearly.

languages/java-babylon/src/main/java/de/jplag/java/babylon/extractor/ExtractionFailedException.java
JFronny: This is used in lambdas, which is why it was unchecked.\
I have now migrated the relevant code to not use lambdas.

languages/java-babylon/src/main/java/de/jplag/java/babylon/extractor/ExtractionFailedException.java
JFronny: Done

languages/java-babylon/src/main/java/de/jplag/java/babylon/pipeline/LazyPrepassExecutor.java
JFronny: Yes, and ideally I will find that just one executor is sufficient during the evaluation, but I don't want to blindly guess which one performs the best.\
Regarding the current state, I expect that the hybrid executor, while superficially equivalent eg. to eager at 1, will perform worse in practice.

languages/java-babylon/src/main/java/de/jplag/java/babylon/pipeline/TransformationPipeline.java
JFronny: Done

languages/java-babylon/src/main/java/de/jplag/java/babylon/tokenizer/impl/FullBabylonTokenizer.java
JFronny: In the case of tokenizers, these are passed to `AbstractBabylonTokenizer.Provider` where they are used to implement `getIdentifier`, which powers the `TokenizerLoader`.\
Since users can specify the tokenizer they want in the CLI (which is necessary to support both low-level and high-level java code models), using the class names seems like it could easily cause unintentional breakage when shuffling classes around.\
Identifiers are also in `public static final` fields so that test code can use them without copying the string, which should make sure it stays up-to-date.

languages/java/src/main/java/de/jplag/java/JavacAdapter.java
JFronny: Done

languages/java-babylon/src/main/java/de/jplag/java/babylon/JavaBabylonLanguage.java
JFronny: Since this probably won't ever support TSN, I have removed the comment.

languages/java-babylon/src/main/java/de/jplag/java/babylon/JavaBabylonLanguage.java
JFronny: I intend to do so as part of my evaluation

languages/java-babylon/src/main/java/de/jplag/java/babylon/transformer/TransformationStepLoader.java
JFronny: If I understand the alternative correctly, that would result in a `ServiceLoader<TransformationStep>`, which would mean I would have to put the `noinspection` rules at the field instead (it would also cause the result of the stream pipeline below to be `List<TransformationStep>` without further trickery, meaning that line would require `noinspection` as well.\
Unless I am misunderstanding or a better alternative comes up, I unfortunately think the current hack is the best available option.

languages/java-babylon/src/main/java/de/jplag/java/babylon/transformer/impl/StreamFusionTransformer.java
JFronny: I have added some explanations in javadoc comments.

languages/java-babylon/src/main/java/de/jplag/java/babylon/transformer/impl/CopyPropagationTransformer.java
JFronny: It could be argued that this transformer tries to do too many things in one place. Unfortunately, splitting this up would worsen the UX for anyone manually specifying transformations even more.\
For now, I have added some additional comments.

languages/java-babylon/src/main/java/de/jplag/java/babylon/BabylonDSL.java
JFronny: I guess. I have replaced it with a utility class to align better with other parts of JPlag

languages/golang/src/main/antlr4/de/jplag/golang/grammar/GoLexer.g4
JFronny: This was due to an issue I had with CI after merging go changes from upstream, causing me to revert them. Reapplying the commits didn't break CI now, so this file should now be unchanged.

languages/golang/src/main/java/de/jplag/golang/GoListener.java
JFronny: This was due to an issue I had with CI after merging go changes from upstream, causing me to revert them. Reapplying the commits didn't break CI now, so this file should now be unchanged.

languages/golang/src/main/java/de/jplag/golang/grammar/GoParserBase.java
JFronny: This was due to an issue I had with CI after merging go changes from upstream, causing me to revert them. Reapplying the commits didn't break CI now, so this file should now be unchanged.

languages/java-babylon/src/main/java/de/jplag/java/babylon/tokenizer/impl/FullBabylonTokenizer.java
TwoOfTwelve: I know what you mean, but I cannot shake the feeling this is kind of weird. Since we aren't merging this for now, there is no reason to change it. If we do in the future, maybe annotations would be a more elegant solution?

PR #3055, opened 2026-07-15
Added antlr grammar licenses to jar

TwoOfTwelve:
Ensures the licenses of the antlr grammars are well visible in the jar file. To achieve this a separate file containing the license text is added next to the antlr files. This is usually a copy of the license comment on top of the antlr file. For two grammars it just says MIT license in the readme. For these the base MIT license text is used. Since each grammar has its own license, I think it's best to keep the license as close as possible to the grammar.

The documentation has also been expanded to reflect this.


Kr0nox:
I am not fully in the loop with the naming of license files, but should we name them just `LICENSE` or is `GRAMMAR_LICENSE` okay?

TwoOfTwelve:
I think it's more common to call licenses just LICENSE. I wanted to make it obvious that the license is not one of our licenses, but one we just re-distribute, though.

docs/4.-Adding-New-Languages.md
```diff
@@ -99,6 +99,10 @@ As the table shows, much of a language module can be reused, especially when usi
 
 JPlag provides a small framework to make it easier to implement language modules with ANTLR
 
+### License
+
+Make sure to include the license of the grammar used in a file called GRAMMAR-LICENSE next to the antlr files. This ensures they are part of our binary distribution.
```
Kr0nox: this is still the old file name

robinmaisch (approved):
While we are at it, we should ensure uniform captialization of ANTLR in the Wiki.
Thank you!

PR #2835, opened 2026-02-20
Add Java CPG language module

robinmaisch:
This PR contains a Java module based on the Fraunhofer AISEC [CPG](https://github.com/fraunhofer-aisec/cpg) framework. This module was designed by @robinmaisch and updated to version 8.3.0 of CPG by @LundiNord.

It enables structural normalization with respect to syntactic alternatives, thereby mitigating a range of typical refactoring attacks on software plagiarism detection systems such as JPlag.

PR #1659, opened 2024-03-19
Add Java-CPG language module

robinmaisch:
This Draft PR contains the new "Java Code Property Graph" module.
It uses a CPG library and graph transformations to normalise substructures of the code prior to tokenization, which should lead to a higher resilience against refactoring attacks.

It also serves to fix SonarCloud errors before a proper PR is opened.

PR #3056, opened 2026-07-15
Write tokenization duration to run information

JFronny:
Currently, the `executionTime` of the `RunInformation` saved in the result file by `ReportObjectFactory.writeRunInformation` is documented to contain the "total time taken for execution". However, it is sourced from `JPlagResult.durationInMillis` which is populated by `LongestCommonSubsequenceSearch` (and `MatchMerging` if enabled), which corresponds only to the comparison phase.
This PR adds a second field carrying the duration of the parsing/token extraction phase, which is useful when analyzing the performance impact of features like token sequence normalization or alternative language module implementations like https://github.com/jplag/JPlag/pull/2835 https://github.com/jplag/JPlag/pull/2766 https://github.com/jplag/JPlag/pull/1659 or https://github.com/jplag/JPlag/pull/3039.
(The naming of `RunInformation.executionTime` is maintained to support existing scripts that parse this file)


reporting/src/main/java/de/jplag/reporting/reportobject/model/RunInformation.java
```diff
@@ -10,9 +10,10 @@
...
  * @param dateOfExecution The date and time the run was executed, formatted as a string.
- * @param executionTime The total time taken for execution in milliseconds.
+ * @param tokenizationTime the time taken for tokenization in milliseconds.
+ * @param executionTime The time taken for comparison in milliseconds.
  * @param totalComparisons The total number of comparisons performed during the run.
  */
-public record RunInformation(Version version, List<FailedSubmission> failedSubmissions, String dateOfExecution, long executionTime,
-        int totalComparisons) {
+public record RunInformation(Version version, List<FailedSubmission> failedSubmissions, String dateOfExecution, long tokenizationTime,
+        long executionTime, int totalComparisons) {
```
Kr0nox: I would suggest renaming `executionTime` to `comparisonTime` to make clear it is not the entire execution time.

> (The naming of RunInformation.executionTime is maintained to support existing scripts that parse this file)

Is not really a concern, as 1) the report file format is not part of the public api and 2) the next release will be a major release anyways

Kr0nox (commented):
Please also update the report viewer part to mirror these changes (so adding them to the RunInformation interface in the model and displaying them in the InformationView)

TwoOfTwelve (approved):

reporting/src/main/java/de/jplag/reporting/reportobject/model/RunInformation.java
JFronny: Done.

report-viewer/parser/tests/factories/assets/ValidRunInformation.json
```diff
@@ -5,6 +5,7 @@
     { "submissionId": "failedTest", "submissionState": "CANNOT_PARSE" }
   ],
   "dateOfExecution": "06/05/25",
-  "executionTime": 471,
+  "tokenizationTime": 471,
+  "comparisonTime": 471,
```
Kr0nox, line 9: Change one of them to a different value, so we can make sure we never accidentally mess something up with the numbers. Make sure the numbers are visibly different, e.g. change one to them to 999

Kr0nox:
How excactly do we define tokenization time? If i understand the code correctly this just takes the sum of the time each tokenization took. From tokenization time I would have intuitivly though it was the time between the start of the tokenization and the end of the tokenization.

Do we calculate the comparison time the same way? I believe there it is just the difference between start and end. It should defineatly be consistent between the two
@TwoOfTwelve @jplag/maintainer what are your thought?

JFronny:
> How excactly do we define tokenization time? If i understand the code correctly this just takes the sum of the time each tokenization took. From tokenization time I would have intuitivly though it was the time between the start of the tokenization and the end of the tokenization.

I don't think that this is necessarily the ideal thing to measure, but I agree that changing that should probably go in a different PR.
I have modified the code to just use the wall clock millis before and after the parallel tokenization, as is done with comparison.

> Do we calculate the comparison time the same way? I believe there it is just the difference between start and end. It should defineatly be consistent between the two

Comparison does use wall clock millis before and after the full parallel pipeline.

Kr0nox:
@JFronny If you quickly update your branch with develop, then sonar should not fail anymore and this is ready to merge

Kr0nox (approved):

Issue #1569, opened 2024-02-19
Open tasks regarding the match merging feature


tsaglam:
The following tasks should be addressed in the future:
- For now: Conservative parameters and default enable to not influence the performance too much (#1430)
- Neighbor length lower threshold check should be in the options and not in the GST class
- Future: Different strategies for hyperparameters: Automatic, manual, off

Issue #1430, opened 2023-12-11
Match merging has performance issues with submission sets of larger programs


tsaglam:
When submissions are > 1000 LoC, the performance degrades. I have not yet investigated whether this is CPU load or due to memory/paging issues.

tsaglam:
I profiled match merging for a large dataset (>400 submissions, ~1500 loc each, with basecode on):

<details><summary>Method profiling</summary>

![image](https://github.com/jplag/JPlag/assets/4396919/6b93f361-53dd-4cd0-862a-bf46536cb10d)

</details>

robinmaisch:
@tsaglam, do you think match merging performance is sufficient now, or is there still significant room for improvement?

tsaglam:
@robinmaisch, the main problem is that the GST overhead of match merging is the main performance impact remaining. This mainly affects large datasets and/or larger programs. The only way around that would probably be some kind of auto-parameterization, which reduces the match-merging aggressiveness when the input data is too large. As long as match merging is off per default, however, this is not a critical issue.

PR #3059, opened 2026-07-18
Improve match merging memory allocation and performance

tsaglam:
For larger datasets (especially many large programs), subsequence match merging (SMM) consumes a lot of memory due to the copying of lists. In a JVM with less memory allocated, this also greatly increases the runtime. This PR improves the SMM performance by eliminating unnecessary submission copies, replacing the O(n²) index lookup with a constant-time IdentityHashMap, and using subList views instead of copying token lists. However, runtime is still affected when SMM is enabled due to the greedy string tiling overhead.

Contributes towards #1430 (and thus also #1569).

**Example Measurements:**

- (macOS, M4 Pro Chip, JDK 26, -Xmx16g)
- 367 programs with ~1500 LOC each
- Allocation measured via `ThreadMXBean.getTotalThreadAllocatedBytes()`

| **Match Merging Algorithm** | **Runtime Mean** | **Alloc/iter** |
|---|---:|---:|
| Current | 2.72s | 22.1 GB |
| New Version | 1.13s | 2.7 GB |

| **GST overhead** | **Runtime Mean** |
|---|---:|
| Merging disabled | 5.7s |
| Merging enabled | 2.8 min |

### DIFF
--- a/core/src/main/java/de/jplag/merging/MatchMerging.java
+++ b/core/src/main/java/de/jplag/merging/MatchMerging.java
@@ -2,7 +2,9 @@ package de.jplag.merging;
 
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.IdentityHashMap;
 import java.util.List;
+import java.util.Map;
 
 import de.jplag.JPlagComparison;
 import de.jplag.JPlagResult;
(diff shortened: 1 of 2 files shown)


uuqjz (commented):
Nice improvement!

robinmaisch (approved):
Really nice! You end up wondering why we did it like that before.

PR #3068, opened 2026-07-23
Parse File Lists instead of Sets to persist file order in Submissions

### LINKED ISSUE #3001: Order of file tokenization depends on submission set
### Bug Description

In `Submission.java`, a `HashSet` of the submission files is created as input for the method `Language::parse`. The iterator of the `HashSet` class, which determines the order in which the files are tokenized, seems to depend on internal state, resulting in a different file order when I extract a subset of submissions into another directory compared to the complete submission set. Specifically, the token order does not usually correspond to the file order in `Submission::files`, which seems counterintuitive.

As `GreedyStringTiling` is implemented to prefer earlier matches (w.r.t. the token list), a different file tokenization order may result in different matching, so we should set a fixed file order for tokenization. I therefore propose to change the `Language` API to accept `List`s of `File`s instead of `Set`s, avoiding randomness in file order for tokenization and enabling consistency with the file order in `Submission::files`.

Find attached two screenshots displaying this behavior: Two comparisons of the same submission pair result in a different matching (118 vs. 120 matches) using the same algorithm, but the submissions are located inside different subsets of the submission set. The resulting average similarity differs by ~0.54 percentage points: 40.172 % vs. 40.715 %.

Image

Image

### JPlag Version

develop/7.0.0-SNAPSHOT

### Operating System

Windows 11

### Java Version

Temurin 25

robinmaisch:
This PR changes the `Language` API to parse `List`s of `File`s instead of `Set`s. This fixes the current behavior where a submission's token sequence changes depending on the submission set; this, in turn, led to inconsistent behavior while matching submissions.

Closes #3001.

### DIFF
--- a/core/src/main/java/de/jplag/Submission.java
+++ b/core/src/main/java/de/jplag/Submission.java
@@ -11,7 +11,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -42,7 +41,7 @@ public class Submission implements Comparable<Submission> {
     private final String name; // identifier for the submission (a directory or file name).
     private final File submissionRootFile; // Root of the submission, a director or file (including the subdir if used).
     private final boolean isNew; // old submissions are only checked against new ones.
-    private final Collection<File> files;
+    private final List<File> files;
     private final Language language;
 
     private SubmissionState state; // whether an error occurred during parsing or not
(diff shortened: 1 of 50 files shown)


Kr0nox:
In case we want to preserve that the input should be a Set an alternative to the List would be a SortedSet

JFronny:
This seems to cause the result to depend on the order of `File.list` results, which is [not guaranteed](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/io/File.html#list()) (and probably depends on the order provided by the file system, which is [also not guaranteed](https://reproducible-builds.org/docs/stable-inputs/)). `SortedSet` would solve that, but would also break the implied DFS order. Sorting the array would also work.

Kr0nox:
Plan is to change the File List to an Iterable and handle the sorting after parsing

robinmaisch:
We wil solve this by having the Parsing result be a more complex object that handles sorting

PR #3070, opened 2026-07-26
Avoid virtual threads for LCS

JFronny:
This partially reverts #2412

Using [this](https://github.com/user-attachments/files/30390168/LongestCommonSubsequenceSearch.java) slightly janky measuring code ([results](https://github.com/user-attachments/files/30390187/times.csv)), I measured a decent improvement (from 425 to 364 millis for both the Progpedia19 and Progpedia56 variations from [here](https://zenodo.org/records/18089904)) when using the common fork join pool via `Stream.parallel` instead of virtual threads.
This seems plausible to me since LCS shouldn't contain any IO code that virtual threads could benefit from, which should mean they provide no benefit here.
I assume the performance improvements measured by others for the original PR are from the other changes made there.

### DIFF
--- a/core/src/main/java/de/jplag/comparison/LongestCommonSubsequenceSearch.java
+++ b/core/src/main/java/de/jplag/comparison/LongestCommonSubsequenceSearch.java
@@ -3,11 +3,6 @@ package de.jplag.comparison;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -109,28 +104,17 @@ public class LongestCommonSubsequenceSearch {
             compareSubmissionsToBaseCode(coreAlgorithm, submissionSet);
         }
 
-        // Compare all submission pairs in parallel:
         List<SubmissionTuple> tuples = buildComparisonTuples(submissionSet.getSubmissions());
-        List<JPlagComparison> comparisons = new ArrayList<>();
+
+        // Compare all submission pairs in parallel:
+        List<JPlagComparison> comparisons;
         ProgressBar progressBar = ProgressBarLogger.createProgressBar(ProgressBarType.COMPARING, tuples.size());
-        try (ExecutorService executor = Executors.newVirtualThreadPerTaskExecutor()) {
-            List<Future<Optional<JPlagComparison>>> futures = tuples.stream().map(tuple -> executor.submit(() -> {
+        try {
+            comparisons = tuples.stream().parallel().map(tuple -> {
                 Optional<JPlagComparison> result = compareSubmissions(coreAlgorithm, tuple.left(), tuple.right());
                 progressBar.step();
                 return result;
-            })).toList();
-
-            executor.shutdown();
-            if (!executor.awaitTermination(24, TimeUnit.HOURS)) {
-                throw new ComparisonException("Comparison timed out.");
-            }
-
-            for (Future<Optional<JPlagComparison>> future : futures) {
-                future.get().ifPresent(comparisons::add);
-            }
-        } catch (InterruptedException | ExecutionException e) {
-            Thread.currentThread().interrupt();
-            throw new ComparisonException("Error during comparison algorithm.", e);
+            }).filter(Optional::isPresent).map(Optional::get).toList();
         } finally {
             progressBar.dispose();
         }


tsaglam (commented):
Here a few comments from my side:

- Progpedia is not a realistic dataset for performance benchmarking, as it contains very few programs, that are all relatively small. You need a larger dataset for that. For small datasets, the perfomance is irrelevant, as they are analysed quickly anyways.
-  While the overall gain in #2412 comes from all measures combined, the change with virtual threads did indeed improve perfomance.
- While the GST algorithm is not IO-bound, there are other advantages of virtual threads and in particular `newVirtualThreadPerTaskExecutor`:
    - Virtual threads are particularly light-weight and thus memory efficient. But this is not as important here.
    - The real differentiator is granularity: The VT scheduler distributes work at the granularity of a single pair. A carrier thread that finishes an expensive pair immediately takes the next one from the queue. The makespan tail is bounded by the single most expensive pair.
    - A parallel stream splits the tuple list into fixed-size chunks, each executed atomically. Because the tuple builder emits row-major (submission i paired with all i+1..n), contiguous chunks tend to cluster large submissions together. If the largest few submissions happen to fall into the same chunk, that chunk dominates the tail. The makespan degrades from "worst pair" to "worst chunk". This is a gap on the order of 3000× in task granularity.
    - This is mainly the reason i observed a performance gain back then, but also why I would oppose this PR.

**TLDR:** I do not think the measurements here are sufficient, and the advantage of the virtual thread executor is the granularity.

tsaglam (changes requested):
Ran a small benchmark on a larger dataset:

- **Hardware:** Apple M4 Pro (12 cores)
- **JVM:** OpenJDK 26.0.1, default flags
- **Dataset:** 367 Java submissions, 67K pairs, 561 KLOC
- **Protocol:** 3 warmup + 15 timed runs per version, fresh JVM per run

| Version | Mean | Median | Stdev | Min | Max |
|---|---|---|---|---|---|
| HEAD (virtual threads) | 3,747 ms | 3,775 ms | 118 ms | 3,460 ms | 3,917 ms |
| PR (parallel stream) | 4,169 ms | 4,145 ms | 149 ms | 3,988 ms | 4,479 ms |

Virtual threads are 10.1% faster on mean, 8.9% on median, consistent with the 4-10% range reported in #2412. The parallel-stream version has a systematically higher floor. This matches the granularity hypothesis.

JFronny:
I now have a dataset of lrger submissions and will evaluate alternatives soon (hopefully tomorrow).

JFronny:
I have run some measurements on a larger dataset (104196 pairs with 649055
 total LOC, specifically prog-ws202526 A4/A and A5/A) following the same protocol you described (implemented via a bash `for` loop).

| variant        |   mean |   stddev |   min |   max |   q1 |   median |   q3 |
|:---------------|-------:|---------:|------:|------:|-----:|---------:|-----:|
| parallelStream |   1684 |       60 |  1581 |  1787 | 1642 |     1679 | 1734 |
| upstream       |   1788 |      119 |  1623 |  2117 | 1730 |     1777 | 1822 |
| forkjoinpoolCF |   1697 |       66 |  1601 |  1837 | 1651 |     1694 | 1739 |

(all values in milliseconds)

This is on OpenJDK 26.0.2 on a Ryzen 7600X system. (Perhaps the difference in results is related to the architecture?)
While this is a smaller improvement, `parallelStream` is still slightly faster.
With that said, I specifically believe that the choice of virtual threads is wrong, and do agree with your point that splits could be suboptimal.
(After all, a virtual thread is just a task submitted to a common forkjoinpool with a lot of extra overhead)

In the measuring code provided above, I have also implemented a variant that uses the common fork join pool directly and joins the futures instead of shutting down the executor.
I have included results for that in the table above (with a small change, see [LongestCommonSubsequenceSearch.java](https://github.com/user-attachments/files/30507136/LongestCommonSubsequenceSearch.java)).
I had not originally submitted that as it performed slightly worse than `parallelStream` on my machine (see the table) and is even less pretty (specifically if the timeout, which wasn't there before the PR mentioned above, is desired. If I can remove that, the code becomes much cleaner), but it does avoid virtual threads and still performs better than upstream.

If you prefer, I can update this PR to use that code instead.

robinmaisch:
True, virtual threads are the usual choice for IO-heavy concurrent tasks. The inconclusive measurements presented here, one pointing 5 percent in favor of one solution and the other 5 percent in favor of the other, do not support the claim that virtual threads are a generally "wrong choice" in this particular use case, and certainly not the reversal of the previous PR, which did seem a more performant solution on two different machines using two different datasets. Why the results diverge here is hard to judge from what we know.

Implying that we misinterpreted our results based on an unsuitable dataset of code snippets is not a good look; please be more careful with your conclusions.

tsaglam:
@JFronny  so no approach is universally faster. On Ryzen 7600X, parallel stream leads at 1684 ms, against 1788 for VT and 1697 for forkjoinpoolCF. On M4 Pro, VT leads at 3747 ms, against 4169 for parallel stream and 3876 for forkjoinpoolCF. Yes, hardware may be a slight difference (the Ryzen has only 6 physical cores vs. the 12 of the M4, the M4 is a unified-memory SoC design), but the greater factor probably is the dataset. The forkjoinpoolCF variant is also the most complex code: four extra imports, an explicit pool parameter, and nested try/catch for timeout handling.

I would like to return to the original question: what problem does removing virtual threads solve? You mentioned that virtual threads are just tasks on a common fork join pool with extra overhead, but that is not accurate. The executor created by `newVirtualThreadPerTaskExecutor` does not submit to the common pool. `VirtualThread.java:66` creates its own `ForkJoinPool` at class load time as a static field. That pool is configured with `asyncMode = true` (FIFO mode, line 418) and `parallelism = availableProcessors`. The common pool runs in LIFO mode with `parallelism = availableProcessors - 1` (`ForkJoinPool.java:3081, 4423`). Same class, different instances, fundamentally different scheduling.

The FIFO mode matters for this workload. LIFO is optimized for recursive divide and conquer where a task spawns subtasks and then joins them, which is exactly what `Stream.parallel()` does when it splits the tuple list into chunks. FIFO pulls tasks in submission order, so a carrier that finishes an expensive pair immediately takes the next queued pair. Once you get an unlucky chunk in LIFO mode you are stuck until all pairs in it complete. The scheduling policy is not cosmetic.

Virtual threads also are not only for I/O. The per task work stealing benefits any workload where task costs are uneven: if three submissions with 10k tokens each land in the same parallel stream chunk, that chunk runs serially and dominates the tail. With the VT executor each of those three is an independent task, so carriers steal the remaining cheap pairs while the expensive ones finish. The more variance in submission size, the larger the gap. This is exactly the pattern JPlag sees frequently in real datasets.

I do not see a case for change.

JFronny:
Thanks for the detailed response!
Regarding the purpose:
This PR was opened after looking at the code, being surprised by the choice of executor due to some previous experience on an unrelated project (hence the comment about it being a "wrong" choice above, perhaps saying it was surprising to me would have been more accurate?) and, using the best datasets I had set up at the time, measuring an improvement in performance when using a different executor.
Based on the assumption that the behavior I observed would be reproducible by others, I opened this.
Since your measurement with a different setup showed that my initial assessment was clearly skewed, I now agree that this PR does not seem to provide the benefit I expected.
Sorry for taking your time.

PR #3072, opened 2026-07-28
Zip as input

TwoOfTwelve:
A draft for using Zips as input. To do this the input structure is replaced by a virtual file system that wraps either the actual file system or the zip file.

This draft is not finished yet, but the basic structure can be reviewed


JFronny:
NIO already provides a VFS API including [a Zip file system](https://docs.oracle.com/en/java/javase/25/docs/api/jdk.zipfs/module-summary.html).
It seems like using that FS (or maybe even the full VFS API) could simplify the implementation.

Issue #2595, opened 2025-09-10
Future tasks for match frequency-based highlight extraction


tsaglam:
As soon as possible (after #2563 is merged):
- [ ] Remove less-performant strategies, keeping complete matches and windowing.
- [ ] Remove superfluous weighing function.
- [ ] Check and adapt default parameters if necessary.
- [ ] Parallelize the frequency calculation[^1].
- [ ] Check if `FrequencyAnalysisOptions` can be simplified with the record builder.
- [ ] Deduplicate FrequencyStrategy and FrequencyAnalysisStrategy by deleting the enum.
- [ ] Dissolve FrequencyUtil if possible
- [x] Dissolve MatchFrequency
- [x] Ensure it only runs when the flag is set (also only creates overhead/objects when flag enabled)
- [ ] ensure if it actually overrides the result (and ensure in JPlag.run the syntax makes that clear)
- [ ] Try to make `Submission.getSimilarityDivisor()` non-public again (if possible)
- [x] Check if renaming bug has been resolved (`isFrequencyAnalysisEnabled`)
- [ ] Check if public method should be private (e.g. MatchWeightCalculator.weightAllMatches is only called from tests)

In the future:
- [ ] Ensure compatibility with asymmetric matches (currently, tokens of first submissions in a match are used)
- [ ] Instead of enabling frequency analysis per CLI flag, it could be enabled by default but fed into a separate similarity metric (this is only viable if the performance is fine for large datasets).
- [ ] Especially rare matches could be visually highlighted in the comparison view to draw attention during human inspection.

[^1]: should be trivial for operations that work on a on-match or on-comparison basis.

PR #3088, opened 2026-08-01
Decouple frequency-weighted similarity into a supplementary metric

tsaglam:
This PR changes the frequency analysis feature so that it no longer overrides the similarity metric itself.
Instead, it is now supplied as a secondary metric in the report viewer, just as max similarity (contributes towards #2595).

There is, however, one **open design choice**: what to display in the viewer when the parameter is not used?
* A: Do not show the button for the secondary metric (currently implemented)
* B: Show the button, greyed out, with a mouseover telling the user to run JPlag with --frequency
* C: Show the button, but when clicking it, the values are empty in the comparison list.

Also: There is currently no real icon for the new metric.

**Report Viewer:**

image

image

**Details:**

- Remove the `useFrequencyWeighting` gate and field from `JPlagComparison`. `similarity()` always returns structural similarity.
- Add `WEIGHTED_SIMILARITY` to `SimilarityMetric`, delegating to `frequencyWeightedSimilarity`. Returns −1 when not computed.
- Export in JSON, top comparisons, distributions, and CSV.
- Add `WeightedSimilarityMetricType` to the frontend with N/A for −1. Register in sort, filter, and dropdowns.
- Add to distribution histogram and boxplot metric selectors.
- Hide the metric from all dropdowns when frequency analysis is off, using the existing `frequencyAnalysisOptions.enabled` flag.
- `CompleteMatchesStrategy` converts its result map directly instead of re-iterating all comparisons.
- Clamp negative distribution indices in `JPlagResult` to prevent `ArrayIndexOutOfBoundsException` on −1 values.


core/src/main/java/de/jplag/JPlagResult.java
```diff
@@ -190,6 +190,7 @@ private int[] calculateDistributionFor(List<JPlagComparison> comparisons, ToDoub
         for (JPlagComparison comparison : comparisons) {
             double similarity = similarityExtractor.applyAsDouble(comparison); // extract similarity: 0.0 <= similarity <= 1.0
             int index = (int) (similarity * SIMILARITY_DISTRIBUTION_SIZE); // divide similarity by bucket size to find index of correct bucket.
+            index = Math.max(index, 0);
             index = Math.min(index, SIMILARITY_DISTRIBUTION_SIZE - 1); // index is out of bounds when similarity is 1.0. decrease by one to count
                                                                        // towards the highest value bucket
             similarityDistribution[index]++; // count comparison towards its determined bucket.
```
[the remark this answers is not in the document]
tsaglam: The clamp exists to prevent `ArrayIndexOutOfBoundsException`.

Kr0nox:
I would suggest we make this more generic by having a list of included secondary metrics in the options.json file.
There is at least one more additional metric comming (comments) and it will be less cluttered when we might add even more metrics in the future

tsaglam:
@Kr0nox should be better now!

tsaglam:
@jplag/maintainer bump! ;)

report-viewer/ui-components/widget/comparisonTable/ComparisonTableFilter.vue
```diff
@@ -59,18 +59,30 @@
...
     default: false
+  },
+  secondaryMetrics: {
+    type: Object as PropType<Set<MetricJsonIdentifier>>,
+    default: () =>
+      new Set([
+        MetricJsonIdentifier.MAXIMUM_SIMILARITY,
+        MetricJsonIdentifier.LONGEST_MATCH,
+        MetricJsonIdentifier.MAXIMUM_LENGTH,
+        MetricJsonIdentifier.WEIGHTED_SIMILARITY
```
Kr0nox: As the weighted similarity is not calculated per default in the core, I think its better to remove it here and only have the ones that are gurenteed to be in the report here (so: MAX, LONG, LEN)

report-viewer/ui-components/widget/distributionVisualization/BoxPlotDiagramOptions.vue
```diff
@@ -22,8 +23,36 @@ const model = defineModel<MetricJsonIdentifier>('metric', {
...
-  MetricJsonIdentifier.MAXIMUM_SIMILARITY
-]
+const props = defineProps({
+  secondaryMetrics: {
+    type: Object as PropType<Set<MetricJsonIdentifier>>,
+    default: () =>
+      new Set([
+        MetricJsonIdentifier.AVERAGE_SIMILARITY,
+        MetricJsonIdentifier.MAXIMUM_SIMILARITY,
+        MetricJsonIdentifier.WEIGHTED_SIMILARITY
```
Kr0nox: Same thing as above

report-viewer/ui-components/widget/distributionVisualization/DistributionDiagramOptions.vue
```diff
@@ -42,10 +43,38 @@ const config = defineModel<DistributionChartConfig>({
...
 })
 
+const props = defineProps({
+  secondaryMetrics: {
+    type: Object as PropType<Set<MetricJsonIdentifier>>,
+    default: () =>
+      new Set([
+        MetricJsonIdentifier.AVERAGE_SIMILARITY,
+        MetricJsonIdentifier.MAXIMUM_SIMILARITY,
+        MetricJsonIdentifier.WEIGHTED_SIMILARITY
```
Kr0nox: see above

reporting/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
```diff
@@ -53,6 +54,7 @@ public class ReportObjectFactory {
     private static final Path SUBMISSION_MAPPINGS_FILE_NAME = Path.of("submissionMappings.json");
     private static final Path README_FILE_NAME = Path.of("README.txt");
     private static final Path OPTIONS_FILE_NAME = Path.of("options.json");
+    private static final Path SECONDARY_METRICS_FILE_NAME = Path.of("secondaryMetrics.json");
```
Kr0nox: instead of having a file just for this one thing: Make this a field called `secondaryMetrics` in a file called `reportInformation.json`

Kr0nox (changes requested):
This is what we discussed in our meeting

report-viewer/ui-components/widget/comparisonTable/ComparisonTableFilter.vue
tsaglam: Fixed, now only the default metrics.

report-viewer/ui-components/widget/distributionVisualization/BoxPlotDiagramOptions.vue
tsaglam: Done.

report-viewer/ui-components/widget/distributionVisualization/DistributionDiagramOptions.vue
tsaglam: Done.

reporting/src/main/java/de/jplag/reporting/reportobject/ReportObjectFactory.java
tsaglam: Is now a field, should be better now.

Kr0nox (approved):

robinmaisch (approved):
Excellent! Thank you very much.
I will just note here that during a test run on a fairly large submission set, frequency weighting exploded in runtime, exceeding 30 min, after which I terminated it. I think I found critical redundancies in some calculations, which I will address shortly.

PR #3093, opened 2026-08-08
Migrate from File to Path

TwoOfTwelve:
This PR migrates JPlag from java.io.File to java.nio.Path. This is mostly done so we can support Zip files as input more easily.

It also simplifies some code as a lot of classes have already used Path internally.

I changed all APIs to Path for now. We should discuss if we want to provide legacy APIs that still accept File and where. I don't think it is a big issue, since all a user has to do to adapt their code is to call File::toPath.

This would also allow us to use libraries like https://github.com/google/jimfs for temporary files, so we don't pollute the users file system.

I know this is a big change @jplag/maintainer, tell me if you don't like this. I think that nio is the more flexible library and that it's better than to adapt between File and Path a lot as we used to do in our tests.

There is currently one test failing, we should discuss how to handle that in the next meeting.


TwoOfTwelve:
Should not be merged before #3098

cli/src/test/java/de/jplag/cli/ReadmeCodeExampleTest.java
```diff
@@ -29,19 +31,19 @@ class ReadmeCodeExampleTest {
     @SuppressWarnings("java:S2699")
     void testReadmeCodeExample() {
         Language language = new JavaLanguage();
-        Set<File> submissionDirectories = Set.of(new File("/path/to/rootDir"));
-        File baseCode = new File("/path/to/baseCode");
+        Set<Path> submissionDirectories = Set.of(Path.of("path", "to", "rootDir"));
+        Path baseCode = Path.of("path", "to", "baseCode"); // TODO readme adapt
```
robinmaisch: Don't forget TODO it! :)

language-api/src/main/java/de/jplag/Token.java
```diff
@@ -26,36 +26,10 @@ public class Token {
     private final int length;
     private final int endLine;
     private final int endColumn;
-    private final File file;
+    private final Path file;
```
robinmaisch, line 29: So, should we rename this?

TwoOfTwelve: I decided to leave the variable names unchanged. I think they still fit as this path represents the file the token belongs to. Also, I think, in general you are supposed to name variables after what they represent, rather than their type.

cli/src/test/java/de/jplag/cli/ReadmeCodeExampleTest.java
TwoOfTwelve: Thanks. I thought I had forgotten a TODO, but my IDE told me there were none left :)

PR #3099, opened 2026-08-18
Adapted token position test format

### LINKED ISSUE #2980: Token position tests format
Currently, the token position tests calculate the length of a token as (end - start). This is wrong, as both are inclusive. Therefore, it should be changed to (end - start + 1).

All currently existing tests have to be adapted.

TwoOfTwelve:
The token position tests now treat the start end the end of a token both as inclusive. This means that the length is calculated as (end - start + 1). To reflect this the java token end positions have been adapted and the tests for the cpp language module have been changed.

Since these changes only affect the token positions I don't think they should be considered breaking.

Fixes #2980

### DIFF
--- a/languages/cpp/src/test/resources/de/jplag/cpp/tokenPositions/assign.cpp
+++ b/languages/cpp/src/test/resources/de/jplag/cpp/tokenPositions/assign.cpp
@@ -1,4 +1,4 @@
 > void main() {
 >     int x = 1;
-$         | ASSIGN 0
+$         | ASSIGN 1
 > }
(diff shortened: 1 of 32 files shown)


languages/cpp/src/test/resources/de/jplag/cpp/tokenPositions/assign.cpp
robinmaisch, line 1: Does the `ASSIGN` token represent the expression `x = 1` or just `=`? The length of 1 seems inconsistent with the position.

TwoOfTwelve: I created an issue for that, as I think this should not be part of this PR (#3137)

PR #3112, opened 2026-08-28
Configure Sonar to ignore test coverage of imported ANTLR Java files

robinmaisch:
Sonar keeps failing on PRs that include changes affecting language modules because of low test coverage for ANTLR files imported from the ANTLR/grammar-v4 project. This PR configures Sonar to ignore the test coverage on these files.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,6 +67,7 @@
         <sonar.moduleKey>${project.groupId}:${project.artifactId}</sonar.moduleKey>
         <sonar.organization>jplag</sonar.organization>
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
+        <sonar.coverage.exclusions>**/grammar/*.java</sonar.coverage.exclusions>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -67,6 +67,7 @@
         <sonar.moduleKey>${project.groupId}:${project.artifactId}</sonar.moduleKey>
         <sonar.organization>jplag</sonar.organization>
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
+        <sonar.coverage.exclusions>**/grammar/*.java</sonar.coverage.exclusions>
```
Kr0nox: Would it be better to check all subfolders of grammar too? or could that cause issues?
```suggestion
        <sonar.coverage.exclusions>**/grammar/**/*.java</sonar.coverage.exclusions>
```

robinmaisch: It seems there are currently no subfolders in any grammar directories. But there should be no harm in including them for the future👍

PR #3135, opened 2026-09-21
Remove redundant calculations to drastically improve weighting performance

robinmaisch:
With these optimizations, the runtime on my machine dropped:

- from >30 min. to 2 min. _for a complete run_ on a dataset of fairly large code submissions
- from 40 sec. to <0.5 sec. _for weighting only_ on a dataset of medium-sized code submissions

These are the core changes:
- to get the token types for a match, do not map the complete token list of a submission to token types (I wonder: could we not just operate on integers for the weighting?)
- to determine the `count` of all matches, do not iterate over all comparisons and repeat the calculation for every match, but collect the matches first and do the calculation once per match
- determine the `maxFrequency` of a submission set once instead of once for each comparison


core/src/main/java/de/jplag/frequency/MatchFrequencyEvaluator.java
```diff
@@ -25,20 +23,11 @@ public MatchFrequencyEvaluator(FrequencyAnalysisStrategy strategy) {
...
-        comparisons.parallelStream().forEach(comparison -> weightAllMatches(comparison, matchWeights));
-        return matchWeights;
-    }
-
-    private void weightAllMatches(JPlagComparison comparison, Map<List<TokenType>, Double> matchWeights) {
-        for (Match match : comparison.matches()) {
-            List<TokenType> matchTokens = FrequencyUtil.tokenTypesFor(comparison, match);
-            matchWeights.computeIfAbsent(matchTokens, strategy::calculateMatchCount);
-        }
+    Map<List<TokenType>, Double> getWeightMatches() {
```
TwoOfTwelve, line 28: I guess it is important that the map that is returned is a ConcurrentHashMap? In that case I think it would be better to make the return type ConcurrentHashMap or at least mention in the documentation that the map is thread-safe.

core/src/main/java/de/jplag/JPlagResult.java
```diff
@@ -190,9 +190,7 @@ private int[] calculateDistributionFor(List<JPlagComparison> comparisons, ToDoub
         for (JPlagComparison comparison : comparisons) {
             double similarity = similarityExtractor.applyAsDouble(comparison); // extract similarity: 0.0 <= similarity <= 1.0
             int index = (int) (similarity * SIMILARITY_DISTRIBUTION_SIZE); // divide similarity by bucket size to find index of correct bucket.
-            index = Math.max(index, 0);
-            index = Math.min(index, SIMILARITY_DISTRIBUTION_SIZE - 1); // index is out of bounds when similarity is 1.0. decrease by one to count
-                                                                       // towards the highest value bucket
+            index = Math.clamp(index, 0, SIMILARITY_DISTRIBUTION_SIZE - 1); // keep index within array bounds
```
TwoOfTwelve, line 193: I'd remove the comment, since the code expresses pretty much the same thing.

core/src/main/java/de/jplag/frequency/strategy/FrequencyAnalysisStrategy.java
```diff
@@ -17,9 +19,11 @@
 public abstract class FrequencyAnalysisStrategy {
 
     private final Map<List<TokenType>, Integer> matchCounts;
+    private final Set<List<TokenType>> matchRegistry;
 
     protected FrequencyAnalysisStrategy() {
         this.matchCounts = new ConcurrentHashMap<>();
+        matchRegistry = new HashSet<>();
```
TwoOfTwelve, line 26: Does this have to be thread-safe too?

TwoOfTwelve (changes requested):

core/src/main/java/de/jplag/frequency/MatchFrequencyEvaluator.java
robinmaisch: It is mostly important for construction, not for usage.

core/src/main/java/de/jplag/frequency/strategy/FrequencyAnalysisStrategy.java
tsaglam: I think so, should maybe be
```suggestion
                matchRegistry = ConcurrentHashMap.newKeySet();
```
or similar. `processMatches()` iterates with parallel stream and calls `matchRegistry.add(...)`. `ConcurrentHashMap.newKeySet()` makes this safe, and `getAllMatches()` still returns an immutable view via `Set.copyOf`.

core/src/main/java/de/jplag/frequency/MatchFrequencyEvaluator.java
```diff
@@ -25,20 +23,11 @@ public MatchFrequencyEvaluator(FrequencyAnalysisStrategy strategy) {
...
+        return strategy.getAllMatches().parallelStream().collect(ConcurrentHashMap::new,
+                (map, match) -> map.put(match, strategy.calculateMatchCount(match)), ConcurrentHashMap::putAll);
     }
```
tsaglam: Two small things here:
1. The call `collect(ConcurrentHashMap::new, ..., putAll)` builds per-thread maps and then merges them. Another option may be to use`toConcurrentMap` here, which fills one map directly, and the keys are unique because they come from a set:
```java
   return strategy.getAllMatches().parallelStream() .collect(Collectors.toConcurrentMap(Function.identity(), strategy::calculateMatchCount));
```
2. The method reads like a getter, but it does a parallel computation on every call. It also returns an empty map if `processMatches` wasn't called first. A name like `calculateMatchWeights` would probably fit better.

tsaglam (commented):
Really nice improvements in performance here. With these numbers, and since it no longer modifies the normal similarity values, we could consider enabling frequency analysis by default. This would probably lead to way more people actually using the feature. I have a few comments on concurrency, but the overall change looks good!

PR #3136, opened 2026-09-21
Python 3.14 Support

NikodemMiluch:
Updating the Python language module to support versoin 3.14.6 of Python.

- Used the new ANTLR Grammar from [https://github.com/antlr/grammars-v4](https://github.com/jplag/JPlag/pull/url)
- Added new tokens to handle the grammar
-  Added coverage tests for the new Python version


TwoOfTwelve (commented):
The code looks good. You might consider splitting the large test file into small files and add some tests that check the specific token sequence to make sure those basic cases work as intended.

